PR #55, opened 2023-05-08
fix swoole

jingjingxyk:


jingjingxyk:
https://github.com/crazywhalecc/static-php-cli/pull/78#issuecomment-1614459090

PR #64, opened 2023-05-31
Add support for libargon2

mpociot:
This PR is my attempt to add libargon2 support.
I have only tested this on macOS so far and there are two issues where I could use some help:

1. The "compile check" fails, because password-argon2 support is not an extension. That's why `php --ri password-argon2` fails.
2. When building PHP with the combination of "sodium" and "password-argon-2" (`sodium,password-argon-2`) the linker throws the following errors: 

```
duplicate symbol '_validate_inputs' in:
    /[redacted]/buildroot/lib/libsodium.a(libsodium_la-argon2-core.o)
    /[redacted]/buildroot/lib/libargon2.a(core.o)
duplicate symbol '_fill_memory_blocks' in:
    /[redacted]/buildroot/lib/libsodium.a(libsodium_la-argon2-core.o)
    /[redacted]/buildroot/lib/libargon2.a(core.o)
duplicate symbol '_fill_first_blocks' in:
    /[redacted]/buildroot/lib/libsodium.a(libsodium_la-argon2-core.o)
    /[redacted]/buildroot/lib/libargon2.a(core.o)
duplicate symbol '_initial_hash' in:
    /[redacted]/buildroot/lib/libsodium.a(libsodium_la-argon2-core.o)
    /[redacted]/buildroot/lib/libargon2.a(core.o)
duplicate symbol '_finalize' in:
    /[redacted]/buildroot/lib/libsodium.a(libsodium_la-argon2-core.o)
    /[redacted]/buildroot/lib/libargon2.a(core.o)
duplicate symbol '_initialize' in:
    /[redacted]/buildroot/lib/libsodium.a(libsodium_la-argon2-core.o)
    /[redacted]/buildroot/lib/libargon2.a(core.o)
ld: 6 duplicate symbols for architecture arm64
```

When building PHP only with `password-argon-2` this works fine.

## Tasks after merging

- [ ] Docs https://github.com/static-php/static-php-cli-docs/pull/6 (temporarily disable compatibility with libsodium for linux)
- [x] Integration test
- [ ] Common extensions test
- [ ] Bulk extensions test
- [ ] New issue (modified libargon2 still conflicts with libsodium on Linux)


crazywhalecc:
Question 1.

Extension check names can be overwritten, but only the name can be changed. For example, `mbregex` extension is a function for mbstring, create a function named `getDistName`: https://github.com/crazywhalecc/static-php-cli/blob/refactor/src/SPC/builder/extension/mbregex.php#L15

crazywhalecc:
Question 2.

The libsodium library seems to contain some argon2 code, causing conflicts, but directly using the password-argon2 extension with the libsodium library may also have some linking issues. Perhaps the patch code is still needed in the end.

mpociot:
Thank you for the in-depth analysis of this.

I'm not really sure how common the combination of sodium and argon2 password hashing is, but I stumbled upon this as I was trying to replicate the "default" PHP setup when installing PHP via brew. This does include sodium and argon2.

I agree - patching libargon2 seems to be the "easier" option in this case.
Would you prefer to fork libargon2 and rename the symbols in the fork, or do you think this could just be done on-the-fly in the build process itself?

crazywhalecc:
Well, I don't know much about the architecture of the libargon2 project, so it may take some time to read the source code to avoid potential bugs caused by simply changing the name. Some symbols may not be changed directly.

PHP installed via homebrew is a dynamically linked binary, and the code of the two libraries is not included in the PHP binary file, they can dynamically call the linked library as needed. static-php-cli just adds all of required library archives.

mpociot:
I have forked libargon2 and patched the source files so that there are no duplicate symbols.
It was not necessary to patch anything in the PHP sourcecode itself.

I don't really know what to use as the extension check name though, as password-argon2 does not get registered as a custom extension. It also does not show up in `php -i` besides being mentioned in the configure command.

Maybe it would be good to be able to skip this for extensions - we could still add a sanity check to ensure successful compilation.

crazywhalecc:
I have two ways:

1. change `Extension::getDistName()`, let it be able to return null value. Null value will not run `--ri $name` check.
2. Add a method, example: `Extension::check(string $binary): bool { return shell()->execWithResult("{$binary} --ri \"{$this->name}\"");  }` and overwrite in password-argon2 extension class.

Which one do you think is better?

crazywhalecc:
I introduced the `Extension::runCliCheck()` method in #253 , which overrides the check extension's script.

The issue has not been formally addressed for a long time. But now everything is working fine, and if there's nothing else to add, it's time to merge. 🎉

mpociot:
This looks good to me now, I'm already using this for Laravel Herd since its launch 👍

robdekort:
As a fresh Herd user, thanks for all your work on this @mpociot and @crazywhalecc.

PR #67, opened 2023-06-02
Add intl/ICU support

mpociot:
This PR adds support for the intl extension by compiling ICU statically.

I only tested MacOS support so far but modified the Linux build settings accordingly.

The download link for ICU points to a fixed version number, because the [GitHub releases](https://github.com/unicode-org/icu/releases) also contain other releases that are missing the file we actually want.

### DIFF
--- a/config/lib.json
+++ b/config/lib.json
@@ -460,5 +460,18 @@
         "static-libs-unix": [
             "libsodium.a"
         ]
+    },
+    "icu": {
+        "source": "icu",
+        "static-libs-unix": [
+            "libicudata.a",
+            "libicui18n.a",
+            "libicuio.a",
+            "libicutu.a",
+            "libicuuc.a"
+        ],
+        "headers": [
+
+        ]
     }
 }
--- a/config/source.json
+++ b/config/source.json
@@ -419,5 +419,13 @@
             "type": "file",
             "path": "LICENSE"
         }
+    },
+    "icu": {
+        "type": "url",
+        "url": "https://github.com/unicode-org/icu/releases/download/release-73-1/icu4c-73_1-src.tgz",
+        "license": {
+            "type": "file",
+            "path": "LICENSE"
+        }
     }
 }
--- a/src/SPC/builder/linux/LinuxBuilder.php
+++ b/src/SPC/builder/linux/LinuxBuilder.php
@@ -139,7 +139,7 @@ class LinuxBuilder extends BuilderBase
                 )
             );
         }
-        if ($this->getExt('swoole')) {
+        if ($this->getExt('swoole') || $this->getExt('intl')) {
             $extra_libs .= ' -lstdc++';
         }
         if ($this->getExt('imagick')) {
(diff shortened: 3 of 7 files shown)


crazywhalecc:
Linux build failed with a strange linking issue. This error did not appear in the compilation of ICU, but in the link section of compiling PHP.

```
/usr/bin/ld: /home/jerry/static-php-cli-fork/buildroot/lib/libicuuc.a(udata.ao): in function `doLoadFromCommonData(signed char, char const*, char const*, char const*, char const*, char const*, char const*, char const*, signed char (*)(void*, char const*, char const*, UDataInfo const*), void*, UErrorCode*, UErrorCode*) [clone .constprop.0]':
udata.cpp:(.text._ZL20doLoadFromCommonDataaPKcS0_S0_S0_S0_S0_S0_PFaPvS0_S0_PK9UDataInfoES1_P10UErrorCodeS8_.constprop.0+0x38a): undefined reference to `icudt73_dat'
/usr/bin/ld: udata.cpp:(.text._ZL20doLoadFromCommonDataaPKcS0_S0_S0_S0_S0_S0_PFaPvS0_S0_PK9UDataInfoES1_P10UErrorCodeS8_.constprop.0+0x467): undefined reference to `icudt73_dat'
collect2: error: ld returned 1 exit status
```

I have attempted to change `configure` options but still not working.

```php
<?php

declare(strict_types=1);

namespace SPC\builder\linux\library;

class icu extends LinuxLibraryBase
{
    public const NAME = 'icu';

    protected function build()
    {
        $root = BUILD_ROOT_PATH;
        $cppflag = 'CPPFLAGS="-DU_CHARSET_IS_UTF8=1  -DU_USING_ICU_NAMESPACE=1  -DU_STATIC_IMPLEMENTATION=1"';
        shell()->cd($this->source_dir . '/source')
            ->exec(
                "{$this->builder->configure_env} {$cppflag} ./runConfigureICU Linux " .
                '--enable-static ' .
                '--disable-shared ' .
                '--with-data-packaging=static ' .
                '--enable-release=yes ' .
                '--enable-extras=yes ' .
                '--enable-icuio=yes ' .
                '--enable-dyload=no ' .
                '--enable-tools=yes ' .
                '--enable-tests=no ' .
                '--enable-samples=no ' .
                "--prefix={$root}"
            )
            ->exec('make clean')
            ->exec("make -j{$this->builder->concurrency}")
            ->exec('make install');
    }
}
```

PR can be merged now, but only support macOS.

crazywhalecc:
And don't forget sorting config:

```bash
bin/spc sort-config ext
bin/spc sort-config lib
bin/spc sort-config source
```

crazywhalecc:
This error has been resolved via [Stack Overflow](https://stackoverflow.com/questions/72555103/udata-cpp-undefined-reference-to-icudt71-dat). But a new error has appeared :(

Failed in sanity check: `buildroot/bin/php --ri "intl"` and after running, the terminal outputs half of the content, and then `Aborted`.

I checked it, it caused by C++. That means we can only build icu in alpine docker (like swoole).

mpociot:
Yeah, I just stumbled upon that Stack Overflow answer as well! 👍 

Just for my understanding: is it a problem if this can only be built in alpine docker?

Thanks for letting me know about the sort-config commands and for already taking care of it.

PR #99, opened 2023-07-25
add memcached/libmemcached support for macOS

crazywhalecc:


crazywhalecc:
The reason why I use my [own libmemcached](https://github.com/crazywhalecc/libmemcached-macos) repo: 

1. I use patch files from https://formulae.brew.sh/formula/libmemcached `libmemcached.rb` to make it good to macOS.
2. The repo is just the latest libmemcached source and patched files. The libmemcached has been a long time since it was updated.

PR #128, opened 2023-08-18
mongodb source.json change from tgz to zip

kocoten1992:
since ver https://github.com/mongodb/mongo-php-driver/releases/tag/1.16.2

they don't support tgz anymore.


kocoten1992:
Hi @crazywhalecc, I think maybe we should revert this? this commit was on `Aug 18, 2023`, but on `Aug 22, 2023`, they did add tgz support again, see https://github.com/mongodb/mongo-php-driver/releases/tag/1.16.2

crazywhalecc:
I've reverted it

PR #130, opened 2023-08-20
Do some code quality check and fix #126

### LINKED ISSUE #126: Add `PHP_MICRO_FAKE_CLI` command options
https://www.workerman.net/a/1507#reply_961

crazywhalecc:
This PR is very large. And this PR contains the following changes that may affect the usage of other libraries:

- Add builder options, we deleted some builder class properties and changed it to options array (in order to expand more options in the future). e.g. `$builder->getOption('with-micro-fake-cli')`
- Change builder construct variables to options array.
- Make `FileSystem::replaceFile()` as deprecated, replace it with `FileSystem::replaceFileStr()`, `FileSystem::replaceFileRegex()`, `FileSystem::replaceFileUser()`. (PhpStorm always recognize str as regex and make some error)

The following changes do not affect any behavior:

- Added Unix SystemUtil method: `makeEnvVarString(array $vars): string` (build shell env string)
- Comment update
- Code structure update (like reducing depth of `if` statement)
- Exception PHPDoc and return void value strictly
- Added builder option methods: `getOption(string $key, mixed $default = null): mixed`
- Added builder option methods: `getOptions(): array`
- Added builder options methods: `setOptionIfNotExist(string $key, mixed $value): void`
- Added builder options methods: `setOption(string $key, mixed $value): void`

If anything works well, I will create the first RELEASE version. 😺

### DIFF
--- a/bin/setup-runtime
+++ b/bin/setup-runtime
@@ -22,7 +22,7 @@ esac
 
 # set project dir
 __DIR__=$(cd "$(dirname "$0")" && pwd)
-__PROJECT__=$(cd ${__DIR__}/../ && pwd)
+__PROJECT__=$(cd "${__DIR__}"/../ && pwd)
 
 # set download dir
 __PHP_RUNTIME_URL__="https://dl.zhamao.xin/static-php-cli/php-8.2.6-cli-${__OS_FIXED__}-${__ARCH__}.tar.gz"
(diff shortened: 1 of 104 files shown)


jingjingxyk:
把 `PATH` 变量也给加上，只要扩展一多，这个配置就变成必需的了。

给你举个例子： xml2-config  freetype-config  pg_config  ,这些文件都在 安装目录的 bin 目录下 

也就是构建前，需要导入如下构建环境变量 

image

Issue #132, opened 2023-08-20
help with install `swoole\ext-postgresql`


kocoten1992:
Hi @crazywhalecc , I'm trying to install https://github.com/swoole/ext-postgresql, but I cannot pass this check:

```
libpq libraries was not found.
set LIBPQ_CFLAGS and LIBPQ_LIBS environment or
install following package to obtain them:
libpq-dev (for debian and its varients)
postgresql-devel (for rhel varients)
libpq-devel (for newer fedora)
postgresql-libs (for arch and its varients)
postgresql-dev (for alpine)
postgresql (for homebrew)
configure: error: postgresql support needs libpq libraries to build
```

I was trying to use Docker (alpine linux) to build static cli with `swoole_postgresql` extension:
```
./bin/spc build "swoole_postgresql" --build-cli --debug
```

I already did this:

```diff
--- a/config/ext.json
+++ b/config/ext.json
@@ -380,6 +380,19 @@
         ],
         "unix-only": true
     },
+    "swoole_postgresql": {
+        "type": "external",
+        "source": "swoole_postgresql",
+        "arg-type": "custom",
+        "cpp-extension": true,
+        "lib-depends": [
+            "postgresql"
+        ],
+        "ext-depends": [
+            "swoole"
+        ],
+        "unix-only": true
+    },
     "swow": {
         "type": "external",
         "source": "swow",

```diff
--- a/config/source.json
+++ b/config/source.json
@@ -418,6 +418,16 @@
             "path": "LICENSE"
         }
     },
+    "swoole_postgresql": {
+        "type": "git",
+        "path": "php-src/ext/swoole_postgresql",
+        "rev": "master",
+        "url": "https://github.com/swoole/ext-postgresql",
+        "license": {
+            "type": "text",
+            "text": "This source file is subject to version 2.0 of the Apache license,\n that is bundled with this package in the file LICENSE, and is\n available through the world-wide-web at the following url:\n http://www.apache.org/licenses/LICENSE-2.0.html\n If you did not receive a copy of the Apache2.0 license and are unable\n to obtain it through the world-wide-web, please send a note to\n license@swoole.com so we can mail you a copy immediately."
+        }
+    },
     "swow": {
         "type": "git",
         "path": "php-src/ext/swow-src",
```

But cannot pass `libpq` check (I already install package as suggestion - but it does not work), do you have any idea?

crazywhalecc:
[...]

PR #133, opened 2023-08-21
add swoole pgsql (--enable-swoole-pgsql)

kocoten1992:
fix https://github.com/crazywhalecc/static-php-cli/issues/132

### DIFF
--- a/config/ext.json
+++ b/config/ext.json
@@ -370,7 +370,8 @@
         "arg-type": "custom",
         "cpp-extension": true,
         "lib-depends": [
-            "openssl"
+            "openssl",
+            "pqsql"
         ],
         "ext-depends": [
             "openssl"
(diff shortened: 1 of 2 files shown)


crazywhalecc (changes requested):
I think it's like alternative. Maybe using `ext-suggests: ["pgsql"]` is better?

```php
$arg .= $this->builder->getExt('pgsql') ? ' --enable-swoole-pgsql' : '';
```

crazywhalecc (approved):

crazywhalecc:
I just commit it that we can use swoole-pgsql coroutine support alternatively.

kocoten1992:
Thank you and sorry, I was a bit busy (was planning to do tomorrow), thank you again!

PR #135, opened 2023-08-21
fix: disable Zend Signals and enable Zend Max Execution Timers for ZTS builds

dunglas:
* Zend Signals are known to be currently broken with ZTS builds, they should be disabled
* The new Zend Max Execution Timers feature (specially designed for ZTS) fixes timeout support when using ZTS. It is enabled by default for ZTS build starting with PHP 8.3, but must be enabled manually for 8.1+ to preserve ABI compatibility, which isn't an issue here because we're creating a static build.

Related PHP issues:
* https://github.com/php/php-src/issues/9649#issuecomment-1264330874
* https://github.com/php/php-src/pull/5591#issuecomment-650064098
* https://github.com/php/php-src/pull/10141

The official ZTS Docker images also include these flags.

### DIFF
--- a/src/SPC/builder/linux/LinuxBuilder.php
+++ b/src/SPC/builder/linux/LinuxBuilder.php
@@ -168,8 +168,11 @@ class LinuxBuilder extends BuilderBase
 
         SourcePatcher::patchBeforeConfigure($this);
 
-        $json_74 = $this->getPHPVersionID() < 80000 ? '--enable-json ' : '';
-        $zts = $this->getOption('enable-zts', false) ? '--enable-zts --disable-zend-signals --enable-zend-max-execution-timers ' : '';
+        $phpVersionID = $this->getPHPVersionID();
+
+        $json_74 = $phpVersionID < 80000 ? '--enable-json ' : '';
+        $maxExecutionTimers = $this->getOption('enable-zts', false) && $this->getPHPVersionID() > 81000 ? '--enable-zend-max-execution-timers ' : '';
+        $zts = $this->getOption('enable-zts', false) ? '--enable-zts --disable-zend-signals ' : '';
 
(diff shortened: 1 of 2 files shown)


src/SPC/builder/linux/LinuxBuilder.php
```diff
@@ -168,8 +168,11 @@ public function buildPHP(int $build_target = BUILD_TARGET_NONE): void
 
         SourcePatcher::patchBeforeConfigure($this);
 
-        $json_74 = $this->getPHPVersionID() < 80000 ? '--enable-json ' : '';
-        $zts = $this->getOption('enable-zts', false) ? '--enable-zts ' : '';
+        $phpVersionID = $this->getPHPVersionID();
+
+        $json_74 = $phpVersionID < 80000 ? '--enable-json ' : '';
+        $maxExecutionTimers = $this->getOption('enable-zts', false) && $this->getPHPVersionID() > 81000 ? '--enable-zend-max-execution-timers ' : '';
```
crazywhalecc, line 174: PHP Version ID seems wrong. (8.1 => `80100`)

And reusing `$phpVersionID` for `$maxExecutionTimers` line?

crazywhalecc (commented):
I don't know much about ZTS bugs, but based on the information you provided, we need to disable the zend signal, and there is no better solution.

And `--enable-zend-max-execution-timers` makes ZTS safer.

Am I right?

dunglas:
Exactly.

src/SPC/builder/linux/LinuxBuilder.php
dunglas: Yes, sorry about that. I'll correct.

dunglas: fixed

crazywhalecc (approved):

PR #138, opened 2023-08-22
minor: various Docker optimizations

dunglas:
* make `bin/setup-runtime` compatible with Alpine Linux
* read the Dockerfile from stdin: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#pipe-dockerfile-through-stdin
* optimize Composer autoloader
* disable the cache of `apk`

### DIFF
--- a/bin/setup-runtime
+++ b/bin/setup-runtime
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 
 # set error-quit, verbose, non-variable-quit
 set -eu
@@ -51,6 +51,8 @@ china)
   ;;
 
 esac
+
+apk add --no-cache curl || true
 
 test -d "${__PROJECT__}"/downloads || mkdir "${__PROJECT__}"/downloads
 # download static php binary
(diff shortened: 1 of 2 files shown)


bin/setup-runtime
```diff
@@ -52,6 +52,8 @@ china)
 
 esac
 
+apk add --no-cache curl || true
```
crazywhalecc, line 55: It's working on Alpine, but other distros will show a warning message.

crazywhalecc (approved):
Checked, it works well (macOS, Debian, Alpine)

bin/setup-runtime
dunglas: warning hided

dunglas:
By the way, there is something weird on Mac M1: the Docker image is named `cwcc-spc-x86_64` while actually built for the aarch64 platform. Now that Docker (and Alpine containers) natively supports [multi-platform images](https://docs.docker.com/build/building/multi-platform/), I setting the `--platform` flag instead of changing the `FROM` clause should work.

crazywhalecc:
Oh I found that GH Action build failed. https://github.com/crazywhalecc/static-php-cli/actions/runs/5945970161/job/16125847564

Is that related to this docker optimization?

PR #145, opened 2023-08-26
ci: remove useless step and dump Composer autoloader

dunglas:


.github/workflows/build-linux-x86_64.yml
```diff
@@ -40,11 +40,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      # Install Ubuntu missing packages and mark os suffix
```
dunglas, line 43: This is useless because the binary is built in the Alpine container.

crazywhalecc (approved):

PR #159, opened 2023-09-04
Allow setting multiple license in extensions, libraries & sources, introduce phpunit tests

### LINKED ISSUE #110: Multiple license dump support
Some libraries and extensions have 2 or more, we need to support dumping more.

stloyd:
Fixes #110

Example:
```
stloyd@MacBook-Pro-2 static-php-cli % ls -l buildroot/license            
total 16
-rw-r--r--  1 stloyd  staff  1549 Sep  4 14:14 lib_zstd_0.txt
-rw-r--r--  1 stloyd  staff  1549 Sep  4 14:14 lib_zstd_1.txt
```

### DIFF
--- a/src/SPC/util/LicenseDumper.php
+++ b/src/SPC/util/LicenseDumper.php
@@ -45,46 +45,58 @@ class LicenseDumper
     {
         // mkdir first
         if (is_dir($target_dir) && !FileSystem::removeDir($target_dir)) {
-            logger()->warning('Target dump directory is noe empty, be aware!');
+            logger()->warning('Target dump directory is not empty, be aware!');
         }
         FileSystem::createDir($target_dir);
         foreach ($this->exts as $ext) {
             if (Config::getExt($ext, 'type') !== 'external') {
                 continue;
             }
+
             $source_name = Config::getExt($ext, 'source');
-            $content = $this->getSourceLicense($source_name);
-            file_put_contents($target_dir . '/ext_' . $ext . '.txt', $content);
+            foreach ($this->getSourceLicenses($source_name) as $index => $license) {
+                file_put_contents("{$target_dir}/ext_{$ext}_{$index}.txt", $license);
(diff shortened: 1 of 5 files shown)


src/SPC/util/LicenseDumper.php
```diff
@@ -45,46 +45,58 @@ public function dump(string $target_dir): bool
...
             if (Config::getExt($ext, 'type') !== 'external') {
                 continue;
             }
+
             $source_name = Config::getExt($ext, 'source');
-            $content = $this->getSourceLicense($source_name);
-            file_put_contents($target_dir . '/ext_' . $ext . '.txt', $content);
+            foreach ($this->getSourceLicenses($source_name) as $index => $license) {
+                file_put_contents("{$target_dir}/ext_{$ext}_{$index}.txt", $license);
+            }
```
crazywhalecc, line 59: What do you think about adding a `"suffix"` or `"name"` field here? Sometimes it is possible to distinguish what LICENSE represents.

Like PHP contains PHP License and Zend Engine License, we could use:

 ```json
"license": [
            {
                "type": "file",
                "path": "LICENSE"
            },
            {
                "suffix": "zend",
                "type": "file",
                "path": "Zend/LICENSE"
            }
        ]
```

Result:

```
src_php-src_0.txt
src_php-src_zend.txt
```

Or do you have any better suggestions?

stloyd: I'm open to adding this but also I didn't want to make this PR too big 😉

crazywhalecc: Not a problem, I can merge it first and then consider and update it later when checking the LICENSE for different libraries

stloyd: Here it goes: https://github.com/crazywhalecc/static-php-cli/pull/161

PR #163, opened 2023-09-06
Adjust dependencies in composer.json file

stloyd:
- Removed unsupported Symfony versions
- Bump minimal version of dev dependencies to the latest supported


stloyd:
@crazywhalecc I guess it's good to go as GH Actions are being run on PRs now.

PR #175, opened 2023-09-12
Initial code to get Box for phar building, replace DeployCommand

stloyd:
Info:
```shell
stloyd@MacBook-Pro-2 static-php-cli % box info spc.phar 

API Version: 1.1.0

Compression: None

Signature: SHA-1
Signature Hash: 127A780E883099614EA82B7BE52EE1E1A15E4E5B

Metadata: None

Contents: 691 files (2.35MB)

```

```shell
stloyd@MacBook-Pro-2 static-php-cli % ./spc.phar                                                                  
static-php-cli 2.0-rc6

Usage:
  command [options] [arguments]

Options:
  -h, --help            Display help for the given command. When no command is given display help for the list command
  -q, --quiet           Do not output any message
  -V, --version         Display this application version
      --ansi|--no-ansi  Force (or disable --no-ansi) ANSI output
  -n, --no-interaction  Do not ask any interactive question
  -v|vv|vvv, --verbose  Increase the verbosity of messages: 1 for normal output, 2 for more verbose output and 3 for debug

Available commands:
  build            build CLI binary
  doctor           Diagnose whether the current environment can compile normally
  download         [fetch] Download required sources
  dump-license     Dump licenses for required libraries
  extract          Extract required sources
  help             Display help for a command
  list             List commands
 build
  build:libs       Build dependencies
 dev
  dev:extensions   [list-ext] Helper command that lists available extension details
  dev:php-version  [dev:php-ver] Returns version of PHP located source directory
 micro
  micro:combine    Combine micro.sfx and php code together
```

Issue #124, opened 2023-08-08
Compiling with standalone spc binary cannot pass sanity check


crazywhalecc:
Reason: `/path/to/php phar:///path/to/the/static-php-cli/src/globals/test/xxx.php` run failed.

Solution:
1. Copy test files out of phar and run.
2. `try`, `eval` to run.
3. Use `php -r` to run.

PR #180, opened 2023-09-12
Replace file based extension tests with inline ones

stloyd:
This change simplifies testing of extensions against custom build PHP binaries, also inside the phar!

```shell
stloyd@MacBook-Pro-2 spc-test % ./spc build curl,dom --build-cli
     _        _   _                 _
 ___| |_ __ _| |_(_) ___      _ __ | |__  _ __
/ __| __/ _` | __| |/ __|____| '_ \| '_ \| '_ \
\__ \ || (_| | |_| | (_|_____| |_) | | | | |_) |
|___/\__\__,_|\__|_|\___|    | .__/|_| |_| .__/   v2.0-rc6
                             |_|         |_|
[17:28:32] [INFO] [EXEC] sysctl -n hw.ncpu
[17:28:32] [INFO] Build target: cli
[17:28:32] [INFO] Enabled extensions: curl, dom
[17:28:32] [INFO] Required libraries: zlib, openssl, curl, libiconv, libxml2
(...)
[17:31:01] [INFO] [EXEC] make -j10 EXTRA_CFLAGS='-g -Os' EXTRA_LIBS=' -liconv -framework CoreFoundation -framework SystemConfiguration  /Users/stloyd/Documents/spc-test/buildroot/lib/libxml2.a /Users/stloyd/Documents/spc-test/buildroot/lib/libiconv.a /Users/stloyd/Documents/spc-test/buildroot/lib/libcharset.a /Users/stloyd/Documents/spc-test/buildroot/lib/libcurl.a /Users/stloyd/Documents/spc-test/buildroot/lib/libssl.a /Users/stloyd/Documents/spc-test/buildroot/lib/libcrypto.a /Users/stloyd/Documents/spc-test/buildroot/lib/libz.a -lresolv' cli
[17:33:29] [INFO] [EXEC] dsymutil -f sapi/cli/php
[17:33:30] [INFO] [EXEC] strip sapi/cli/php
[17:33:30] [INFO] Deploying cli file
[17:33:30] [INFO] [EXEC] cp '/Users/stloyd/Documents/spc-test/source/php-src/sapi/cli/php' '/Users/stloyd/Documents/spc-test/buildroot/bin/'
[17:33:30] [INFO] running cli sanity check
[17:33:30] [INFO] [EXEC] /Users/stloyd/Documents/spc-test/buildroot/bin/php -r "echo \"hello\";"
[17:33:31] [INFO] [EXEC] /Users/stloyd/Documents/spc-test/buildroot/bin/php -r "assert(function_exists('curl_init'));"
[17:33:32] [INFO] [EXEC] /Users/stloyd/Documents/spc-test/buildroot/bin/php -r "assert(class_exists('\DOMDocument'));$doc = new DOMDocument();$doc->loadHtml('<html><head><meta charset="UTF-8"><meta http-equiv="Content-Type" content="text/html; charset=UTF-8"></head><body id='app'>Hello</body></html>');assert($doc->getElementById('app')->nodeValue === 'Hello');"
[17:33:32] [INFO] Build complete, used 299.775 s !
[17:33:32] [INFO] Static php binary path: /Users/stloyd/Documents/spc-test/buildroot/bin/php
[17:33:32] [INFO] License path: /Users/stloyd/Documents/spc-test/buildroot/license/
```

### DIFF
--- a/config/ext.json
+++ b/config/ext.json
@@ -4,17 +4,30 @@
         "source": "apcu"
     },
     "bcmath": {
-        "type": "builtin"
+        "type": "builtin",
+        "test": [
+            "assert(function_exists('bcscale'));"
+        ]
     },
     "bz2": {
         "type": "builtin",
         "arg-type": "with-prefix",
         "lib-depends": [
             "bzip2"
+        ],
+        "test": [
+            "$str = 'This is bz2 extension test';",
+            "assert(function_exists('bzdecompress'));",
+            "assert(function_exists('bzcompress'));",
+            "assert(bzdecompress(bzcompress($str, 9)) === $str);"
         ]
     },
     "calendar": {
-        "type": "builtin"
+        "type": "builtin",
+        "test": [
+            "assert(function_exists('cal_info'));",
+            "assert(is_array(cal_info(0)));"
+        ]
     },
     "ctype": {
         "type": "builtin"
@@ -24,6 +37,9 @@
         "arg-type": "with",
         "lib-depends": [
             "curl"
+        ],
+        "test": [
+            "assert(function_exists('curl_init'));"
         ]
     },
     "dba": {
@@ -37,6 +53,12 @@
         "lib-depends": [
             "libxml2",
             "zlib"
+        ],
+        "test": [
+            "assert(class_exists('\\DOMDocument'));",
+            "$doc = new DOMDocument();",
+            "$doc->loadHtml('<html><head><meta charset=\"UTF-8\"><meta http-equiv=\"Content-Type\" content=\"text/html; charset=UTF-8\"></head><body id='app'>Hello</body></html>');",
+            "assert($doc->getElementById('app')->nodeValue === 'Hello');"
         ]
     },
     "event": {
@@ -67,7 +89,10 @@
         "type": "builtin"
     },
     "filter": {
-        "type": "builtin"
+        "type": "builtin",
+        "test": [
+            "assert(function_exists('filter_var'));"
+        ]
     },
     "ftp": {
         "type": "builtin",
@@ -91,6 +116,12 @@
             "libwebp",
             "libjpeg",
             "freetype"
+        ],
+        "test": [
+            "assert(function_exists('gd_info'));",
+            "assert(gd_info()['PNG Support'] ?? false);",
+            "assert(gd_info()['GIF Create Support'] ?? false);",
+            "assert(gd_info()['GIF Read Support'] ?? false);"
         ]
     },
     "gettext": {
@@ -149,6 +180,9 @@
         "cpp-extension": true,
         "lib-depends": [
             "icu"
+        ],
+        "test": [
+            "assert(class_exists(NumberFormatter::class));"
         ]
     },
     "ldap": {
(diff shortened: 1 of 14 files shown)


stloyd:
This (along with #175) should allow to fix the https://github.com/crazywhalecc/static-php-cli/issues/124

Tested locally, logs in description is generated by custom-built PHP binary with SPC phar combined!

crazywhalecc:
I feel that including code in json is not a good solution. Maybe dealing with the existing xxx.php directly is an option, or putting these codes in the Extension.php file and using an abstract method.

stloyd:
I may agree that it looks confusing to have php-alike code in json file. I will play with other options today.

In the worst case, are you open to replacing that json file with another approach? (like php files or array?)

crazywhalecc:
I have previously considered using PHP to replace the current JSON format configuration file, for example `lib.json` becomes `lib.php`, but it will look a bit ambiguous with today's Library class. The first local version I had was even a version that did not use configuration files at all (all configurations were in the corresponding library and extension classes).

Now in the end I decided to use json and completely separate the PHP code for a simple purpose: everything seems to be low coupling, these configuration files are more like metadata that does not change, and the compiled code and parameters are like functions part.

If you use PHP as a configuration file, it is inevitable that dynamically executed code will be included in the configuration file. Maybe writing all configuration items directly as classes and deleting the configuration files is a more reasonable choice (just like the solution I originally abandoned).

crazywhalecc:
Now it is indeed a troublesome question: should the test code of these extensions be part of the logic code or a static configuration? Honestly I have no idea.

simonhamp:
I vote for separation of test code from configuration metadata. The tests MUST be PHP (if you're using PHP to run them), the other can be anything (e.g. JSON)

Why not simply include a relative path reference in the JSON metadata to any test file? e.g.

```json
    "bcmath": {
        "type": "builtin"
        "type": "builtin",
        "tests": "./tests/bcmath.php"
    },
```

This is explicit, can easily be resolved from whichever context is executing the build process, and acts as a pointer for any developer looking to see how things are wired up without having to dive into code.

crazywhalecc:
@simonhamp This seems to be possible and a compromise. Anyway, I think I finally decided to use `-r` to run the extended code. So if we want to use a file, spc needs to read the file and remove the `<?php` tag.

simonhamp:
@crazywhalecc the `-r` flag only seems to have been added in this PR. The "original" way (at least from the perspective of the changes here) was pointing to a PHP path - I don't see that needs to change

I guess I'm struggling to see how changing to using `-r` has made the tests pass - it looks like the failure is due to a path resolution issue when the test files are inside a .phar... isn't the fix here to find that and resolve it rather than moving all of the testing code from one execution context to another?

It feels super brittle to be interpolating PHP from a file (any file, JSON or PHP) into a shell string.

crazywhalecc:
> It feels super brittle to be interpolating PHP from a file (any file, JSON or PHP) into a shell string.

@simonhamp This is true. But in general, these test codes are theoretically controllable, because they are only in the src directory within the project, and everyone can check their security. What may need to be done is to escape special characters.

I also mentioned three options I can think of in issue #124, the first two seem riskier than this one.

simonhamp:
@crazywhalecc I agree with you, there's minimal concern from a security perspective.

I meant 'brittle' in the sense that they could be the source of various gotchas with interpolation depending on how convoluted the test needs to be, something that even escaping may not help with.

It still feels like a path resolution issue to me...

PR #183, opened 2023-09-13
PoC - Add new GH action to build binary SPC

stloyd:

PR #184, opened 2023-09-13
Add new GH action to build binary SPC

### LINKED ISSUE #179: Rewrite of "deploy" command to use new phar generation
> After introducing box (https://github.com/crazywhalecc/static-php-cli/pull/175), it is a good choice to directly rewrite `DeployCommand`. Ideally, we can automatically package single executable SPC files for different platforms and upload them when releasing a new version:

Roadmap:
- [ ] Use the command `bin/spc download --all` to download dependencies
- [ ] Compile the phpmicro SAPI with minimal extension required by static-php-cli itself: `bin/spc build pcntl,posix,mbstring,tokenizer,phar --build-micro`
- [ ] Package project: `vendor/bin/box compile`
- [ ] Generate executable binary: `bin/spc micro:combine spc.phar`
- [ ] Upload to release artifacts: `spc-macos-x86_64.tar.gz`

crazywhalecc:
It fixes #179 .
Related PR is #183 .

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -60,7 +60,7 @@ jobs:
       - name: "Build phpmicro"
         run: |
           SPC_USE_SUDO=yes bin/spc doctor --auto-fix
-          bin/spc build pcntl,posix,mbstring,tokenizer,phar --build-micro --debug
+          bin/spc build pcntl,posix,mbstring,tokenizer,phar --build-micro
 
       - name: "Build PHAR file"
         run: "composer build:phar"
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -1,0 +1,96 @@
+name: Build Release Artifacts
+
+on:
+  release:
+    types:
+      - published
+  workflow_dispatch:
+
+jobs:
+  build-release-artifacts:
+    name: "Build Release Artifacts"
+    runs-on: ${{ matrix.operating-system }}
+    strategy:
+      fail-fast: false
+      matrix:
+        php-version:
+          - "8.1"
+        operating-system:
+          - "ubuntu-latest"
+          - "macos-latest"
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v4"
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          coverage: none
+          tools: composer:v2
+          php-version: "${{ matrix.php-version }}"
+          ini-values: memory_limit=-1
+
+      - name: "Get Composer Cache Directory"
+        id: composer-cache
+        run: |
+          echo "dir=$(composer config cache-files-dir)" >> $GITHUB_OUTPUT
+
+      - name: "Cache Composer dependencies"
+        uses: "actions/cache@v3"
+        with:
+          path: "${{ steps.composer-cache.outputs.dir }}"
+          key: "php-${{ matrix.php-version }}-locked-composer-${{ hashFiles('**/composer.lock') }}"
+          restore-keys: |
+            php-${{ matrix.php-version }}-locked-composer-
+      - name: "Install locked dependencies"
+        run: "composer install --no-interaction --no-progress"
+
+      # Cache downloaded source
+      - id: cache-download
+        uses: actions/cache@v3
+        with:
+          path: downloads
+          key: php-${{ matrix.php-version }}-dependencies
+
+      # If there's no dependencies cache, fetch sources
+      - if: steps.cache-download.outputs.cache-hit != 'true'
+        name: "Download sources"
+        run: bin/spc download --with-php=${{ matrix.php-version }} --all
+
+      - name: "Build phpmicro"
+        run: |
+          SPC_USE_SUDO=yes bin/spc doctor --auto-fix
+          bin/spc build pcntl,posix,mbstring,tokenizer,phar --build-micro --debug
+
+      - name: "Build PHAR file"
+        run: "composer build:phar"
+
+      - name: "Generate Executable"
+        run: "bin/spc micro:combine spc.phar -O spc"
+
+      - name: "Archive Executable"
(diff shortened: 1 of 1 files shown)


.github/workflows/release-build.yml
```diff
@@ -0,0 +1,96 @@
...
+      - name: "Get Composer Cache Directory"
+        id: composer-cache
+        run: |
+          echo "dir=$(composer config cache-files-dir)" >> $GITHUB_OUTPUT
+
+      - name: "Cache Composer dependencies"
+        uses: "actions/cache@v3"
+        with:
+          path: "${{ steps.composer-cache.outputs.dir }}"
+          key: "php-${{ matrix.php-version }}-locked-composer-${{ hashFiles('**/composer.lock') }}"
```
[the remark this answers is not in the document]
stloyd: Additionally it will allow to introduce @dependabot or similar tool for maintain.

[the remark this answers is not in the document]
stloyd: Proposal to hold `composer.lock` in repository: https://github.com/crazywhalecc/static-php-cli/pull/185

.github/workflows/release-build.yml
```diff
@@ -0,0 +1,96 @@
...
+          restore-keys: |
+            php-${{ matrix.php-version }}-locked-composer-
+      - name: "Install locked dependencies"
+        run: "composer install --no-interaction --no-progress"
```
dunglas, line 46: Tiny optim:

```suggestion
        run: "composer install --no-interaction --no-progress --no-dev -a"
```

.github/workflows/release-build.yml
```diff
@@ -0,0 +1,96 @@
...
+
+      # Cache downloaded source
+      - id: cache-download
+        uses: actions/cache@v3
+        with:
+          path: downloads
+          key: php-${{ matrix.php-version }}-dependencies
+
+      # If there's no dependencies cache, fetch sources
+      - if: steps.cache-download.outputs.cache-hit != 'true'
```
dunglas, line 56: What if dependencies have been updated (e.g.: security fix) since the last retrieval?
Shouldn't the command always be run?

dunglas:
That would be great to also publish in GitHub releases compiled versions of the SAPIs with most (all) supported PHP extensions. This would allow CIs and scripts to just download the latest release instead of compiling it. For instance if we could download an always up-to-date version of libphp from GitHub, this would dramatically speedup the builds of FrankenPHP. Currently, building libphp again and again takes ~30 minutes: https://github.com/dunglas/frankenphp/actions/runs/6178375726/job/16771499374

.github/workflows/release-build.yml
crazywhalecc: `dev` packages is required because it contains `box` which is used to making phar.

.github/workflows/release-build.yml
crazywhalecc: It depends on what our definition of a security fix is. If there are security updates for dependent libraries included in the compilation results, this is generally not necessary because this Action only distributes spc binaries. If it is a security update for the project itself, it has nothing to do with the cache of the download itself.

crazywhalecc:
@dunglas I think this can be done by creating a new workflow for automatically building libphp in the Fork repository, because many people want to Fork the project and then build it manually. If there are too many irrelevant Actions, it may be confusing.

.github/workflows/release-build.yml
stloyd: I think we could use a lock hash similar to the composer one with the usage of download lock: `DOWNLOAD_PATH . '/.lock.json'`

crazywhalecc: `.lock.json` is now more like a download meta of downloaded files. I am not very clear about the reason and method of using it as a cache key.

stloyd: The reason I can imagine is a stale cache with security issues, but we can add a clean-up workflow to be scheduled ie once a week, that would cover this case too

crazywhalecc: If we just make a weekly build action, I think it could be done without any download caches. What I want to know more is, can the Action Cache take effect in the Fork warehouse?

stloyd: AFAIK cache & artifacts are not shared with the forks, the only way it is possible is when the fork does PR to the main repository and must use the same workflow, cause cache is per workflow file and jobs, but it can be shared within the same workflow (requires marking jobs are depending on each other).

stloyd: I think that working on the https://github.com/crazywhalecc/static-php-cli/issues/187 idea would be the only way to solve this problem on the DX side, in this repository we could setup weekly builds to ensure security fixes are always included in binaries.

.github/workflows/release-build.yml
```diff
@@ -0,0 +1,96 @@
...
+        name: "Download sources"
+        run: bin/spc download --with-php=${{ matrix.php-version }} --all
+
+      - name: "Build phpmicro"
+        run: |
+          SPC_USE_SUDO=yes bin/spc doctor --auto-fix
+          bin/spc build pcntl,posix,mbstring,tokenizer,phar --build-micro --debug
```
stloyd, line 63: I would change (or remove) `--debug` to be env based.

stloyd:
@crazywhalecc is there anything missing to get this merged and be tested in the wild?

I would love to get feedback and improve if needed rather than wait 😉

stloyd (approved):

stloyd:
Great, thank you @crazywhalecc ! Just one more thing :) Can you trigger a run of that workflow? So we can download the pre-built binary & test it? Or do you want to wait for the release of 2.0?
https://github.com/crazywhalecc/static-php-cli/actions/workflows/release-build.yml

crazywhalecc:
@stloyd undoubted https://github.com/crazywhalecc/static-php-cli/actions/runs/6206145910

PR #188, opened 2023-09-15
Add GH action to build library artifacts

stloyd:
Generated artifacts can be downloaded & re-used instead of built locally.

Example of `curl` library pre-build for osx: https://github.com/stloyd/static-php-build/actions/runs/6198320093

Extracted locally to `buildroot` dir:
```shell
stloyd@MacBook-Pro-2 static-php-cli % bin/spc build curl --build-micro    
     _        _   _                 _           
 ___| |_ __ _| |_(_) ___      _ __ | |__  _ __  
/ __| __/ _` | __| |/ __|____| '_ \| '_ \| '_ \ 
\__ \ || (_| | |_| | (_|_____| |_) | | | | |_) |
|___/\__\__,_|\__|_|\___|    | .__/|_| |_| .__/   v2.0-rc6
                             |_|         |_|    
[13:30:27] [INFO] [EXEC] sysctl -n hw.ncpu
[13:30:27] [INFO] Build target: micro
[13:30:27] [INFO] Enabled extensions: curl
[13:30:27] [INFO] Required libraries: zlib, openssl, curl
[13:30:29] [NOTI] lib [zlib] already built
[13:30:29] [NOTI] lib [openssl] already built
[13:30:29] [NOTI] lib [curl] already built
[13:30:29] [INFO] patching before-configure for curl checks
[13:30:29] [INFO] Extension [curl] patched before buildconf
[13:30:29] [INFO] Entering dir: /Users/stloyd/Documents/static-php-cli/source/php-src
[13:30:29] [INFO] [EXEC] ./buildconf --force
[13:30:35] [INFO] Extension [curl] patched before configure
[13:30:35] [INFO] Entering dir: /Users/stloyd/Documents/static-php-cli/source/php-src
[13:30:35] [INFO] Using configure: --with-curl
(...)
```

### DIFF
--- a/.github/workflows/build-libraries.yml
+++ b/.github/workflows/build-libraries.yml
@@ -1,0 +1,130 @@
+name: Build Library Dependency Artifacts
+
+on:
+    release:
+        types:
+            - published
+    workflow_dispatch:
+
+jobs:
+    build-release-artifacts:
+        name: "Build Library Dependency Artifacts"
+        runs-on: ${{ matrix.operating-system }}
+        strategy:
+            fail-fast: false
+            matrix:
+                php-version:
+                    - "8.1"
+                operating-system:
+                    - "ubuntu-latest"
+                    - "macos-latest"
+                library:
+                    # Commented out libraries are not yet supported
+                    - brotli
+                    - bzip2
+                    - curl
+                    - freetype
+                    #- glfw
+                    - gmp
+                    - icu
+                    - imagemagick
+                    - libavif
+                    - libevent
+                    #- libffi
+                    - libiconv
+                    - libjpeg
+                    #- libmcrypt
+                    #- libmemcached
+                    - libpng
+                    - libsodium
+                    - libssh2
+                    - libwebp
+                    - libxml2
+                    - libxslt
+                    - libyaml
+                    - libzip
+                    - mcrypt
+                    - ncurses
+                    - nghttp2
+                    - onig
+                    - openssl
+                    #- pkg-config
+                    - postgresql
+                    #- pthreads4w
+                    - readline
+                    - snappy
+                    - sqlite
+                    - xz
+                    - zlib
+                    - zstd
+
+        steps:
+            - name: "Checkout"
+              uses: "actions/checkout@v4"
+
+            - name: "Install PHP"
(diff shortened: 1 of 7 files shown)


crazywhalecc:
This is a good idea, but we may not do it at the moment. Because each compiled extension only includes the libraries it must depend on, ignoring optional dependencies. I know we could include all suggested libs, but it will make it impossible to compile without suggesting libs. Additionally there are some `lib-suggests` not working very well and some libraries have almost identical build results.

stloyd:
1. Suggested libraries are already included when doing a build of libraries: https://github.com/crazywhalecc/static-php-cli/blob/main/src/SPC/util/DependencyUtil.php#L115,
2. If something will fail on actions, we can exclude it for now,
3. I'm aware that we need to skip some builds, but OSX / Ubuntu ones work quite well,
4. Future improvement may include direct usage of such pre-built binaries directly in commands, so we can have new options like `--no-cache` which will always build a library, and cache link can be also directly included in `config/lib.json`

I think we need to start somewhere ;)

stloyd:
Reworked the action & it now uses pre-generated binary 😮‍💨

crazywhalecc:
I have an idea for distributing `spc` binary: push `spc` binary to docs repo. But this also requires uploading binaries for different systems and different architectures. 

When I hosted statically compiled PHP on my own server, the server added an automatic jump link to a specified version of the PHP binary based on the architecture and version passed in. If pages hosting is used entirely here, this request may not be handled dynamically.

WDYT?

crazywhalecc:
I have the following questions to figure out.

For Actions, what is the scope of its cache? Assuming others use `uses: static-php/setup-static-php`, how do you end up using this compiled cache to speed things up?

Because there are too many libraries now with optional dependencies. Optional dependencies for compilation are always dynamic and can be combined in too many ways.

**Also**, I think it would be better to set up a separate `setup-static-php-libs` Actions.

1. `setup-static-php` also depends on it.
2. `setup-static-php-deps` will first check the build cache for the current combination of extensions and libraries.
3. If there is a cache of the incoming combination in `setup-static-php-deps`, use it directly, otherwise build it.
4. `setup-static-php` continues building PHP.

PR #193, opened 2023-09-18
rhel support & ldap support

DubbleClick:
I wasn't quite able to get ldap to work when building static-php-cli, but it's trivial on rhel/debian - just run

```
dnf install sqlite-devel openldap-devel
ln -s /usr/lib64/libldap.so /usr/lib/libldap.so && ln -s /usr/lib64/liblber.so /usr/lib/liblber.so
php-src-folder/configure --with-ldap=/usr
```

openldap-devel is called libldap2-dev on debian.

Edit: If you want to add a pipeline for rhel (not required imo) this is the full install script to get things running on rhel.
epel-release is required for most of the things to install, remi only for php 8.2. The default still downloads 8.0 by default, which doesn't work.

```
dnf install epel-release
dnf install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm 
dnf install -y https://rpms.remirepo.net/enterprise/remi-release-9.rpm  
dnf module enable php:remi-8.2
dnf install tar git zip php php-sodium php-zip libsodium wget

cd /opt
git clone https://github.com/crazywhalecc/static-php-cli.git
wget https://musl.libc.org/releases/musl-1.2.4.tar.gz

tar -xzf musl-1.2.4.tar.gz
cd musl-1.2.4.tar.gz
./configure
make -j
make install
cd ../static-php-cli

dnf install bison flex bzip2 cmake patch
dnf install sqlite-devel openldap-devel
ln -s /usr/lib64/libldap.so /usr/lib/libldap.so && ln -s /usr/lib64/liblber.so /usr/lib/liblber.so

php-src-dir/configure --with-ldap=/usr
```

### DIFF
--- a/config/ext.json
+++ b/config/ext.json
@@ -153,9 +153,15 @@
     },
     "ldap": {
         "type": "builtin",
-        "arg-type": "with",
+        "arg-type": "with-prefix",
         "lib-depends": [
             "ldap"
+        ],
+        "lib-suggests": [
+            "openssl"
+        ],
+        "ext-suggests": [
+            "openssl"
         ]
     },
     "mbregex": {
--- a/config/ext.json
+++ b/config/ext.json
@@ -158,10 +158,14 @@
             "ldap"
         ],
         "lib-suggests": [
-            "openssl"
+            "openssl",
+            "zlib",
+            "gmp",
+            "libsodium"
         ],
         "ext-suggests": [
-            "openssl"
+            "openssl",
+            "zlib"
         ]
     },
     "mbregex": {
--- a/config/ext.json
+++ b/config/ext.json
@@ -158,14 +158,11 @@
             "ldap"
         ],
         "lib-suggests": [
-            "openssl",
-            "zlib",
             "gmp",
             "libsodium"
         ],
         "ext-suggests": [
-            "openssl",
-            "zlib"
+            "openssl"
         ]
     },
     "mbregex": {
@@ -239,6 +236,9 @@
         "arg-type": "with",
         "lib-depends": [
             "openssl"
+        ],
+        "ext-depends": [
+            "zlib"
         ]
     },
     "pcntl": {
--- a/src/SPC/builder/unix/library/ldap.php
+++ b/src/SPC/builder/unix/library/ldap.php
@@ -1,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace SPC\builder\unix\library;
+
+trait ldap
+{
+    protected function build(): void
+    {
+        shell()->cd($this->source_dir)
+            ->exec(
+                $this->builder->configure_env . ' ' .
+                'LDFLAGS="-static"' .
+                ' ./configure ' .
+                '--enable-static ' .
+                '--disable-shared ' .
+                '--disable-slapd ' .
+                '--without-systemd ' .
+                ($this->builder->getLib('openssl') ? '--with-tls=openssl ' : '') .
+                '--prefix='
+            )
+            ->exec('make clean')
+            ->exec('make depend')
+            ->exec("make -j{$this->builder->concurrency}")
+            ->exec('make install DESTDIR=' . BUILD_ROOT_PATH);
+        $this->patchPkgconfPrefix(['ldap.pc', 'lber.pc']);
+    }
+}
--- a/src/SPC/builder/unix/library/ldap.php
+++ b/src/SPC/builder/unix/library/ldap.php
@@ -11,22 +11,22 @@ trait ldap
         shell()->cd($this->source_dir)
             ->exec(
                 $this->builder->configure_env . ' ' .
-                'CC="musl-gcc -I' . BUILD_INCLUDE_PATH . '" ' .
-                'LDFLAGS="-static -L' . BUILD_LIB_PATH . '" ' .
-                ($this->builder->getLib('openssl') && $this->builder->getExt('zlib') ? 'LIBS="-lssl -lcrypto -lz" ' : '') .
+                $this->builder->makeAutoconfFlags(AUTOCONF_LDFLAGS | AUTOCONF_CPPFLAGS) .
                 ' ./configure ' .
                 '--enable-static ' .
                 '--disable-shared ' .
                 '--disable-slapd ' .
                 '--disable-slurpd ' .
                 '--without-systemd ' .
+                '--without-cyrus-sasl ' .
                 ($this->builder->getLib('openssl') && $this->builder->getExt('zlib') ? '--with-tls=openssl ' : '') .
                 ($this->builder->getLib('gmp') ? '--with-mp=gmp ' : '') .
                 ($this->builder->getLib('libsodium') ? '--with-argon2=libsodium ' : '') .
                 '--prefix='
             )
             ->exec('make clean')
-            ->exec('make depend')
+            // remove tests and doc to prevent compile failed with error: soelim not found
+            ->exec('sed -i -e "s/SUBDIRS= include libraries clients servers tests doc/SUBDIRS= include libraries clients servers/g" Makefile')
             ->exec("make -j{$this->builder->concurrency}")
             ->exec('make install DESTDIR=' . BUILD_ROOT_PATH);
         $this->patchPkgconfPrefix(['ldap.pc', 'lber.pc']);
--- a/src/SPC/builder/unix/library/ldap.php
+++ b/src/SPC/builder/unix/library/ldap.php
@@ -8,6 +8,13 @@ trait ldap
 {
     protected function build(): void
     {
+        $alt = '';
+        // openssl support
+        $alt .= $this->builder->getLib('openssl') && $this->builder->getExt('zlib') ? '--with-tls=openssl ' : '';
+        // gmp support
+        $alt .= $this->builder->getLib('gmp') ? '--with-mp=gmp ' : '';
+        // libsodium support
+        $alt .= $this->builder->getLib('libsodium') ? '--with-argon2=libsodium ' : '';
         shell()->cd($this->source_dir)
             ->exec(
                 $this->builder->configure_env . ' ' .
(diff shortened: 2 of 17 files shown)


src/SPC/builder/unix/library/ldap.php
```diff
@@ -0,0 +1,16 @@
...
+declare(strict_types=1);
+
+namespace SPC\builder\unix\library;
+
+trait ldap
+{
+    protected function build(): void
+    {
+        shell()->cd($this->source_dir)
+            ->exec('dnf install -y openldap-devel') // libldap2-dev for debian
```
DubbleClick, line 12: this is probably not the right place, please redirect me there!

config/ext.json
```diff
@@ -153,7 +153,7 @@
     },
     "ldap": {
         "type": "builtin",
-        "arg-type": "with",
+        "arg-type": "with-prefix",
```
DubbleClick, line 156: this must be "with-ldap=/usr"

src/SPC/builder/unix/library/ldap.php
DubbleClick: this should probably be in LinuxMuslCheckList.php?

src/SPC/doctor/item/LinuxMuslCheck.php
```diff
@@ -39,7 +39,9 @@ public function fixMusl(array $distro): bool
         $install_cmd = match ($distro['dist']) {
             'ubuntu', 'debian' => 'apt-get install musl musl-tools -y',
             'alpine' => 'apk add musl musl-utils musl-dev',
-            default => throw new RuntimeException('Current linux distro is not supported for auto-install musl packages'),
+            'rhel' => 'dnf install tar wget git zip bison flex bzip2 cmake patch && wget https://musl.libc.org/releases/musl-1.2.4.tar.gz && tar -zxvf musl-1.2.4.tar.gz && rm musl-1.2.4.tar.gz && cd musl-1.2.4 && ./configure && make -j && make install && cd ..',
```
DubbleClick, line 42: maybe the version shouldn't be hardcoded here, open for suggestions

DubbleClick: also, since this builds from source, it would work on every unix based os - make this the default branch?

config/ext.json
stloyd: Should be a `custom` here + add a new method `getUnixConfigureArg()` on the extension class.

DubbleClick: thanks, changed it

jingjingxyk:
please with  openldap  static link library 

build reference : https://github.com/jingjingxyk/swoole-cli/blob/new_dev/sapi/src/builder/library/openldap.php#L37

crazywhalecc:
@DubbleClick Only libraries supported by static compilation can be added to static-php-cli. Currently, all packages installed by dnf are dynamically linked. So the core point is to write a method that can statically compile the openldap library.

src/SPC/doctor/item/LinuxToolCheckList.php
```diff
@@ -16,15 +16,15 @@ class LinuxToolCheckList
     use UnixSystemUtilTrait;
 
     public const TOOLS_ALPINE = [
-        'make', 'bison', 'flex',
+        'perl', 'make', 'bison', 'flex',
```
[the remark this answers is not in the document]
crazywhalecc: We could already compile openssl. Is there any place where perl must be used?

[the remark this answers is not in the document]
DubbleClick: https://github.com/openssl/openssl/issues/13761

FindBin.pm is only provided by perl and necessary to compile openssl.

According to one of the maintainers, perl was always needed: https://github.com/openssl/openssl/issues/13761#issuecomment-848644058

src/SPC/doctor/item/LinuxMuslCheck.php
DubbleClick: edit: openssl apparently requires the full perl installation, gets stuck on openssl build otherwise:

```
[root@localhost openssl]# PKG_CONFIG='/opt/static-php-cli/buildroot/bin/pkg-config' PKG_CONFIG_PATH='/opt/static-php-cli/buildroot/lib/pkgconfig' CC='musl-gcc' CXX='g++' PATH='/opt/static-php-cli/buildroot/bin:/usr/local/musl/bin:/root/.local/bin:/root/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin' PKG_CONFIG='/opt/static-php-cli/buildroot/bin/pkg-config' PKG_CONFIG_PATH='/opt/static-php-cli/buildroot/lib/pkgconfig' CFLAGS='' CC='musl-gcc -static -idirafter /opt/static-php-cli/buildroot/include -idirafter /usr/include/  -idirafter /usr/include/x86_64-linux-gnu/ '  ./Configure no-shared zlib --prefix=/ --libdir=lib -static --with-zlib-include=/opt/static-php-cli/buildroot/include --with-zlib-lib=/opt/static-php-cli/buildroot/lib no-legacy linux-x86_64
Can't locate IPC/Cmd.pm in @INC (you may need to install the IPC::Cmd module) (@INC contains: /opt/static-php-cli/source/openssl/util/perl /usr/local/lib64/perl5/5.32 /usr/local/share/perl5/5.32 /usr/lib64/perl5/vendor_perl /usr/share/perl5/vendor_perl /usr/lib64/perl5 /usr/share/perl5 /opt/static-php-cli/source/openssl/external/perl/Text-Template-1.56/lib) at /opt/static-php-cli/source/openssl/util/perl/OpenSSL/config.pm line 19.
BEGIN failed--compilation aborted at /opt/static-php-cli/source/openssl/util/perl/OpenSSL/config.pm line 19.
Compilation failed in require at ./Configure line 23.
BEGIN failed--compilation aborted at ./Configure line 23.
```

src/SPC/builder/unix/library/ldap.php
```diff
@@ -0,0 +1,28 @@
...
+        shell()->cd($this->source_dir)
+            ->exec(
+                'CC="musl-gcc -I' . $openssl_include . ' -L' . $openssl_lib . '" LDFLAGS="-static" ./configure ' .
+                '--enable-static=yes ' .
+                '--disable-shared ' .
+                '--disable-slapd ' .
+                '--without-systemd ' .
+                '--with-tls=openssl ' .
+                '--prefix=' . BUILD_ROOT_PATH
+            )
```
DubbleClick, line 22: this doesn't work with the openssl build produced here, errors out with

`checking for ssl_ctx_set_ciphersuites in -lssl... no`

Doing it manually and supplying $openssl_install/include and $openssl_install/lib to the build ldap works.

```
#!/bin/bash

openssl_version=1.1.1w
openldap_version=2.6.6

# Build OpenSSL
wget https://openssl.org/source/openssl-$openssl_version.tar.gz
tar -xvf openssl-$openssl_version.tar.gz
cd openssl-$openssl_version
openssl_install=`pwd`/local

CC="musl-gcc -I$openssl_install/include/openssl -L$openssl_install/lib -static -idirafter /usr/include/ -idirafter /usr/include/x86_64-linux-gnu/" ./config no-shared -static --prefix=$openssl_install
make -j
make install
cd ..

# Build OpenLDAP
wget ftp://ftp.openldap.org/pub/OpenLDAP/openldap-release/openldap-$openldap_version.tgz
tar -xvf openldap-$openldap_version.tgz
cd openldap-$openldap_version

CC="musl-gcc -I$openssl_install/include -L$openssl_install/lib" LDFLAGS="-static" ./configure --disable-shared --disable-slapd --with-tls=openssl
make depend
make -j

cd ..

# Cleanup
rm openssl-$openssl_version.tar.gz
rm openldap-$openldap_version.tgz
```

DubbleClick: I think I might have to pass the psth to LDFLAGS too

DubbleClick:
ready for review now, I think everything is working as it should

DubbleClick:
For what it's worth, I've only required `perl` for RHEL distros now, but I think it should be a requirement for debian and alpine too. Doesn't seem like a good idea to rely on perl being installed by default.

DubbleClick:
Cleaned up a few last things: enabled ldap for MacOS and moved the source to /php-src/ext/ldap.

DubbleClick:
When compiling ldap with openssl support, the zlib extension needs to be enabled. If compiled without openssl support, zlib is not required. I didn't put zlib in ext-requires because of that and instead check for openssl lib & zlib extension before enabling --with-tls

crazywhalecc:
> For what it's worth, I've only required `perl` for RHEL distros now, but I think it should be a requirement for debian and alpine too. Doesn't seem like a good idea to rely on perl being installed by default.

Debian and Alpine has installed perl I guess.

crazywhalecc:
> When compiling ldap with openssl support, the zlib extension needs to be enabled. If compiled without openssl support, zlib is not required. I didn't put zlib in ext-requires because of that and instead check for openssl lib & zlib extension before enabling --with-tls

My understanding: If you want ldap to support openssl, then it must include the zlib library. This can actually be solved like this: openssl extension adds `ext-depends: zlib`.

DubbleClick:
Correct, sure I can add that.

crazywhalecc:
Although the openssl extension does not require the zlib extension, according to the current extension inclusion logic, the zlib extension should be enabled when you include the zlib library, so this is reasonable.

crazywhalecc:
## Documentation Task

- [ ] extension.md
- [ ] CliGenerator.vue
- [ ] doctor for RHEL part

DubbleClick:
The doctor command doesn't have notes about specific operating systems. Rhel would just count under linux, I think we have nothing to document there.

crazywhalecc (changes requested):
I tested on my local AlmaLinux VM, and I found that the compilation part worked perfectly, but the doctor part needed some changes: The path was not added correctly when compiling musl. We need to manually write `~/.bashrc` or execute `export` once in the current terminal.

crazywhalecc (approved):

PR #196, opened 2023-09-20
Use pre-generated binary in `spc-alpine-docker`

stloyd:
Due to a lack of access to GH action artifacts from outside of GH, I have used workaroud: https://nightly.link

When the release contains binary files, we probably can replace that workaround.


crazywhalecc:
Using pre-built `spc` binaries is probably not a very good idea when the source code already exists. Because many times debugging on Linux requires testing of local code, using pre-built binaries will become conflicting.

PR #197, opened 2023-09-20
Add way to run docker build from pre-build binary

stloyd:
```shell
stloyd@MacBook-Pro-2 static-php-cli % ./spc.phar build snappy --build-micro --docker
     _        _   _                 _           
 ___| |_ __ _| |_(_) ___      _ __ | |__  _ __  
/ __| __/ _` | __| |/ __|____| '_ \| '_ \| '_ \ 
\__ \ || (_| | |_| | (_|_____| |_) | | | | |_) |
|___/\__\__,_|\__|_|\___|    | .__/|_| |_| .__/   v2.0-rc6
                             |_|         |_|    
[15:56:17] [INFO] [EXEC] echo | gcc -E -x c - -march=corei7
cc1: error: unknown value 'corei7' for '-march'
cc1: note: valid arguments are: armv8-a armv8.1-a armv8.2-a armv8.3-a armv8.4-a armv8.5-a armv8.6-a armv8.7-a armv8.8-a armv8-r armv9-a armv9.1-a armv9.2-a armv9.3-a native
[15:56:17] [INFO] [EXEC] echo | gcc -E -x c - -mtune=core-avx2
cc1: error: unknown value 'core-avx2' for '-mtune'
cc1: note: valid arguments are: cortex-a34 cortex-a35 cortex-a53 cortex-a57 cortex-a72 cortex-a73 thunderx thunderxt88p1 thunderxt88 octeontx octeontx81 octeontx83 thunderxt81 thunderxt83 ampere1 ampere1a emag xgene1 falkor qdf24xx exynos-m1 phecda thunderx2t99p1 vulcan thunderx2t99 cortex-a55 cortex-a75 cortex-a76 cortex-a76ae cortex-a77 cortex-a78 cortex-a78ae cortex-a78c cortex-a65 cortex-a65ae cortex-x1 cortex-x1c ares neoverse-n1 neoverse-e1 octeontx2 octeontx2t98 octeontx2t96 octeontx2t93 octeontx2f95 octeontx2f95n octeontx2f95mm a64fx tsv110 thunderx3t110 zeus neoverse-v1 neoverse-512tvb saphira cortex-a57.cortex-a53 cortex-a72.cortex-a53 cortex-a73.cortex-a35 cortex-a73.cortex-a53 cortex-a75.cortex-a55 cortex-a76.cortex-a55 cortex-r82 cortex-a510 cortex-a710 cortex-a715 cortex-x2 cortex-x3 neoverse-n2 demeter neoverse-v2 generic
[...]

PR #198, opened 2023-09-23
Another way to run docker in spc binary

crazywhalecc:
Related PR: #197.

pros:

1. When docker has already built, we can always use the latest `spc` binary in docker.
2. We can always use the latest `src` directory in docker if we are using source mode.
3. Zero file modification.

cons:

1. If we call `--docker --help` in binary or phar mode, we will get non-color help message for `symfony/console`.
2. Additional command.

### DIFF
--- a/bin/spc
+++ b/bin/spc
@@ -6,9 +6,43 @@ use SPC\exception\ExceptionHandler;
 
 require_once __DIR__ . '/../vendor/autoload.php';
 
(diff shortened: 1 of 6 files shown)


crazywhalecc:
Currently based on musl-toolchain, we can compile normally on non-alpine, and using docker becomes a non-essential choice.

PR #203, opened 2023-09-23
Add tonyenc extension support

crazywhalecc:
Known issue:

- Not working with micro SAPI


crazywhalecc:
I decided to drop tonyenc support temporarily. This extension is currently not supported under phpmicro. Although it can be used normally under cli, the purpose of introducing tonyenc is to combine with phpmicro to distribute binaries.

PR #211, opened 2023-09-30
Throw proper exception when PHP source is not available

stloyd:


crazywhalecc:
It is indeed better to have a thrown exception, and we do need to make some changes here.

Because I didn't know where to put this code before and I wrote twice, here is a duplicate code here: https://github.com/crazywhalecc/static-php-cli/blob/3b83c1fa7b533aaade1b4368dab0dd45ad4d4f9e/src/SPC/util/Util.php#L12

It would be better if you could remove it in this PR and change the libraries and extensions that use this method to `$this->builder->getPHPVersionID()`.

crazywhalecc (approved):

PR #214, opened 2023-10-01
reference pr that shall not be merged

### LINKED ISSUE #94: libxml2 with icu errors
When libxml2 depends on icu, other libraries relying on libxml2 will report an error. Caused by 3745dfc931fbb188eb355f64d02a7be17446481f

### LINKED ISSUE #148: imagick not working on Linux
```
/gcc/x86_64-alpine-linux-musl/13.1.1/../../../../x86_64-alpine-linux-musl/bin/ld: ext/imagick/imagick.o: in function `zm_shutdown_imagick':
/app/source/php-src/ext/imagick/imagick.c:1209:(.text+0x5c1): undefined reference to `omp_pause_resource_all'
collect2: error: ld returned 1 exit status
```

### LINKED ISSUE #212: C++ extension incompatibility with linux
What's the reason for C++ extensions not being compatible with linux, but working on alpine? Surely this is something we can fix? My C knowledge may be rather limited but I'm fairly familiar with C++ build systems.

DubbleClick:
relies on imap to be merged first

closes #212 
closes #148 

## Summary

The following changes have been made to this PR, checked is tested.

- [ ] Add `ext-zip` dependency.
- [x] Simplify `cpp-extension` detection.
- [x] Remove redundant `LibraryBase::tryBuild()` method and add patching status.
- [ ] macOS library `libiconv` use `.a` directly for linking. (Why?)
- [ ] Add imap extension support.
- [ ] Use cross-make toolchain for non-alpine Linux.
- [ ] After building embed for Linux, replace `php-config` wrong order.
- [ ] Use `-fPIE` for embed SAPI. (Improve `libphp` security?)
- [x] Adjust Linux distro detection.
- [ ] Fix icu C++ support for Linux (Here I would rather not hard-code the path to `libstdc++.a`).
- [ ] Fix #94 (libxml2 with icu support, please write some comments).
- [ ] Add library `imap` support for Linux. (It is recommended to separate the mac and Linux build methods)
- [ ] Add `LIBS` for library `ldap`. (Maybe we could use pkg-config to generate it in the future)
- [ ] Add library libwebp patches for `pkg-config`.
- [ ] Fix library libzip `pkg-config` libs order.
- [ ] Fix C++ support for library postgresql.
- [ ] Add suggested library `ldap`, `libpam` for postgresql. (Bug: `getLib('pam')` -> `getLib('libpam')`)
- [x] Make `--with-clean` working for `build` command.
- [x] Add `patchFile` method.


src/SPC/builder/linux/library/libxml2.php
```diff
@@ -34,14 +34,16 @@ public function build(): void
...
                 '-DLIBXML2_WITH_PYTHON=OFF ' .
                 '-DLIBXML2_WITH_PROGRAMS=OFF ' .
                 '-DLIBXML2_WITH_TESTS=OFF ' .
                 '..'
             )
             ->exec("cmake --build . -j {$this->builder->concurrency}")
-            ->exec('make install');
+            ->exec('make install')
+            ->exec('mv ' . BUILD_ROOT_PATH . '/lib64/pkgconfig/* ' . BUILD_ROOT_PATH . '/lib/pkgconfig/')
+            ->exec('mv ' . BUILD_ROOT_PATH . '/lib64/*.a ' . BUILD_ROOT_PATH . '/lib/');
```
DubbleClick, line 47: the c++ libraries with cmake build into /lib64, need to copy it over

src/SPC/builder/unix/library/snappy.php
```diff
@@ -29,6 +29,8 @@ protected function build(): void
                 '../..'
             )
             ->exec("cmake --build . -j {$this->builder->concurrency}")
-            ->exec('make install');
+            ->exec('make install')
+            ->exec('mv ' . BUILD_ROOT_PATH . '/lib64/pkgconfig/* ' . BUILD_ROOT_PATH . '/lib/pkgconfig/')
+            ->exec('mv ' . BUILD_ROOT_PATH . '/lib64/*.a ' . BUILD_ROOT_PATH . '/lib/');
```
DubbleClick, line 34: likewise here

src/SPC/command/DownloadCommand.php
```diff
@@ -28,7 +28,7 @@ public function configure(): void
         $this->addArgument('sources', InputArgument::REQUIRED, 'The sources will be compiled, comma separated');
         $this->addOption('shallow-clone', null, null, 'Clone shallow');
         $this->addOption('with-openssl11', null, null, 'Use openssl 1.1');
-        $this->addOption('with-php', null, InputOption::VALUE_REQUIRED, 'version in major.minor format like 8.1', '8.1');
+        $this->addOption('with-php', null, InputOption::VALUE_REQUIRED, 'version in major.minor format (default 8.2)', '8.2');
```
DubbleClick, line 31: I think for the 2.0 release we should default to php 8.2 (if not 8.3)

src/SPC/doctor/item/LinuxMuslCheck.php
```diff
@@ -8,54 +8,77 @@
...
+            'url' => 'https://github.com/richfelker/musl-cross-make',
+            'rev' => 'master',
+        ];
+        Downloader::downloadSource('musl-cross-make', $musl_cross_compile_source);
+        Downloader::downloadSource('musl-1.2.4', $musl_source);
+        FileSystem::extractSource('musl-1.2.4', DOWNLOAD_PATH . '/musl-1.2.4.tar.gz');
+        $arch = arch2gnu(php_uname('m')) === 'x86_64' ? 'x86_64-linux-musl' : 'aarch64-linux-musl';
+        $install_musl_wrapper_cmd = 'cd ' . DOWNLOAD_PATH . '/musl-cross-make && \
+                       make install TARGET=' . $arch . ' OUTPUT=/usr/local/musl -j';
+        $musl_install = 'cd ' . SOURCE_PATH . '/musl-1.2.4 && \
```
DubbleClick, line 61: we need to install both, musl-cross-make for c++ support, musl-gcc to be able to run dynamically linked binaries built with musl-cross-make

src/SPC/builder/linux/LinuxBuilder.php
```diff
@@ -185,8 +188,10 @@ public function buildPHP(int $build_target = BUILD_TARGET_NONE): void
         $enableMicro = ($build_target & BUILD_TARGET_MICRO) === BUILD_TARGET_MICRO;
         $enableEmbed = ($build_target & BUILD_TARGET_EMBED) === BUILD_TARGET_EMBED;
 
+        $arch = arch2gnu(php_uname('m')) === 'x86_64' ? 'x86_64-linux-musl' : 'aarch64-linux-musl';
         shell()->cd(SOURCE_PATH . '/php-src')
             ->exec(
+                'LD_LIBRARY_PATH=/usr/local/musl/' . $arch . '/lib ' .
```
DubbleClick, line 194: it tries loading shared memory to test opcache

src/SPC/builder/linux/library/icu.php
```diff
@@ -28,7 +29,7 @@ protected function build(): void
                 "--prefix={$root}"
             )
             ->exec('make clean')
-            ->exec("make -j{$this->builder->concurrency}")
+            ->exec("LD_LIBRARY_PATH=/usr/local/musl/{$arch}/lib make -j{$this->builder->concurrency}")
```
DubbleClick, line 32: likewise, tries to load shared lib for some reason

DubbleClick:
I'm not 100% sure yet if *dynamic* binaries built by this will be usable on systems that do not have musl-gcc installed. fully *static* binaries definitely work.

the reason why this might matter is opcache. I will test if opcache works tomorrow on a vm where I don't have musl installed.

crazywhalecc:
I haven't gotten home yet. It seems that it is feasible to implement C++ using cross-compilation, but I am not very clear about it. It seems that just replacing CXX is enough?

DubbleClick:
It's not really cross compilation (we could do that too, but no need). It essentially compiles gcc/g++ targeting musl libc/libstdc++. Then we use these compilers to compile libs and php. It's essentially what happens on Alpine, except that the pre-installed ones there already target musl.

DubbleClick:
last commit closes #215 #216

crazywhalecc:
It seems that this PR is based on the PR of imap. I will check the imap part first and then merge this.

DubbleClick:
both x86_64 and arm64 work on RHEL based systems now

still running into some trouble building musl-cross-make on debian (x86)

DubbleClick:
todo: alpine support, macos support (can't do myself)

DubbleClick:
wow this is extremely annoying, using the prebuilt musl-cross-make from musl.cc results in intl extension causing a segfault. compiling musl-cross-make myself it works just fine - but we cannot compile musl-cross-make ourselves on debian

crazywhalecc:
Okay.... The current PR is too large. If not necessary, maybe some insignificant updates can be separated.

Or you can write clearly what each modification does in the first comment of the PR.

composer.json
```diff
@@ -12,6 +12,7 @@
         "php": ">= 8.1",
         "ext-mbstring": "*",
         "ext-pcntl": "*",
+        "ext-zip": "*",
```
crazywhalecc, line 15: Where is the `zip` extension needed?

crazywhalecc:
I feel like testing all of these features at once in a PR would be disastrous, but taking them apart would also be a huge amount of work.

My idea is that this PR can be kept first. For some pre-functions that can work stably (such as doctor, cross-make-gcc), start from the main branch and make specific modifications. Then I will quickly review and merge it into main branch, and then merge the main branch into the current branch to reduce the number of files modified by this branch.

crazywhalecc:
I merged some general small fixes into the main branch first to reduce the amount of file changes in this branch.

composer.json
DubbleClick: Extracting some sources failed for me without the zip extension installed. Let me make sure this is actually the case or if it was just a random fail.

DubbleClick:
> I feel like testing all of these features at once in a PR would be disastrous, but taking them apart would also be a huge amount of work.
> 
> My idea is that this PR can be kept first. For some pre-functions that can work stably (such as doctor, cross-make-gcc), start from the main branch and make specific modifications. Then I will quickly review and merge it into main branch, and then merge the main branch into the current branch to reduce the number of files modified by this branch.

I agree, I will pick it apart into a bunch of separate branches as well as possible once everything is working. I'm not yet sure how to fix the musl-compiler issue. Worst case we can't support intl on Debian yet.

composer.json
crazywhalecc: Currently, only the `unzip` command is used, and PHP's zip extension is not used. It may be a related problem caused by the downloaded file being damaged.

DubbleClick:
@crazywhalecc what do you think is the best course of action for the musl compilers - alpine always stays the same and directly uses gcc / g++.

A) Debian + RHEL: Use prebuilt musl-native toolchains from https://musl.cc - problem: intl extension causes a segfault that I do not understand.
B) RHEL: Built musl-cross-make from https://github.com/richfelker/musl-cross-make - everything works. Debian: Use prebuilt musl-native toolchain from musl.cc - intl doesn't work. - Cannot manage to build musl-cross-make on Debian.
C) Upload and use a gcc 11.2.0 version of musl-cross-make that we compiled ourselves - works on both RHEL and Debian.
D) Do not support c++ extensions on debian.
E) Drop Debian support entirely (also simplifies imap since rhel/alpine builds do not require libpam).
F) Only support c++ extensions build on alpine (what we have now).

crazywhalecc:
I also want to introduce C++ compilation support in Debian and RHEL, but if it is difficult or complicated, I think it is better to "only support compilation of C++ extensions under Alpine".

Because the original `bash-version` is somewhat similar to the current version, which is to use the system's compilation tool chain as directly as possible. Compilation of these system-level libraries will first produce certain changes in the system itself. I certainly hope that future `spc` binaries can run on as many distributions as possible, but the current situation is that the distributions are very different, and it is impossible to make every distribution support 100% of all compilation features.

But don't give up, I think if `cross-make` can compile most extensions under RHEL and Debian, and the `cross-make` package can be easily installed, we can document it as an optional feature, or adding an additional compile command.

DubbleClick:
C++ support under RHEL and Debian is working with the musl toolchains. Just intl extension segfaults when using the pre-built versions from musl.cc - maybe I can compile a version that we host somewhere that spc can download and use?

crazywhalecc:
Is your idea to pre-compile the Linux musl binary and related libraries, and then let spc download them directly for use?

I don't know much about your local progress. If they can be compiled successfully on a certain Linux distribution and they can be transplanted to other distributions to run normally, it may be a good choice. If the musl related source code is mirrored on GitHub, we can fork it and build it in Actions, and then use spc's Downloader to download it from the Release file. 

These are the solutions I can think of so far.

DubbleClick:
Yes, we can compile musl-cross-make, tar the produced output and have spc download it. Compiling musl-cross-make from source doesn't work on Debian (I do not know why, but after manually fixing like 10 different errors I gave up). But Debian can use the downloaded musl toolchain.

I don't think there is much point in compiling musl-cross-make in GitHub actions, though. Just compile it once, host it somewhere and let spc download it. It currently already does this (see MuslFix)

DubbleClick:
Found the issue, intl doesn't work when gcc was configured with --enable-default-pie. I created a bug report in the php repo.

In the meantime, I've compiled a x86 toolchain without it that we can use. Will compile one for arm too.

DubbleClick:
I will pick this abomination of a branch into different pr's now

DubbleClick:
first pr with only the c++ support is done

I took a different approach to fixing the lib64 problem this time, so we don't have to copy or check in every individual extension that may build to lib64 rather than lib.

composer.json
DubbleClick: Figured it out.  On RHEL based distros, composer automatically installs ext-zip. If you uninstall it, it breaks composer autoloading. It's not a direct dependency of spc though, so I removed it.

crazywhalecc:
If a new imap branch is opened separately, this branch should be closed, right?

PR #216, opened 2023-10-02
Fix pgsql linker cpp

jingjingxyk:
当pgsql 依赖 icu 时 需要链接 stdc++

### DIFF
--- a/src/SPC/builder/unix/library/postgresql.php
+++ b/src/SPC/builder/unix/library/postgresql.php
@@ -36,7 +36,8 @@ trait postgresql
         $output = shell()->execWithResult($env . " {$pkgconfig_executable} --libs-only-l --static " . $packages);
         if (!empty($output[1][0])) {
             $libs = $output[1][0];
-            $envs .= " LIBS=\"{$libs}\" ";
+            $libcpp = $this instanceof MacOSLibraryBase ? ' -lc++' : ' -lstdc++ ';
+            $envs .= " LIBS=\"{$libs} {$libcpp} \" ";
         }
 
         FileSystem::resetDir($this->source_dir . '/build');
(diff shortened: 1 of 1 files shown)


DubbleClick:
also need to patch the libxml-2.0.pc required libraries to change the order, I integrated this into #214

crazywhalecc:
postgresql library 在依赖 icu 时，编译 postgresql 自身库时候也会走 linker 吗？我记得如果只生成库不生成二进制的话是不会在编译库时候走 linker 的。如果是这样，我觉得可以加一个 `cpp-library = true` 类似的东西，将一部分 C++ 的依赖转移到 `lib.json` 中。

jingjingxyk:
生成的二进制文件是这个： pg_config

DubbleClick:
> postgresql library 在依赖 icu 时，编译 postgresql 自身库时候也会走 linker 吗？我记得如果只生成库不生成二进制的话是不会在编译库时候走 linker 的。如果是这样，我觉得可以加一个 `cpp-library = true` 类似的东西，将一部分 C++ 的依赖转移到 `lib.json` 中。但如果是自己要生成二进制，那就没办法了，只能这么 patch 一下。

Yes, when you compile php with phsql + icu it links these parts. When you only use intl (icu) and build the `embed` SAPI, a program using the embed API must  link with -lstdc++ itself.

I have no idea how we could make it so that necessary libstdc++ symbols always get linked into the libraries. It's the same thing with openssl & zlib. Openssl only needs the zlib lib to compile, but the necessary zlib symbols only get linked into php when the zlib extension is used too. It would be great if we could force linking of all dependencies into the libs we produce.

jingjingxyk:
use  `CPPFLAGS` `LDFLAGS` `LIBS`  

example : 

https://php-cli.jingjingxyk.com/extra-tools/make-env.sh
https://php-cli.jingjingxyk.com/extra-tools/make-export-variables.sh

crazywhalecc:
已在 #231 中解决。

PR #230, opened 2023-10-15
Add FreeBSD support

crazywhalecc:
If we just implement a standalone php binary in FreeBSD, it is very simple (I only took 2 hours to implement). FreeBSD support is now available with only a few modifications.

The current compilation result is not statically linked, but only `libc.so`, `librt.so`, etc. included by FreeBSD by default are linked. Tested (compiled from 13, run on 12) it's perfect.

I also tried compiling a fully static binary using `-all-static`, and the result of compiling it was very strange: `file php` showed `dynamically linked`, and running it or `ldd php` showed a segfault. Maybe it has something to do with `clang/llvm` and the underlying ELF parsing mode of FreeBSD, but there is very, very little information about this on the Internet.

Documentation PR: https://github.com/crazywhalecc/static-php-cli-docs/pull/2

### DIFF
--- a/README-en.md
+++ b/README-en.md
@@ -8,10 +8,12 @@ Compile a purely static php-cli binary file with various extensions to make PHP 
 
 <img width="600" alt="2023-05-02 15 53 13" src="https://user-images.githubusercontent.com/20330940/235610282-23e58d68-bd35-4092-8465-171cff2d5ba8.png">
 
-You can also use the micro binary file to combine php binary and php source code into one for distribution!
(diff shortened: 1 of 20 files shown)


crazywhalecc:
@kdkd Now we will have FreeBSD support 🎉

kdkd:
This is amazing, I will give this a test as soon as I can!

crazywhalecc:
@kdkd I haven't added some extensions that rely on external libraries, so there are currently a lot of extensions that are not supported. But supporting them should be easy, unless some libraries are not supported when building in FreeBSD.

Issue #217, opened 2023-10-03
Compile failed on Linux(aarch64) with fpm SAPI


crazywhalecc:
## Build Command

```bash
bin/spc build bcmath --debug --build-fpm
```

## Console Log

```
/root/static-php-cli/source/php-src/sapi/fpm/fpm/fpm_atomic.h:136:2: error: #error Unsupported processor. Please open a bug report (bugs.php.net).
  136 | #error Unsupported processor. Please open a bug report (bugs.php.net).
      |  ^~~~~
/root/static-php-cli/source/php-src/sapi/fpm/fpm/fpm_atomic.h:140:32: error: unknown type name ‘atomic_t’
  140 | static inline int fpm_spinlock(atomic_t *lock, int try_once) /* {{{ */
      |                                ^~~~~~~~
In file included from /root/static-php-cli/source/php-src/sapi/fpm/fpm/fpm.c:19:
/root/static-php-cli/source/php-src/sapi/fpm/fpm/fpm_scoreboard.h:23:17: error: unknown type name ‘atomic_t’
   23 |                 atomic_t lock;
      |                 ^~~~~~~~
/root/static-php-cli/source/php-src/sapi/fpm/fpm/fpm_scoreboard.h:52:17: error: unknown type name ‘atomic_t’
   52 |                 atomic_t lock;
      |                 ^~~~~~~~
make: *** [Makefile:393: sapi/fpm/fpm/fpm.lo] Error 1
make: *** Waiting for unfinished jobs....
In file included from /root/static-php-cli/source/php-src/sapi/fpm/fpm/fpm_scoreboard.h:13,
                 from /root/static-php-cli/source/php-src/sapi/fpm/fpm/fpm_children.c:26:
/root/static-php-cli/source/php-src/sapi/fpm/fpm/fpm_atomic.h:136:2: error: #error Unsupported processor. Please open a bug report (bugs.php.net).
  136 | #error Unsupported processor. Please open a bug report (bugs.php.net).
      |  ^~~~~
/root/static-php-cli/source/php-src/sapi/fpm/fpm/fpm_atomic.h:140:32: error: unknown type name ‘atomic_t’
  140 | static inline int fpm_spinlock(atomic_t *lock, int try_once) /* {{{ */
      |                                ^~~~~~~~
In file included from /root/static-php-cli/source/php-src/sapi/fpm/fpm/fpm_children.c:26:
[...]

PR #231, opened 2023-10-16
Linux c++ support using musl-cross-make toolchains

DubbleClick:
This PR will

closes https://github.com/crazywhalecc/static-php-cli/issues/148
closes https://github.com/crazywhalecc/static-php-cli/issues/212
closes https://github.com/crazywhalecc/static-php-cli/issues/217
closes https://github.com/crazywhalecc/static-php-cli/issues/234
closes https://github.com/crazywhalecc/static-php-cli/issues/94

## Tasks

- [X] Check #148 
- [X] Check #217 
- [X] Check #94 
- [X] Check build common extensions on Linux x86_64
- [X] Check build common extensions on Linux aarch64
- [ ] Update docs related to this PR (create another PR)


src/SPC/doctor/item/LinuxMuslCheck.php
```diff
@@ -8,57 +8,122 @@
...
+     */
+    public function installMuslCrossMake(): void
+    {
+        $arch = arch2gnu(php_uname('m'));
+        $musl_compile_source = [
+            'type' => 'url',
+            'url' => "https://musl.cc/{$arch}-linux-musl-cross.tgz",
+        ];
+        Downloader::downloadSource('musl-compile', $musl_compile_source);
+        FileSystem::extractSource('musl-compile', DOWNLOAD_PATH . "/{$arch}-linux-musl-cross.tgz");
```
DubbleClick, line 126: We should replace this with our own self-hosted pre-compiled version, like at https://github.com/DubbleClick/static-php-cli/releases/tag/tools but tagged on this repository or something.

For one, because musl.cc is not an official musl website, but second, because the version there is compiled with --enable-default-pie which means intl segfaults. This is not the case with the version I built myself.

src/SPC/doctor/item/LinuxMuslCheck.php
```diff
@@ -8,57 +8,122 @@
...
-            return true;
-        } catch (RuntimeException) {
-            return false;
-        }
+        $musl_source = [
+            'type' => 'url',
+            'url' => 'https://musl.libc.org/releases/musl-1.2.4.tar.gz',
+        ];
+        Downloader::downloadSource('musl-1.2.4', $musl_source);
+        FileSystem::extractSource('musl-1.2.4', DOWNLOAD_PATH . '/musl-1.2.4.tar.gz');
```
DubbleClick, line 106: Sounds counter-intuitive, but we should keep this version hard-coded as it should match the version that the musl-cross-make was built with!

Maybe add a comment there?

DubbleClick:
I will see to fix pgsql compilation and add imap back once this is merged, so it doesn't become a giant PR again. No point creating a PR doing these things now and changing them later when this is merged.

crazywhalecc:
Is there any difference between `/lib` and `/lib64`? The distributions I have seen so far (64-bit) do not distinguish between the two and put all libraries in `/lib`.

src/SPC/doctor/item/LinuxMuslCheck.php
crazywhalecc: I can host these files on dl.static-php.dev. But I think it's better to build it directly in Actions and automatically upload to server, similar to [static-php-cli-hosted](https://github.com/crazywhalecc/static-php-cli-hosted).

src/SPC/doctor/item/LinuxMuslCheck.php
```diff
@@ -8,57 +8,122 @@
...
+    public function installMuslCrossMake(): void
+    {
+        $arch = arch2gnu(php_uname('m'));
+        $musl_compile_source = [
+            'type' => 'url',
+            'url' => "https://musl.cc/{$arch}-linux-musl-cross.tgz",
+        ];
+        Downloader::downloadSource('musl-compile', $musl_compile_source);
+        FileSystem::extractSource('musl-compile', DOWNLOAD_PATH . "/{$arch}-linux-musl-cross.tgz");
+        shell(true)->exec('cp -rf ' . SOURCE_PATH . '/musl-compile/* /usr/local/musl');
```
crazywhalecc, line 127: Is it possible to install cross-make musl in user directory?

DubbleClick: That's the default path for it. We could technically put it anywhere as long as we add the bin folder to the $PATH

src/SPC/command/BuildCliCommand.php
```diff
@@ -29,6 +29,7 @@ public function configure(): void
         $this->addOption('build-all', null, null, 'build cli, micro, fpm, embed');
         $this->addOption('no-strip', null, null, 'build without strip, in order to debug and load external extensions');
         $this->addOption('enable-zts', null, null, 'enable ZTS support');
+        $this->addOption('disable-opcache-jit', null, null, 'disable opcache jit');
```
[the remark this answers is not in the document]
DubbleClick: I don't think it should be the default, an option is fine.

I only stumbled upon the issue with building FrankenPHP on Alpine with some extension combinations. Most work fine without disabling.

Funny enough the exact same extension combinations worked without disabling opcache on RHEL.

DubbleClick:
> Is there any difference between `/lib` and `/lib64`? The distributions I have seen so far (64-bit) do not distinguish between the two and put all libraries in `/lib`.

Some libraries build into /lib64 for me. I don't know why.

src/SPC/doctor/item/LinuxMuslCheck.php
DubbleClick: You can do that if you want, but I don't see much point in rebuilding it over and over. We would have to maintain patches for the repository ourselves because the main repo rarely gets updated and new gcc/linux headers/etc versions are not used by default.

This will very rarely update and mostly just be a big waste of time.

DubbleClick: I see, you'll have to manually edit the workflow to update the versions used, but I will post a step-by-step how to build them (on Alpine, RHEL may be overkill for an action runner) tomorrow.

crazywhalecc:
In the past, I have tried to avoid compiling the toolchains necessary for the compilation itself in user mode. Because whether you use a downloaded binary and add it to PATH, or manually compile a gcc, they will consume a lot of time and disk space. Using package management can speed things up, especially in Alpine environments where they can be installed directly.

However, if musl-cross-make is stable and feasible now, we could add this pre-dependent library when detecting non-Alpine environments, which will only extend the time for compiling PHP in these distributions. At this time it is more appropriate to become the latter, but we also need to consider compatibility with other libraries (so this will be another big project)

crazywhalecc:
But anyway, I don't want to make this very complicated. Maybe it would be cool to support purely static compilation for glibc-based distributions, but I don't want to have to make extensive changes to every library that already implements the compilation command. The final impact on the compiler is best implemented only in `LinuxBuilder.php`.

DubbleClick:
> In the past, I have tried to avoid compiling the toolchains necessary for the compilation itself in user mode. Because whether you use a downloaded binary and add it to PATH, or manually compile a gcc, they will consume a lot of time and disk space and it is intolerable. Using package management can speed things up, especially in Alpine environments where they can be installed directly.
> 
> However, if musl-cross-make is stable and feasible now, we could add this pre-dependent library when detecting non-Alpine environments, which will only extend the time for compiling PHP in these distributions. At this time it is more appropriate to become the latter, but we also need to consider compatibility with other libraries (so this will be another big project)

I don't plan to have everyone compile musl-cross-make themselves, downloading the archive you see in MuslFix takes like 5 seconds. And it doesn't happen on Alpine, only on glibc based distros.

crazywhalecc:
I successfully built the C++ extension on Debian using `https://dl.static-php.dev/static-php-cli/deps/musl-toolchain/` and it works just fine.

I will modify some logic tomorrow:

1. By default, musl-toolchain will become optional and will not be downloaded from the doctor. When the current distribution is not alpine, you can use the parameter `--with-musl-cross` to build C++ extensions in Debian/RHEL.
2. The Debian series uses musl-gcc installed through package management instead of compilation. It is only compiled in environments such as RHEL that do not specify packages.
3. Modify some environment parameters to `putenv()` to simplify the compilation command. Because the commands called through `shell()` are all child processes of the current php, env variables can be passed to the child process.

DubbleClick:
> 1. By default, musl-toolchain will become optional and will not be downloaded from the doctor. When the current distribution is not alpine, you can use the parameter `--with-musl-cross` to build C++ extensions in Debian/RHEL.

I do not like this. It's more to maintain for us (musl-gcc wrapper AND $arch-linux-musl).

> 2. The Debian series uses musl-gcc installed through package management instead of compilation. It is only compiled in environments such as RHEL that do not specify packages.

Bad idea, we have no control over which musl-gcc wrapper version it will install and if our $arch-linux-musl is compatible with it. Musl *tries* to be future compatible, but it's hard to guarantee it.

> 3. Modify some environment parameters to `putenv()` to simplify the compilation command. Because the commands called through `shell()` are all child processes of the current php, env variables can be passed to the child process.

Good idea!

> 4. Try to build aarch64 musl-cross-toolchain (but my mac is MacBook Air M2 16GB, small RAM 😢 ).

I built it no problem on a 4 core 8 gig ram aarch64 azure VM with Alma and Alpine. The RAM requirement seems to only be true for Docker compilation.

crazywhalecc:
> I do not like this. It's more to maintain for us (musl-gcc wrapper AND $arch-linux-musl).

I mainly consider that I do not want the compilation process to appear in the doctor. If I only choose to maintain one set, I can detect the system environment during build and compile and install them by adding pre-dependent libraries.

In addition, if you think you can only maintain one set, it seems that you only need to download musl-cross by default and unzip it to buildroot when building.

crazywhalecc:
> I built it no problem on a 4 core 16 gig ram aarch64 azure VM with Alma and Alpine. The RAM requirement seems to only be true for Docker compilation. Exact model: Standard D4ps v5 (4 vcpus, 16 GiB memory)
If you run into any errors, just relaunch the make -j command until it works (yes, seriously lol, I used it like 5x and then it built).

This is strange. When I compile using a virtual machine (Alpine aarch64, 8G) on my Mac, it will crash when the memory is exhausted.

DubbleClick:
> I mainly consider that I do not want the compilation process to appear in the doctor. If I only choose to maintain one set, I can detect the system environment during build and compile and install them by adding pre-dependent libraries.

You mean the compilation of musl-libc (which also installs musl-gcc wrapper)? I do not think that's optional, if we want to use the musl toolchain we must do it. But it's super stable and very fast, no matter the system.
 
> In addition, if you think you can only maintain one set, it seems that we could only need to download musl-cross by default and unzip it to buildroot when building.

Musl-libc = installs musl system libraries (+ musl-gcc wrapper).
Musl-cross-make = Compiler toolchain targeting musl-libc
If we used MUSL_VER=1.2.4 in `config.mak` we should also install version 1.2.4 of musl-gcc.

Musl-cross-make should not be in buildroot. We should either download it in doctor and extract it to `./source` or extract it to /usr/local/musl, where musl-libc installs.

crazywhalecc:
My local testing found that cross-make can also work normally under buildroot. I just specify the `cc`, `ar` and other parameters you added in the cplus branch under `buildroot/musl/xxx`, and finally it can be compiled correctly.

DubbleClick:
> My local testing found that cross-make can also work normally under buildroot. I just specify the `cc`, `ar` and other parameters you added in the cplus branch under `buildroot/musl/xxx`, and finally it can be compiled correctly.

Yes it works, but there is no reason for it to be there. It's not really output of our build, it's a dependency we use to build.

crazywhalecc:
Suddenly a strange idea came to me. If we use `putenv()` as the variable setting of the compilation environment, can we put these compiled files into a user-mode directory?

I kind of don't want to write these non-package management installed things to the system directory.

src/SPC/doctor/item/LinuxMuslCheck.php
```diff
@@ -98,32 +80,36 @@ public function installMuslWrapper(): void
...
-        Downloader::downloadSource('musl-1.2.4', $musl_source);
-        FileSystem::extractSource('musl-1.2.4', DOWNLOAD_PATH . '/musl-1.2.4.tar.gz');
-        $musl_gcc_install_cmd = 'cd ' . SOURCE_PATH . '/musl-1.2.4 && \
-                         ./configure --enable-wrapper=gcc && \
-                         make -j && make install';
-        shell(true)->exec($prefix . $musl_gcc_install_cmd);
+        Downloader::downloadSource($musl_version_name, $musl_source);
+        FileSystem::extractSource($musl_version_name, DOWNLOAD_PATH . "/{$musl_version_name}.tar.gz");
+        shell(true)->cd(SOURCE_PATH . "/{$musl_version_name}")
+            ->exec('./configure --enable-wrapper=gcc')
```
DubbleClick, line 91: I think we can remove the --enable-wrapper-gcc now.

src/SPC/doctor/item/LinuxMuslCheck.php
```diff
@@ -98,32 +80,36 @@ public function installMuslWrapper(): void
...
+     * @throws RuntimeException
+     * @throws WrongUsageException
      */
     public function installMuslCrossMake(): void
     {
         $arch = arch2gnu(php_uname('m'));
         $musl_compile_source = [
             'type' => 'url',
-            'url' => "https://musl.cc/{$arch}-linux-musl-cross.tgz",
+            'url' => "https://dl.static-php.dev/static-php-cli/deps/musl-toolchain/{$arch}-musl-toolchain.tgz",
```
DubbleClick, line 107: 🎉

DubbleClick:
> Suddenly a strange idea came to me. If we use `putenv()` as the variable setting of the compilation environment, can we put these compiled files into a user-mode directory?
> 
> I kind of don't want to write these non-package management installed things to the system directory.

Technically, yes. But then we must also put LIBRARY_PATH and LD_LIBRARY_PATH to the non-system directory. The default search for musl-libc libraries is /usr/local/musl.

crazywhalecc:
I'm finishing up using `getenv` and `putenv` to replace `$builder->configure_env`, which requires modifying a lot of files.

crazywhalecc:
I've tested almost all libraries with `bin/spc build:libs pkg-config,brotli,bzip2,curl,freetype,gmp,icu,imagemagick,ldap,libavif,libevent,libffi,libiconv,libjpeg,libpng,libsodium,libssh2,libwebp,libxml2,libxslt,libyaml,libzip,ncurses,nghttp2,onig,openssl,postgresql,readline,snappy,sqlite,xz,zlib,zstd` for aarch64 linux and macOS.

crazywhalecc:
This PR will

closes #148 
closes #212 
closes #217 
closes #234 
closes #94 

🤯

src/SPC/command/DownloadCommand.php
```diff
@@ -59,105 +59,110 @@ public function initialize(InputInterface $input, OutputInterface $output): void
...
-        // Define PHP major version
-        $ver = $this->php_major_ver = $this->getOption('with-php') ?? '8.1';
-        define('SPC_BUILD_PHP_VERSION', $ver);
-        preg_match('/^\d+\.\d+$/', $ver, $matches);
-        if (!$matches) {
-            logger()->error("bad version arg: {$ver}, x.y required!");
-            return static::FAILURE;
-        }
+            // Define PHP major version
+            $ver = $this->php_major_ver = $this->getOption('with-php') ?? '8.1';
```
DubbleClick, line 88: 8.2 please!

crazywhalecc: The default value for the `--with-php` option is already 8.2, of course, this was just a missed update.

DubbleClick: ah yeah can just remove the ?? then

crazywhalecc:
When I was testing each extension, I encountered a very strange issue that prevented me from compiling most extensions after including a C++ extension. Upon checking the `php-src/config.log` file, it seems there are namespace-related errors during compilation.
[config.log](https://github.com/crazywhalecc/static-php-cli/files/13071730/config.log)

crazywhalecc:
I feel that including C++ extensions in PHP is not a problem, but if a C library depends on a C++ library and requires the inclusion of LIBS=-lstdc++ during compilation, the final pkg-config will then record this symbol, potentially causing this confusion during PHP compilation. Generating binary C++ libraries or considering dependencies on C++ libraries when generating binary C libraries should be done carefully.

src/SPC/builder/unix/library/bzip2.php
```diff
@@ -8,9 +8,14 @@ trait bzip2
...
     {
+        $cc = getenv('CC');
+        $ar = getenv('AR');
+        if ($ar === false) {
+            $ar = 'ar';
+        }
         shell()->cd($this->source_dir)
             ->exec("make PREFIX='" . BUILD_ROOT_PATH . "' clean")
-            ->exec("make -j{$this->builder->concurrency} PREFIX='" . BUILD_ROOT_PATH . "' libbz2.a")
+            ->exec("make -j{$this->builder->concurrency} PREFIX='" . BUILD_ROOT_PATH . "' CC={$cc} AR={$ar} libbz2.a")
```
DubbleClick, line 18: does this actually change anything? it should use the environment variables anyway

crazywhalecc: Yes, it's a bit strange. It should have used env instead. But if I don't specify CC, the compilation result will have a linking error at the end, indicating `undefined reference fopen64`. When I looked for the corresponding source code in the bz code, I found that line of code is fopen().

DubbleClick: Yeah the ancient bzip2 1.0x source (which we use) defines the variables without checking the environment, so we need to pass them with the command line.

I will create a helper method to build the correct string for this.

crazywhalecc:
What are your thoughts on compiling C++ libraries and C++ extensions? Tried many times and I'm still not sure about it. In the compilation of libraries, if libstdc++ is included, it will cause PHP's configure to ultimately include lstdc++, which leads to some checks failing to pass.

DubbleClick:
> What are your thoughts on compiling C++ libraries and C++ extensions? Tried many times and I'm still not sure about it. In the compilation of libraries, if libstdc++ is included, it will cause PHP's configure to ultimately include lstdc++, which leads to some checks failing to pass.

That's what the ld_library_path is for, so that it can load libstdc++ during compilation tests.

src/SPC/builder/linux/LinuxBuilder.php
```diff
@@ -176,6 +176,7 @@ public function buildPHP(int $build_target = BUILD_TARGET_NONE): void
 
         shell()->cd(SOURCE_PATH . '/php-src')
             ->exec(
+                "{$this->getOption('ld_library_path')} " .
```
DubbleClick, line 179: this right here

./configure checks with dynamic compilation so it must be able to load shared system libraries at the time. /usr/local/musl/arch-linux-musl/lib is not in the normal library load paths so we have to specify it.

during actual compilation, libstdc++.a is used, so it still produces a fully static binary

crazywhalecc: image
image

The latest code under the current branch, I compiled `bin/spc build --build-all --debug apcu,bcmath,bz2,calendar,ctype,curl,dba,dom,event,exif,filter,fileinfo,ftp,gd ,gmp,iconv,imagick,inotify,intl,ldap,mbregex,mbstring,memcache,mongodb,mysqli,mysqlnd,openssl,pcntl,pdo,pdo_mysql,pdo_pgsql,pdo_sqlite,pgsql,phar,posix,protobuf,readline,redis,session ,shmop,simplexml,snappy,soap,sockets,sodium,sqlite3,sysvmsg,sysvsem,sysvshm,tokenizer,xlswriter,xml,xmlreader,xmlwriter,xsl,yaml,zip,zlib,zstd`, finally in PHP's `./configure `Still encountering this problem.

DubbleClick: md5_init should be in -lcrypto, so I'm not sure why it wouldn't find it. Have you tried `rm -rf source buildroot download` and restarting? It builds fine for me, except for imagick because you removed the `-lgomp` from extra_libs.

Will add that back again, then it should work.

DubbleClick: As I expected, builds fine on both RHEL and Alpine.

crazywhalecc: Yeah I just removed `source` and `buildroot` every time. I feel like this issue could be caused by a lot of different issues. On the surface, it may be that `pkg-config` does not include the `--static` option in PHP's `configure`.

But anyway, I always feel that there is a feasible solution that prevents us from having to replace `-lxxx` with `libxxx.a` before `configure`?

DubbleClick: `-L/path/to/libs -lexample` is the same as `/path/to/libs/libexample.a`

So instead of patching it in the ./configure command we can set LDFLAGS or EXTRA_LDFLAGS or LIBS to `-L/static-php-cli/buildroot/lib`

DubbleClick: OH, I think I found the problem for your config.log
```console
checking Check libxlsxwriter library... use the openssl md5 library
found in /usr
```

Shouldn't this use the one in BUILD_LIB_PATH?

DubbleClick: I think that's what the --with-openssl-dir option is for. Let me know if the next commit fixes the issue.

crazywhalecc: It seems to be just a problem with xlswriter extending its own compilation options, which can be solved alone. But when I removed xlswriter and continued compilation, the error `unknown type name 'namespace'` appeared again. This time it appears in the libzip test.

DubbleClick: What OS is this one? Debian x86?

crazywhalecc: Debian x86_64. My local code is a bit messy, I'll start with the cleanest one and download it again.

BTW I'm trying a simple replacement case: replace the `-lssh2` for `configure` in the ssh2 extension with a list of `.a` files. Then I discovered that it seemed that it did not use the `--static` in `pkg-config` in the `configure` stage to add `-lssl -lcrypto` and other dependencies.

DubbleClick: I'm not too experienced with pkg-config, cannot help there :/

I can test on Debian later, I tested RHEL & Alpine x86 and compilation with the extensions above worked.

DubbleClick: Ok, I get the same error message for xlswriter.
```console
checking system libxlsxwriter... no
checking system libxlsxio... no
checking openssl MD5... /opt/static-php-cli/buildroot
checking enable xlsx reader support... yes
checking Check libxlsxwriter library... use the openssl md5 library
found in /opt/static-php-cli/buildroot
checking for MD5_Init in -lcrypto... no
configure: error: Wrong openssl MD5_Init not found
```

src/SPC/doctor/item/LinuxMuslCheck.php
```diff
@@ -26,7 +26,7 @@ public function checkMusl(): CheckResult
         }
 
         $musl_wrapper_lib = sprintf('/lib/ld-musl-%s.so.1', php_uname('m'));
-        if (file_exists($musl_wrapper_lib)) {
+        if (file_exists($musl_wrapper_lib) && file_exists('/usr/local/musl/bin/musl-gcc')) {
```
DubbleClick, line 29: I think this is the one file we shouldn't check, because there's really no point in enabling the gcc wrapper with our install. I will change this.

crazywhalecc: Because I need to test the `/usr/local/musl` directory many times, I will delete this directory frequently and I must ensure that the directory exists.

src/SPC/builder/linux/LinuxBuilder.php
crazywhalecc: This xlswriter problem should create a new Issue to handle it.

crazywhalecc:
For compiling imagemagick under the macOS platform, I found a strange point: if I do not use the `LIBS=` variable to force the imagemagick library to be specified, a `not a mach-o file` error will appear; if not specified, let It looks for it automatically, it will end up with a link error of `undefined reference`. In the case of the latter, I found that the three libraries related to the error reported were all compiled using cmake, while the others compiled using autoconf+make did not appear here.

DubbleClick:
Hmm, I can't test MacOS compilation, sorry :(

crazywhalecc:
That's okay. I will figure it out.
发自我的 iPhone

在 2023年10月30日，03:12，DubbleClick ***@***.***> 写道：

﻿

Hmm, I can't test MacOS compilation, sorry :(

crazywhalecc:
I've tested the common extensions on Linux x86_64, as well as the compilation results of icu, imagemagick, postgresql, and ldap in cross-make. At the moment, all of these seem to be working well, and it's time to merge it. 🎉 

Additionally, there are two new issues that need to be created: 

1. A bug in xlswriter.
2. When using `--build-all`, the generated cli, fpm, and micro binaries include debugging symbols.

DubbleClick:
make it rc-8, I still want to add imap support after this for 2.0 release!

crazywhalecc (approved):

PR #238, opened 2023-10-17
Fix libphp requiring non-existent lib, use license from file rather than hardcoding it.

DubbleClick:
for what it's worth, this prevented imagick compilation with freetype or png support. It copied the symlink rather than resolving it, leading to an error in the linking phase.


DubbleClick:
Yeah the c++ support branch should really be merged because it gives us a consistent compiler toolchain, rather than using the one the OS provides.

crazywhalecc (approved):

PR #240, opened 2023-10-18
Prevent ghrel from downloading prerelease

### LINKED ISSUE #239: ghrel source pulls pre-releases
example would be icu, pulls 74.1 beta rather than 73.2 stable
or libpng using 16.41 instead of 16.40

crazywhalecc:
Fix #239 .

### DIFF
--- a/config/source.json
+++ b/config/source.json
@@ -128,8 +128,9 @@
         }
     },
     "icu": {
-        "type": "url",
-        "url": "https://github.com/unicode-org/icu/releases/download/release-73-1/icu4c-73_1-src.tgz",
+        "type": "ghrel",
+        "repo": "unicode-org/icu",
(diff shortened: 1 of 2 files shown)


DubbleClick:
Maybe make sure that tags also don't contain 'beta'.

DubbleClick:
> @DubbleClick libpng GitHub repo tag cannot get stable version, we still need hard-coded url to download it.
> 
> https://api.github.com/repos/glennrp/libpng/tags
> 
> It seems that the `https://git.code.sf.net/p/libpng/code` we are using now is the stable branch of libpng16.

If you check in the files it's downloading, it pulls the 16.41 version which is not released yet. 16.40 is the current stable version which is tagged on github/sourceforge.

PR #251, opened 2023-10-31
add imap support

### LINKED ISSUE #202: imap support
@DubbleClick 

> but even that requires a lot of patches both in the form of .patch files as well as replacing contents in files that no .patch file is available for. This project doesn't have any logic to apply patch files yet.

I understand. I tried the patch code in my previous implementation of libmemcached, where I originally planned to put the patch code in the `src/globals/patch/` directory, like this. But in the end, I directly used homebrew's build logic and independent mirror repo, and did not choose patch because it has not been updated for a long time: https://github.com/crazywhalecc/libmemcached-macos

image

> It also needs multiple services to be running on the system and a (self signed is fine) certificate. It's a complete mess that requires decisions to be taken to integrate it here that I'm not in a position to make.

From the results of my current local testing, it requires at least kerberos, and I'm currently stuck on compiling kerberos. 

Under the existing architecture, it may not be a good idea to use pre-installed dynamic libraries or packages on the system, because this will limit the environment in which spc can run.

henderkes:
this was considerably easier with the musl toolchains since it allows debian/ubuntu to build slx, which doesn't require libpam

closes #202 

tests:
- [x] x86 alpine
- [x] x86 rhel
- [x] x86 debian
- [x] aarch64 alpine
- [x] aarch64 rhel
- [x] aarch64 debian

todo: (have no experience with bsd, don't own a mac or aarch64 hardware)
- [ ] bsd x86
- [ ] bsd aarch64
- [x] mac aarch64

### DIFF
--- a/src/SPC/builder/linux/library/imap.php
+++ b/src/SPC/builder/linux/library/imap.php
@@ -7,7 +7,6 @@ namespace SPC\builder\linux\library;
 use SPC\exception\FileSystemException;
 use SPC\exception\RuntimeException;
 use SPC\store\FileSystem;
-use SPC\store\SourcePatcher;
 
 class imap extends LinuxLibraryBase
 {


DubbleClick:
If you can create a debian aarch64 vhdx that I can upload to azure, I'll be able to test aarch64 debian in the future, by the way. Otherwise, it will have to wait until the 12c snapdragon releases - will finally get a laptop then.

crazywhalecc:
I'm writing a manually triggered testing actions [here](https://github.com/static-php/static-php-cli-hosted/blob/master/.github/workflows/test-extensions.yml). It can specify different branches and repositories to test the effectiveness of the extension in the matrix of PHP8.0-8.2, Linux, macOS (using the official runner and local arm matrix).

I am now separating the download file process to avoid repeated test failures due to accidental download errors, but I don’t know why only arm-macos cannot hit the cache and cannot get the output of the previous workflow. I'm working on it.

crazywhalecc:
Debian aarch64 is good. But macOS support is still bad:

<details>
./tcp_unix.c:535:19: error: variable has incomplete type 'struct pollfd'
    struct pollfd pfd;
                  ^
./tcp_unix.c:535:12: note: forward declaration of 'struct pollfd'
    struct pollfd pfd;
           ^
./tcp_unix.c:545:20: error: use of undeclared identifier 'POLLIN'
      pfd.events = POLLIN;
                   ^
./tcp_unix.c:550:6: error: call to undeclared function 'poll'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
        i = poll (&pfd, 1, ti ? tmo * 1000 : -1);
            ^
./tcp_unix.c:591:17: error: variable has incomplete type 'struct pollfd'
  struct pollfd pfd;
                ^
./tcp_unix.c:591:10: note: forward declaration of 'struct pollfd'
  struct pollfd pfd;
         ^
./tcp_unix.c:602:18: error: use of undeclared identifier 'POLLIN'
    pfd.events = POLLIN;
                 ^
./tcp_unix.c:606:11: error: call to undeclared function 'poll'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
      i = poll (&pfd, 1, ti ? tmo * 1000 : -1);
          ^
./tcp_unix.c:659:17: error: variable has incomplete type 'struct pollfd'
  struct pollfd pfd;
                ^
./tcp_unix.c:659:10: note: forward declaration of 'struct pollfd'
  struct pollfd pfd;
         ^
./tcp_unix.c:670:18: error: use of undeclared identifier 'POLLOUT'
    pfd.events = POLLOUT;
                 ^
./tcp_unix.c:674:11: error: call to undeclared function 'poll'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
      i = poll (&pfd, 1, ti ? tmo * 1000 : -1);
          ^

</details>

crazywhalecc:
Updates:

I found that macOS is basically caused by the lack of system headers. Among them, `tcp_unix.c` file is missing `#include <poll.h>`, and at least 10 other files are missing `#include <time.h>` and `#include <utime.h>`. After adding them, use `osx` instead of `slx` to compile successfully.

crazywhalecc:
But this causes a problem. I don't want to repeat so much repetitive content in the patch, and making the patch file seems to conflict with the existing patch.

I wonder if we can directly fork `uw-imap/imap` and then apply the relevant patches to the code (except environment-related ones, such as SSL and CC). For macOS, a special .patchw file will be run before compilation to fix this error.

crazywhalecc:
Currently, it has been tested and runs well on Linux and macOS. Is it time to merge?

DubbleClick:
Checking code for imap extension? The default check with php --ri "imap" should work, right?

crazywhalecc:
> Checking code for imap extension? The default check with php --ri "imap" should work, right?

In fact, it is a runtime test script, such as [gd](https://github.com/crazywhalecc/static-php-cli/blob/main/src/globals/tests/gd.php).

If we don’t add it, and just use `--ri` to make sure it can run, then it doesn’t matter if we don’t add this test. It’s just an option.

DubbleClick:
Yeah we don't need to write a test here I think. Just the default check is fine, since it won't show up if imap isn't supported.

PR #260, opened 2023-11-08
Add extension rar support

crazywhalecc:


crazywhalecc:
Bug: rar with common extensions compilation is failed.

PR #267, opened 2023-11-15
fix pthreads4w URL

dunglas:
The current URL is down.

### DIFF
--- a/config/lib.json
+++ b/config/lib.json
@@ -311,8 +311,7 @@
         ],
         "lib-suggests-windows": [
             "icu",
-            "xz",
-            "pthreads4w"
+            "xz"
         ]
     },
     "libxslt": {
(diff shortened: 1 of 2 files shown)


crazywhalecc (approved):

crazywhalecc:
Oh, it seems pthreads4w is redundant, we could just remove it now.

dunglas:
From `source.json` or also from `lib.json`?

crazywhalecc:
Yeah. I just removed it :)

crazywhalecc:
And now we have `download --for-extensions=xxx,yyy` options. If your workflow doesn't have caching, or if you don't need many extensions, you can reduce download time by only downloading the content related to the extensions. (By default, all sources of the specified extensions are included.)

Issue #269, opened 2023-11-15
Freetype doesn't compile


dunglas:
https://github.com/dunglas/frankenphp/actions/runs/6881409688/job/18717691144?pr=304

Adding `--debug` doesn't give more informations

crazywhalecc:
https://github.com/crazywhalecc/static-php-cli/commit/e3671387fc3b2f09c11ba800e7b11236a3e46a53

It seems to be related to the fact that I changed the invalid freetype address yesterday. The tag source code of the GitHub mirror repository seems to be different from the source code of gnu.org.

dunglas:
Better logs:

```
#18 26.54 [10:06:05] [INFO] Building required library [freetype]
#18 26.54 [10:06:05] [INFO] Entering dir: /static-php-cli/source/freetype
#18 26.54 [10:06:05] [INFO] [EXEC] ./configure --enable-static --disable-shared --without-harfbuzz --prefix= --with-png --with-bzip2=/static-php-cli/buildroot --without-brotli 
#18 26.54 [10:06:05] [DEBU] Running command with direct output: cd '/static-php-cli/source/freetype' && ./configure --enable-static --disable-shared --without-harfbuzz --prefix= --with-png --with-bzip2=/static-php-cli/buildroot --without-brotli 
#18 26.56 
#18 26.56 FreeType build system -- automatic system detection
#18 26.56 
#18 26.56 The following settings are used:
#18 26.56 
#18 26.56   platform                    unix
#18 26.56   compiler                    gcc
#18 26.56   configuration directory     ./builds/unix
#18 26.56   configuration rules         ./builds/unix/unix.mk
#18 26.56 
#18 26.56 If this does not correspond to your system or settings please remove the file
#18 26.56 `config.mk' from this directory then read the INSTALL file for help.
#18 26.56 
#18 26.56 Otherwise, simply type `make' again to build the library,
#18 26.56 or `make refdoc' to build the API reference (this needs Python >= 3.5).
#18 26.56 
#18 26.56 Checking out submodule in `subprojects/dlg'
#18 26.56 git --git-dir=. submodule init
#18 26.57 fatal: not a git repository: '.'
#18 26.57 make: *** [builds/toplevel.mk:173: check_out_submodule] Error 128
[...]

PR #270, opened 2023-11-16
fix: FreeType build

dunglas:
Closes https://github.com/crazywhalecc/static-php-cli/issues/269.
See for the rationale: https://gitlab.freedesktop.org/freetype/freetype/-/issues/1158

Also let the SourceForge download page choose an appropriate mirror instead of hardcoding it.

## Tasks
- [X] Linux x86_64
- [X] Linux aarch64
- [x] macOS x86_64
- [X] macOS aarch64

### DIFF
--- a/config/source.json
+++ b/config/source.json
@@ -111,8 +111,9 @@
         }
     },
     "freetype": {
-        "type": "url",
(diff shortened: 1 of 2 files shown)


crazywhalecc:
Seems that we cannot dynamically obtain the latest version from the GitHub mirror repository.

PR #301, opened 2023-12-26
Initial Windows Support

crazywhalecc:
## What does this PR do?

- Fix bin/setup-runtime and bin/spc generated composer wrapper
- Bump version to 2.1.0-beta.1
- Add WindowsBuilder
- Add FileSystem tests

## Checklist before merging

- [x] If it's a extension or dependency update, make sure adding related extensions in `src/global/test-extensions.php`.
- [x] If you changed the behavior of static-php-cli, add docs in [static-php/static-php-cli-docs](https://github.com/static-php/static-php-cli-docs) . Related docs PR: https://github.com/static-php/static-php-cli-docs/pull/8 

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -114,6 +114,7 @@ jobs:
         os:
           - ubuntu-latest
           - macos-latest
+          - windows-latest
       fail-fast: false
     steps:
       - name: "Checkout"
--- a/README-zh.md
+++ b/README-zh.md
@@ -254,14 +254,13 @@ bin/spc micro:combine my-app.phar -I "memory_limit=4G" -I "disable_functions=sys
 
 ## 开源协议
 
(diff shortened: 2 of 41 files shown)


mario-deluna:
Omg, yes! You are amazing. Do you think this will work for GLFW as well?

Regardless, do you have PayPal, Patreon, or something? I would really like to send you a small sign of gratitude.

crazywhalecc:
@mario-deluna Thank you for your continued attention!

Support on Windows is currently in a very, very early stage and may only support the zlib extension at this time. In addition, I have never done Windows C++ and PowerShell development before, so the progress may be far less than that of Linux and macOS.

There should be some possibility of glfw extension, but there is a high probability that it cannot be purely statically compiled and can only be used by placing `dll` in the same directory (Windows programs will search for dll libraries in the same directory by default). 

I appreciate your sponsorship, but since I have a full-time job, I only have limited time to update it. And due to financial restrictions for personal in China, I can currently only use the Buy Me a Coffee platform to accept international sponsorships.

Issue #377, opened 2024-03-11
Add an option to build all libraries with debug symbols


dunglas:
This would be super useful to have the ability to build everything with debug symbols. This would make debugging issues such as https://github.com/php/php-src/issues/13648 easier.

To implement this, we could add a new `cflags` option, that will automatically set the `CFLAGS` env var for all built libs.
Then, something like `spc build --cflags='-g  -O0' --no-strip` should allow to get good stack traces of everything.

Would you accept such a patch?

dunglas:
Another option, probably better, would be to enable the debug mode in every library (if supported), if the `--no-strip` option is passed. This can be done as and when we need it.

withinboredom:
haha, I literally came here to ask the exact same question! Personally, I'd prefer a way to generically set CFLAGS and CXXFLAGS.

I haven't confirmed this is the cause, but it appears that the built libraries lack any optimizations `-O2`, for example, and thus statically compiled php appears to be orders of magnitudes slower than expected.

crazywhalecc:
I prepare to add `SPC_XXX` series to static-php-cli (EnvManager). First, we will cover common global variables, such as:

- `SPC_LIBS_EXTRA_CFLAGS`: for every library default extra cflags, default is empty. (will add for every lib build command before)
- `SPC_PHP_CFLAGS`: for php library default extra cflags, default is current state (maybe we need to initialize them before build).
- `SPC_LIB_LIBEVENT_CFLAGS`: for specific lib cflags. 

But this will have some problems:

1. We need to repeatedly define a bunch of very long environment variable names that may never be used.
2. Some libraries can be directly passed to CMake or Makefile through `export` before compilation, without the need to define SPC again.
3. The variable passing method does not solve part of the customization of the compilation command.

I still don't know what is best for spc. There may not be a perfect solution: because we cannot allow all parameters to be customized, it will greatly increase the complexity of the builder.

crazywhalecc:
After a month of trying and compromising, I finally decided to customize only some of the compile flags as needed. See #403 

This might not be a great solution, but I also don't want to get stuck in infinite customization hell.

crazywhalecc:
Finished, docs: https://static-php.dev/en/guide/env-vars.html

PR #378, opened 2024-03-11
feat: compile OpenSSL with debug symbols if --no-strip is passed

dunglas:
## What does this PR do?

It enables the debug symbols of OpenSSL if `--no-strip` is passed.

First step for https://github.com/crazywhalecc/static-php-cli/issues/377.

### DIFF
--- a/src/SPC/builder/linux/library/openssl.php
+++ b/src/SPC/builder/linux/library/openssl.php
@@ -58,6 +58,12 @@ class openssl extends LinuxLibraryBase
             $zlib_extra = '';
         }
 
+        if ($this->builder->getOption('no-strip')) {
+            $debug_extra = '-d ';
+        } else {
+            $debug_extra = '';
+        }
+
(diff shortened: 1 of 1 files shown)


src/SPC/builder/linux/library/openssl.php
```diff
@@ -58,6 +58,12 @@ public function build(): void
             $zlib_extra = '';
         }
 
+        if ($this->builder->getOption('no-strip')) {
+            $debug_extra = '-d ';
```
bukka, line 62: This actually completely disables the optimization (sets `-O0`) so it might be better to just use `-g ` which passes the flag.

dunglas: So https://github.com/crazywhalecc/static-php-cli/pull/379 will definitely be better.

dunglas:
Closing in favor of https://github.com/crazywhalecc/static-php-cli/pull/379.

PR #414, opened 2024-04-19
fix: use C++ 17 compiler for ICU

dunglas:
## What does this PR do?

ICU [now requires C++17](https://icu.unicode.org/download/75).
This patch fixes the build.

## Checklist before merging

> If your PR involves the changes mentioned below and completed the action, please tick the corresponding option.
> If a modification is not involved, please skip it directly.

- [x] If it's a extension or dependency update, make sure adding related extensions in `src/global/test-extensions.php`.
- [x] If you changed the behavior of static-php-cli, add docs in [static-php/static-php-cli-docs](https://github.com/static-php/static-php-cli-docs) .
- [x] If you updated `config/xxxx.json` content, run `bin/spc dev:sort-config xxx`.

PR #415, opened 2024-04-19
fix: downgrade to ICU 74

dunglas:
## What does this PR do?

It doesn't look currently possible to use ICU 75 with PHP because this version requires C++ 17.
https://github.com/php/php-src/pull/14002 and #414 will fix the issue. In the meantime, we have to downgrade ICU to v74.

## Checklist before merging

- [x] If it's a extension or dependency update, make sure adding related extensions in `src/global/test-extensions.php`.
- [x] If you changed the behavior of static-php-cli, add docs in [static-php/static-php-cli-docs](https://github.com/static-php/static-php-cli-docs) .
- [x] If you updated `config/xxxx.json` content, run `bin/spc dev:sort-config xxx`.

### DIFF
--- a/config/source.json
+++ b/config/source.json
@@ -187,9 +187,8 @@
         }
     },
     "icu": {
-        "type": "ghrel",
-        "repo": "unicode-org/icu",
-        "match": "icu4c.+-src\\.tgz",
+        "type": "url",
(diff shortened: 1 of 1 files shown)


dunglas:
Closing in favor of #414.

Issue #419, opened 2024-04-24
Windows Defender SmartScreen


simonhamp:
The first hurdle when trying to run a compiled `php.exe` seems to be this

![Screenshot 2024-04-24 at 13 04 11](https://github.com/crazywhalecc/static-php-cli/assets/31628/f8b6f313-2857-4563-8954-75dec05b38c0)

Clicking the 'More info' link reveals this:

![Screenshot 2024-04-24 at 13 04 33](https://github.com/crazywhalecc/static-php-cli/assets/31628/0c7c1e45-95ed-411b-bc87-792a5d7ecf30)

Obviously, we can't expect every Windows user to disable SmartScreen. I'm guessing there's some sort of code signing that's needed to overcome this?

crazywhalecc:
Unfortunately I know nothing about windows executable signing. A self-signed certificate may be able to sign, but may not avoid SmartScreen. In addition, using a third-party issuance of compiled `php.exe` may also be a violation of the agreement?

simonhamp:
Which agreement?

crazywhalecc:
@simonhamp I just assumed that I used a self-signed certificate and uploaded `php.exe` on static-php.dev. I don't know if doing so would violate the PHP License or other agreements or licenses.

The problem depends on how to avoid SmartScreen. If SmartScreen can be avoided by signing, and doing so doesn't violate any protocols, that's OK.

simonhamp:
Yeh I understand. I will do more digging

crazywhalecc:
I will temporarily close this issue until the issue becomes closely related to static-php-cli again.

Issue #418, opened 2024-04-24
Compiled Windows executable isn't running


simonhamp:
I've been able to compile PHP8.2 on a Windows GitHub Action 🎉

However, when I then try to use it on a Windows machine, I get the following error:
![image](https://github.com/crazywhalecc/static-php-cli/assets/31628/e4d52d2f-568d-4d75-befb-3b39b28d097b)

Seems that some part (I'm guessing curl?) is depending on `nghttp2` but this isn't statically link.

My extension string was:
```
bcmath,ctype,zlib,openssl,curl,iconv,xml,dom,fileinfo,filter,mbstring,pdo,sqlite3,pdo_sqlite,session,simplexml,sockets,tokenizer
```

crazywhalecc:
Fixed in main branch.

PR #420, opened 2024-04-24
Windows build GitHub Action

simonhamp:
## What does this PR do?
Adds a GitHub Action for building on Windows.

This is a first pass that builds an executable, but it's still WIP - please leave this PR open as I will be pushing to the original branch as we work through fixing it.

Current known issues:
- [x] ~It currently seems to only build PHP8.2 - I haven't started to dig into why~ This was because my cache key logic was broken. It works now
- [ ] #419 
- [x] When you get past SmartScreen, you hit #418 

## Checklist before merging

### DIFF
--- a/.github/workflows/build-windows-x86_64.yml
+++ b/.github/workflows/build-windows-x86_64.yml
@@ -56,7 +56,7 @@ jobs:
         shell: bash
         run: |
           INPUT_HASH=$(echo "${{ runner.os }}-${{ inputs.version }}-${{ inputs.extensions }}" | sha256sum | awk '{print $1}')
-          echo "INPUT_HASH=${INPUT_HASH}" >> $env:GITHUB_ENV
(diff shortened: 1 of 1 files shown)


simonhamp:
The test failures here seem to be due to hitting GitHub's rate limits

crazywhalecc:
It seems that my GITHUB_TOKEN has expired, I will update it.

.github/workflows/build-windows-x86_64.yml
```diff
@@ -0,0 +1,120 @@
...
+          - '7.4'
+      build-cli:
+        description: build cli binary
+        default: true
+        type: boolean
+      build-micro:
+        description: build phpmicro binary
+        type: boolean
+      build-fpm:
+        description: build fpm binary
```
crazywhalecc, line 25: Windows doesn't support fpm sapi.

crazywhalecc (changes requested):

crazywhalecc (approved):

PR #427, opened 2024-04-26
Fix Intel Mac build action

simonhamp:
## What does this PR do?

Fixes the Intel Mac build GitHub Action as it was missing the PHP runtime, Composer and other parts necessary to run the build process from SPC source.

## Checklist before merging

### DIFF
--- a/.github/workflows/build-macos-x86_64.yml
+++ b/.github/workflows/build-macos-x86_64.yml
@@ -46,6 +46,14 @@ jobs:
           brew install automake gzip
           echo "SPC_BUILD_OS=macos" >> $GITHUB_ENV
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build-macos-x86_64.yml
```diff
@@ -37,7 +37,7 @@ env:
 jobs:
   build:
     name: build ${{ inputs.version }} on macOS x86_64
-    runs-on: macos-latest
+    runs-on: macos-12
```
crazywhalecc, line 40: `macos-13` can be used now.

crazywhalecc (approved):

PR #428, opened 2024-04-26
Update actions

simonhamp:
## What does this PR do?

- Updates action versions to their latest ones
- Fixes the dependency cache key naming logic, bringing the other environments into line with what's implemented in #420 - caches weren't being used at all when specifying an explicit set of extensions because the key contained commas

This reduces errors and improves build times.

## Checklist before merging

### DIFF
--- a/.github/workflows/build-macos-x86_64.yml
+++ b/.github/workflows/build-macos-x86_64.yml
@@ -57,12 +57,17 @@ jobs:
       - if: steps.cache-composer-deps.outputs.cache-hit != 'true'
         run: composer update --no-dev --classmap-authoritative
 
+      - name: Generate hashed key for download cache
+        run: |
+          INPUT_HASH=$(echo "${{ runner.os }}-${{ inputs.version }}-${{ inputs.extensions }}" | sha256sum | awk '{print $1}')
+          echo "INPUT_HASH=${INPUT_HASH}" >> $GITHUB_ENV
+
       # Cache downloaded source
       - id: cache-download
         uses: actions/cache@v4
         with:
           path: downloads
-          key: php-${{ inputs.version }}-dependencies-${{ inputs.extensions }}
+          key: php-${{ env.INPUT_HASH }}
 
       # With or without debug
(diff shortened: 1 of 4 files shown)


crazywhalecc:
Previously I modified [test workflow](https://github.com/crazywhalecc/static-php-cli/blob/d98d6aeffb347dceb8ed50a0ecec3ae4b530275a/.github/workflows/tests.yml#L148C1-L153C50) for dynamic caching: Just cache downlods dir again and again. When new extension added, it will automatically skip downloading other sources that has been cached.

simonhamp:
@crazywhalecc I'm not sure I follow you. The cache in these didn't appear to be working at all because the key name was incorrect. This had the result of always downloading the extension files.

This PR fixes that so each action will now correctly cache the set of extensions that were requested.

This should make builds _of the same PHP-version+OS+extension set_ a little faster, at least during the cache's lifetime.

simonhamp:
In case it's not clear, this dependency cache is the extension build files (the result of `bin/spc download`).

I think the variance you were seeing in `cache-hit != true` was exactly down to this problem of commas in the key name.

This is why I've hashed the details of the build into a key name that will _never_ have commas regardless of how the build is configured.

.github/workflows/build-macos-x86_64.yml
```diff
@@ -39,7 +39,7 @@ jobs:
     name: build ${{ inputs.version }} on macOS x86_64
     runs-on: macos-latest
```
crazywhalecc, line 40: macos-13 (macos-latest is arm64 now)

crazywhalecc:
What I mean is that today's best caching strategies don't perform well in Actions. It has nothing to do with this PR. After fixing this small problem, we merge it into the main branch and test it again.

.github/workflows/build-macos-x86_64.yml
simonhamp: This was already fixed in #427, which has been merged, so this should be ok.

crazywhalecc (approved):

PR #445, opened 2024-05-09
Add initial intl-icu windows support

### LINKED ISSUE #435: Extension request: `intl` on Windows
Slightly more urgent than #429 

I am going to dig into how to do it, but just wanted to raise the ticket.

crazywhalecc:
## What does this PR do?

Try to fix #435 

## Checklist before merging

### DIFF
--- a/config/lib.json
+++ b/config/lib.json
@@ -133,6 +133,12 @@
             "libicuio.a",
             "libicuuc.a",
             "libicudata.a"
+        ],
+        "static-libs-windows": [
+            "icuuc.lib",
+            "icudt.lib",
+            "icuin.lib",
+            "icuio.lib"
         ]
     },
     "imagemagick": {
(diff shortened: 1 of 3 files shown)


crazywhalecc:
@simonhamp 

I traveled across almost entire internet, and found a lot of things that proved we cannot statically build icu with Visual Studio. If we want to compile statically on Windows, we may have to use Cygwin or MSYS2, but they are also not suitable for static linking.

The first evidence is the mail list from [SourceForge](https://sourceforge.net/p/icu/mailman/icu-support/thread/CY4PR1301MB19444E4488134CCAA6D5496EF4FD0%40CY4PR1301MB1944.namprd13.prod.outlook.com/#msg36217130): In short, it can't be done.

The second one is from [icu official docs](https://unicode-org.github.io/icu/userguide/icu4c/build.html#how-to-build-and-install-on-windows-with-cygwin): the following configure options (including `--enable-static`) currently do not work on Windows with Microsoft’s compiler.

I also tried to patch icu code, such as adding `/MT`, replacing `.dll` with `.lib`, applying other static build patch, and all of my work just failed to achieve this.

Although I don't want to admit it, it is almost impossible for me to statically compile the icu library with VC++ now, unless the icu official or others who are more familiar with VC++ can provide more details.

simonhamp:
Thanks for your work! and thorough explanation. This is super helpful. I will explore your suggestion as an alternative on Windows to see where we can get to.

Do I need to anything special when _building_ `static-php.exe` in order to enable the extension to be loaded dynamically in this way?

crazywhalecc:
> Do I need to anything special when building static-php.exe in order to enable the extension to be loaded dynamically in this way?

I haven't tested more extensions, but what is certain is that the default compilation results in working directly with `php.ini` and `extension=C:\path\to\ext.dll`. But the target of dll extension compilation needs to be consistent with PHP (VS main version, thread safety, PHP version)

PR #475, opened 2024-06-07
replace ext-zend-opcache with ext-opcache for spc extension list

DubbleClick:
This is useful because if your composer.json lists "ext-zend-opcache" (as "ext-opcache" is invalid) and use `composer check-platform-reqs` to select the appropriate extensions, spc will complain that the zend-opcache extension does not exist.

This would also affect the zend-guard and zend-debugger extensions, but they're severely outdated and/or not included in the free available php anyway.

### DIFF
--- a/src/SPC/command/BaseCommand.php
+++ b/src/SPC/command/BaseCommand.php
@@ -147,6 +147,7 @@ abstract class BaseCommand extends Command
 
     protected function parseExtensionList(string $ext_list): array
     {
+        $ext_list = str_replace('zend-opcache', 'opcache', $ext_list);
         $a = array_map('trim', explode(',', $ext_list));
         return array_values(array_filter($a, function ($x) {
             $filter_internals = [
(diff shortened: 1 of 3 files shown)


DubbleClick:
Example where this is used: https://github.com/dunglas/frankenphp/blob/main/build-static.sh#L27

crazywhalecc:
Regarding the standardization of extension names, I think we can maintain a global constant to save them, and move these internal extensions into the constants. And `opcache` may also have `Zend Opcache` and `zendopcache`(old pecl version) name. WDYT?

DubbleClick:
I agree, I thought about creating an "alias" section in ext.json but wasn't sure if it would be worth the effort with opcache being the only "common" extension that is in non-standard scheme in composer.

crazywhalecc:
I would like this, make it a constant, and filter it in parseExtensionList.

crazywhalecc (approved):

PR #511, opened 2024-08-07
Add option to manually specify config file path

mobot95:
## What does this PR do?
Add option to manually specify config file path

## Checklist before merging

- [x] If you changed the behavior of static-php-cli, update docs in `./docs/`.


crazywhalecc:
Sorry for late reply, I noticed this argument is not included for MacOSBuilder, and Windows PHP build process seems don't use `--with-config-file-path`. It would be even better if these could be improved.

PR #593, opened 2025-01-29
add opentelemetry extension support

CyberLine:
## What does this PR do?

Add Support for opentelemetry instrumentation in PHP > 8

## Checklist before merging
- [X] If you updated `config/xxx.json` content, run `bin/spc dev:sort-config xxx`.

### DIFF
--- a/config/ext.json
+++ b/config/ext.json
@@ -432,6 +432,21 @@
         "type": "builtin",
         "arg-type-unix": "custom"
     },
+    "opentelemetry": {
+        "notes": true,
+        "support": {
+            "Windows": "wip"
+        },
+        "arg-type": "enable",
+        "source": "opentelemetry",
+        "unix-only": true,
+        "ext-suggests": [
+            "grpc",
+            "protobuf",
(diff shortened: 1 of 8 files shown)


crazywhalecc:
Windows static build will be officially available after https://github.com/open-telemetry/opentelemetry-php-instrumentation/pull/164 merged.

config/ext.json
```diff
@@ -446,18 +446,11 @@
...
     },
     "opentelemetry": {
-        "notes": true,
         "support": {
-            "Windows": "wip"
+            "BSD": "wip"
         },
-        "source": "opentelemetry",
-        "unix-only": true,
-        "ext-suggests": [
```
CyberLine: This should be kept as its part of the optional extension dependencies:

image

crazywhalecc: I didn't find any part in the source code that depends on the other extensions above. Maybe the document says that there are optional acceleration combinations, but this does not belong to "optional extensions at the build level"

PR #599, opened 2025-02-06
Add command to dump required PHP extensions based on vendor/composer/…

### LINKED ISSUE #338: Add dev command: extension dumper for SPC
## What is extension dumper

Print required extensions from your composer project.

1. Read custom `vendor/composer/installed.json` or `vendor/composer/installed.php` file.
2. Get all extensions from composer generated file (reading `ext-*` word).
3. Print extensions.

CyberLine:
…installed.json, composer.lock, composer.json (in this order)

## What does this PR do?

This fixes #338

### DIFF
--- a/src/SPC/command/DumpExtensionsCommand.php
+++ b/src/SPC/command/DumpExtensionsCommand.php
@@ -4,67 +4,157 @@ declare(strict_types=1);
 
 namespace SPC\command;
 
+use SPC\store\FileSystem;
 use Symfony\Component\Console\Attribute\AsCommand;
-use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputOption;
 
 #[AsCommand(name: 'dump-extensions', description: 'Determines the required php extensions')]
 class DumpExtensionsCommand extends BaseCommand
 {
-    private array $files = [
-        'vendor/composer/installed.json',
-        'composer.lock',
-        'composer.json',
-    ];
+    protected bool $no_motd = true;
+
+    public function configure(): void
+    {
+        // path to project files or specific composer file
+        $this->addArgument('path', InputArgument::OPTIONAL, 'Path to project root', '.');
+        $this->addOption('format', 'F', InputOption::VALUE_REQUIRED, 'Parsed output format', 'default');
+        // output zero extension replacement rather than exit as failure
+        $this->addOption('no-ext-output', 'N', InputOption::VALUE_REQUIRED, 'When no extensions found, output default combination (comma separated)');
+        // no dev
+        $this->addOption('no-dev', null, null, 'Do not include dev dependencies');
+        // no spc filter
+        $this->addOption('no-spc-filter', 'S', null, 'Do not use SPC filter to determine the required extensions');
+    }
 
     public function handle(): int
     {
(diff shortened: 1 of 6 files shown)


crazywhalecc:
1. To be accurate, it seems we can only determine which extensions are needed from both `installed.json` and `composer.json`. What do you think?
2. I think we can add some project parameters, as well as remove redundant output, optional extended validation, so that static-php-cli can use the output directly. (Don't worry, I have already written some of the functionality, but I still need to figure out the first problem)

CyberLine:
> 1. To be accurate, it seems we can only determine which extensions are needed from both `installed.json` and `composer.json`. If these two files do not exist in current or passed directory, it will fail directly. What do you think?

The most accurate result is given, if you have all 3 files exist. Maybe i can add an error if composer install was not run in before (so installed.json is not present) or at least composer.lock must be present.

> 2. I think we can add some project parameters, as well as remove redundant output, optional extension validation using spc config, so that static-php-cli can use the output directly. (Don't worry, I have already written some of the functionality, but I still need to figure out the first problem)

I've just created the logic to get all extensions from all possible files. Gimme some more informations on what your goal is and i will change it.

src/SPC/command/DumpExtensionsCommand.php
```diff
@@ -4,67 +4,157 @@
...
 {
-    private array $files = [
-        'vendor/composer/installed.json',
-        'composer.lock',
-        'composer.json',
-    ];
+    protected bool $no_motd = true;
 
-    public function handle(): int
+    public function configure(): void
```
crazywhalecc, line 17: I added some options, we can customize our dumps.

src/SPC/command/DumpExtensionsCommand.php
```diff
@@ -4,67 +4,157 @@
...
     {
-        $fs = new Filesystem();
```
crazywhalecc: Filesystem class is in spc's dev packages, we cannot use it when spc is built as standalone binary.

crazywhalecc (commented):
I mainly separated files to parse and add some options, and if you think there's anything that needs improvement, feel free to suggest it.

CyberLine:
Looks fine and works for me. Maybe we can add `--for-extensions=auto` which then internally calls this command?

crazywhalecc:
> Looks fine and works for me. Maybe we can add `--for-extensions=auto` which then internally calls this command?

It is possible to implement internal calls, but `--for-extensions=auto` may not be appropriate because there are cases where the project path is not the current working directory. We need a way to pass in the path.

I am not very good at parameter naming and design, especially in mixed parameter situations (sometimes I need to merge the list of extensions exported in the project and the list of extensions specified via the command line).

CyberLine:
We also need to add back composer.json to the list of parsed files. Why? Because if i have custom extension requirements in my project, they will only be found in composer.json, not reflected in composer.lock

crazywhalecc:
> they will only be found in composer.json, not reflected in composer.lock

@CyberLine It can be found in `platform` segment:

```json
{
    "platform": {
        "php": ">= 8.1",
        "ext-mbstring": "*",
        "ext-zlib": "*"
    }
}
```

crazywhalecc:
I decided to defer the issue of `--for-extensions=auto`, which may be a difficult path to choose and requires careful consideration.

PR #611, opened 2025-03-05
add libacl for php-fpm compilation

DubbleClick:
## What does this PR do?
adds libacl for php-fpm compilation. php-fpm commonly uses listen.owner and listen.group, without acl, php-fpm will crash when encountering this config directive

## Checklist before merging
- [ ] add tests for library?
- [ ] add MacOS compatibility. It should build without issues, but it might need a similar patch as LinuxBuilder adding -L/buildroot/libs
- [ ] add BSD compatibility. It should build without issues, but it might need a similar patch as LinuxBuilder adding -L/buildroot/libs
- [x] doesn't exist on windows, so can skip that
- [ ] instead of checking for the presence of the library, should we instead add a separate option for compilation with fpm?

### DIFF
--- a/config/env.ini
+++ b/config/env.ini
@@ -80,7 +80,7 @@ SPC_CMD_PREFIX_PHP_CONFIGURE="${SPC_PHP_DEFAULT_LD_LIBRARY_PATH_CMD} ./configure
 ; make command
 SPC_CMD_PREFIX_PHP_MAKE="make -j${CPU_COUNT}"
 ; embed type for php, static (libphp.a) or shared (libphp.so)
-SPC_CMD_VAR_PHP_EMBED_TYPE="shared"
+SPC_CMD_VAR_PHP_EMBED_TYPE="static"
 
 ; *** default build vars for building php ***
 ; CFLAGS for configuring php


DubbleClick:
I'm not quite happy with the EXTRA_LDFLAGS_PROGRAM thing yet. The problem is that php's Makefile automatically adds -lacl at the end of a makefile when /path/to/buildroot/lib/libacl.a is already set in EXTRA_LIBS. I think it would be a better idea to patch that -lacl out of the command. Let me revert the changes.

How do we add a pathBeforeMake for a library that is not actually an extension?

DubbleClick:
I think that makes less sense than using --with-libs.

The real alternative I would see would be --build-fpm --with-fpm-acl. Bit then we'd need to add logic to conditionally build the library, rather than using the already existing getLib check.

src/SPC/builder/linux/LinuxBuilder.php
```diff
@@ -278,7 +278,7 @@ protected function buildMicro(): void
      */
     protected function buildFpm(): void
     {
-        $vars = SystemUtil::makeEnvVarString($this->getMakeExtraVars());
```
crazywhalecc, line 281: makeEnvVarString is different from `getEnvString`. `getEnvString` is more like a environment dumper, and `makeEnvVarString` is to generate k-v array to string.

crazywhalecc (changes requested):

src/SPC/builder/linux/LinuxBuilder.php
DubbleClick: Sorry, that was entirely unintentional

DubbleClick:
Actually, thinking about it, is there a reason why we don't just default this behaviour? fpm with acl can do everything that fpm without acl can't do. If the directives aren't used, it makes no difference.

Maybe we should just default this lib for fpm builds?

crazywhalecc:
> Maybe we should just default this lib for fpm builds?

That's a good question. As you know, currently spc only depends on pkg-config by default on *nix. If it is enabled by default, it means that in addition to pkg-config, these two libraries must be downloaded.

If most distributions override this library, adding it as a default I think is feasible, but we should also add it to the list of precompiled libraries to speed up the build.

With this change, I think we need to separate the way such libraries (pkg-config, libacl) are specified to support the continued addition of related libraries in the future. I'll think about where we should list them.

DubbleClick:
> If most distributions added and used this library, adding it as a default I think is feasible, but we should also add it to the list of precompiled libraries to speed up the build:

I doubt this will happen as SPC is still for static php, while all distros ship dynamic builds.

All php-fpm packages I know ship with acl, though. At least on Ubuntu, Debian and RHEL based distros. Not entirely sure about alpine, but alpine also has ACL as a utility.

crazywhalecc (commented):
I've added root node for lib.json via #618 , and we can add fpm deps in `php`.

crazywhalecc:
1. We assume that all users want to build with fpm, we certainly need to download libacl every time, same as phpmicro.
2. I prefer to add acl by default and add it to `lib-depends-unix` of `php`, but the `configure` still keep `if` state whether this library is added, so that we can quickly build the result without libacl by modifying the config. Adding `libacl` below looks good. https://github.com/crazywhalecc/static-php-cli/blob/36dc18012aa39a61fa22224dcb48cbda0d75e7ff/config/lib.json#L13

DubbleClick:
I like that idea. However, there's no php-fpm node, there's only the php node. If we add it there, it'll be required even for the CLI binary, which doesn't actually need it.

crazywhalecc:
> I like that idea. However, there's no php-fpm node, there's only the php node. If we add it there, it'll be required even for the CLI binary, which doesn't actually need it.

Since we currently have no way to pass the SAPI we want to build when using the `download` command, the simplest solution is to download them every time we build PHP on *nix.

Another solution is to use `--with-fpm-acl`, but this requires adding it to both download and build, but this will increase complexity.

Also, we can add some special deps patch, for example: building cli only will remove `libacl`. Or we add `php-fpm` and `php-cli` node first and building target SAPI will removes unused SAPI node from `php` node dynamically before compiling PHP to control which libraries will be built.

We can also split the `build` command into `build:cli`, `build:micro`, etc., and then select different dependency nodes when building different targets, as you said, but in this way we cannot build multiple SAPIs at the same time and need refactor.

DubbleClick:
Hmm, yeah I think you're right, it'd be fine to download and build libacl no matter what. It builds very fast so it shouldn't be much of an issue.

DubbleClick:
Now this is getting interesting. the ubuntu runner on php 8.3 works, the one on php 8.4 doesn't work because it adds -lacl despite /path/to/buildroot/lib/libacl.a already being specified.

I had this issue on RHEL at first, but then it stopped happening and now I cannot reproduce it anymore. Might have to path before make.

DubbleClick:
okay it works fine on alpine, rhel and ubuntu now.

I can't really work on macos with just runners as my reference, I think you'll have to take a look @crazywhalecc

DubbleClick:
it looks like it's trying to use l-prefixed versions of the functions which aren't defined. Maybe patching opt_deref to true would fix the issue?

crazywhalecc:
Oops, fpm-acl is more like a linux-only feature. I can't find a way to have fpm on a mac that has acl enabled.

DubbleClick:
That's strange, macos uses the posix permission model and supports acl. If you want I can disable it for mac os, but I don't understand why it wouldn't work.

crazywhalecc:
More old information: https://github.com/php/php-src/pull/8395

crazywhalecc:
I tried this manually, and got undefined `ACL_USER`, and I cannot find any header file containing this identifier.

```c
#include <sys/acl.h>
int main (void)
{
        acl_t acl;
        acl_entry_t user, group;
        acl = acl_init(1);
        acl_create_entry(&acl, &user);
        acl_set_tag_type(user, ACL_USER);
        acl_create_entry(&acl, &group);
        acl_set_tag_type(user, ACL_GROUP);
        acl_free(acl);
  ;
  return 0;
}
```

```
$ clang a.c -I/Library/Developer/CommandLineTools/SDKs/MacOSX15.2.sdk/usr/include
a.c:9:32: error: use of undeclared identifier 'ACL_USER'
    9 |         acl_set_tag_type(user, ACL_USER);
      |                                ^
a.c:11:32: error: use of undeclared identifier 'ACL_GROUP'
   11 |         acl_set_tag_type(user, ACL_GROUP);
      |                                ^
2 errors generated.
```

DubbleClick:
Okay, I'll remove it from MacOS then. We may need to split up lib-depends-unix to lib-depends-linux and lib-depends-macos x)

DubbleClick:
is there a reason why there isn't an alpine image in the docker workflows btw? just realized it's missing and I need to check on my vm manually

DubbleClick:
How does providing pre builts work, by the way? If I change the line to true, what else needs to be done to actually have it automatically download a prebuilt version?

crazywhalecc:
> is there a reason why there isn't an alpine image in the docker workflows btw? just realized it's missing and I need to check on my vm manually

Seems I forgot to change test ci... But it's just working, with musl-toolchain.

> How does providing pre builts work, by the way

I've created a PR in "hosted" repo: https://github.com/static-php/static-php-cli-hosted/pull/7

Once this pr merged, I'll merge it soon. This side we only need to set `provide-pre-built` to `true` and downloader will automatically search pre-built libraries.

DubbleClick:
huh, this closed automatically and I cannot reopen it

DubbleClick:
wtf

crazywhalecc:
What happened? Is there a problem with main branch merge?

PR #612, opened 2025-03-05
fix SPC_NO_MUSL_PATH not working in .env.ini

DubbleClick:
## What does this PR do?
prior to this patch, setting SPC_NO_MUSL_PATH has to be defined in the environment. setting it in .env.ini had no effect.

## Checklist before merging

- [] If it's an extension or dependency update, make sure adding related extensions in `src/global/test-extensions.php`.

### DIFF
--- a/src/SPC/builder/linux/LinuxBuilder.php
+++ b/src/SPC/builder/linux/LinuxBuilder.php
@@ -29,11 +29,13 @@ class LinuxBuilder extends UnixBuilderBase
         // check musl-cross make installed if we use musl-cross-make
         $arch = arch2gnu(php_uname('m'));
 
-        // set library path, some libraries need it. (We cannot use `putenv` here, because cmake will be confused)
-        $this->setOptionIfNotExist('library_path', "LIBRARY_PATH=/usr/local/musl/{$arch}-linux-musl/lib");
-        $this->setOptionIfNotExist('ld_library_path', "LD_LIBRARY_PATH=/usr/local/musl/{$arch}-linux-musl/lib");
-
         GlobalEnvManager::init($this);
+
+        // set library path, some libraries need it. (We cannot use `putenv` here, because cmake will be confused)
+        if (getenv('SPC_NO_MUSL_PATH') !== '1') {
+            $this->setOptionIfNotExist('library_path', 'LIBRARY_PATH=/usr/local/musl/lib');
+            $this->setOptionIfNotExist('ld_library_path', 'LD_LIBRARY_PATH=/usr/local/musl/lib');
+        }
 
         if (str_ends_with(getenv('CC'), 'linux-musl-gcc') && !file_exists("/usr/local/musl/bin/{$arch}-linux-musl-gcc") && (getenv('SPC_NO_MUSL_PATH') !== 'yes')) {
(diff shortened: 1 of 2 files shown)


src/SPC/builder/linux/LinuxBuilder.php
```diff
@@ -29,12 +29,14 @@ public function __construct(array $options = [])
...
         $arch = arch2gnu(php_uname('m'));
 
-        // set library path, some libraries need it. (We cannot use `putenv` here, because cmake will be confused)
-        $this->setOptionIfNotExist('library_path', "LIBRARY_PATH=/usr/local/musl/{$arch}-linux-musl/lib");
-        $this->setOptionIfNotExist('ld_library_path', "LD_LIBRARY_PATH=/usr/local/musl/{$arch}-linux-musl/lib");
-
         GlobalEnvManager::init($this);
 
+        // set library path, some libraries need it. (We cannot use `putenv` here, because cmake will be confused)
+        if (getenv('SPC_NO_MUSL_PATH') !== '1') {
```
crazywhalecc, line 35: Use `yes` instead?

crazywhalecc (changes requested):

src/SPC/builder/linux/LinuxBuilder.php
DubbleClick: the env reader interprets "yes" or "on" as "1".

DubbleClick: At least when the env file is parsed. I'm not sure about the case when it's set as an environment variable.

If it behaves differently there and passes "yes", we may have to compare in_array($var, ["1", "yes", "on", "true"]).

crazywhalecc: Currently getenv doesn't do the conversion, so a better solution might be `filter_var`.

DubbleClick: I'll push it later, sorry for not considering it earlier

crazywhalecc (approved):

crazywhalecc:
It seems that the dependency management of spc itself is more difficult. Limiting spc to only support the latest version of PHP (8.4) may reduce a lot of work. WDYT?

DubbleClick:
Php 8.4 to run spc? I don't see an issue with that.

Only compile php 8.4? I personally wouldn't mind because I immediately update anyway, but I think a lot of users would be upset. I think unless necessary, we should support building all supported versions (8.2+).


Which do you mean?

crazywhalecc:
@DubbleClick What I mean is that we will distribute the standalone spc anyway, so the PHP version that spc depends on can always be kept up to date, reducing the pain of composer dependencies. Of course, the PHP that spc can compile is still 8.0~8.4, which will not change.

DubbleClick:
>But an additional consideration is that many other workflows that depend on this project may use git to pull spc, and a violent incompatible update will conflict with the existing PHP version, which may be a potential problem.

Inconsequential, in my opinion. We can require php 8.4 to run spc and when someone doesn't wish to install php 8.4, they can use the standalone binary, frankenphp or another static php 8.4 binary to run.

DubbleClick:
I would tackle that in another PR, though.

DubbleClick:
yes, looks all good.

PR #615, opened 2025-03-07
Switch static-php-cli to support only PHP 8.4

crazywhalecc:
## What does this PR do?

Switch static-php-cli to support only PHP 8.4.

The reasons for this are:

- static-php-cli is not a library type, so there is no need to consider compatibility with a wide range of PHP runtimes.
- static-php-cli provides a standalone binary `spc` that does not require a PHP environment.
- There may be Composer dependency conflicts under multiple PHP versions, which will increase the maintenance cost of static-php.
- The PHP version required by static-php-cli does not conflict with the PHP version that static-php-cli supports compilation.

> cc @DubbleClick 

## Checklist before merging

### DIFF
--- a/README-zh.md
+++ b/README-zh.md
@@ -58,7 +58,7 @@ static-php-cli（简称 `spc`）有许多特性：
 
 ### 编译环境需求
 
-- PHP >= 8.1（这是 spc 自身需要的版本，不是支持的构建版本）
+- PHP 8.4（这是 spc 自身需要的版本，不是支持的构建版本）
 - 扩展：`mbstring,tokenizer,phar`
 - 系统安装了 `curl` 和 `git`
 
@@ -179,6 +179,9 @@ bin/spc --version
 ```bash
 # 检查环境依赖，并根据尝试自动安装缺失的编译工具
 ./bin/spc doctor --auto-fix
+
+# 输出目标项目依赖的扩展列表
+./bin/spc dump-extensions /path/to/your/project --format=text
 
 # 拉取所有依赖库
 ./bin/spc download --all
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
         }
     ],
     "require": {
-        "php": ">= 8.1",
+        "php": "~8.4.0",
         "ext-mbstring": "*",
         "ext-zlib": "*",
         "laravel/prompts": "^0.1.12",
@@ -19,7 +19,7 @@
     "require-dev": {
         "captainhook/captainhook-phar": "^5.23",
         "captainhook/hook-installer": "^1.0",
-        "friendsofphp/php-cs-fixer": "^3.25",
+        "friendsofphp/php-cs-fixer": "^3.60",
         "humbug/box": "^4.5.0 || ^4.6.0",
         "nunomaduro/collision": "^7.8",
         "phpstan/phpstan": "^1.10",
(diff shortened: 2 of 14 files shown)


composer.json
```diff
@@ -9,7 +9,7 @@
         }
     ],
     "require": {
-        "php": ">= 8.1",
+        "php": "~8.4.0",
```
DubbleClick, line 12: I would change this to "=>8.4". Otherwise either everyone immediately updates to 8.5 once it comes out, or nobody does. Package managers do not all release new versions on the same day.

README-zh.md
```diff
@@ -58,7 +58,7 @@ static-php-cli（简称 `spc`）有许多特性：
 
 ### 编译环境需求
 
-- PHP >= 8.1（这是 spc 自身需要的版本，不是支持的构建版本）
+- PHP 8.4（这是 spc 自身需要的版本，不是支持的构建版本）
```
DubbleClick, line 61: => 8.4 for the future.

DubbleClick (changes requested):

crazywhalecc:
At present, at least common distributions (Ubuntu, Alpine, Fedora) cannot directly install PHP8.4 through package management. Reducing it to 8.3~8.4 may be a compromise solution.

DubbleClick:
RHEL still installs 8.0 by default and doesn't have 8.2+ in the main rpm repo. I don't think we need to shackle ourselves to ancient versions just because users will have to enable extra modules in one command.

crazywhalecc:
> RHEL still installs 8.0 by default and doesn't have 8.2+ in the main rpm repo. I don't think we need to shackle ourselves to ancient versions just because users will have to enable extra modules in one command.

That's true. But I still think that the restrictions in composer.json should not be too strict. We have already fully used and required the latest version in the documentation, workflow and `setup-runtime`, and there is almost no essential difference between 8.3 and 8.4.

DubbleClick:
> and there is almost no essential difference between 8.3 and 8.4.

Technically true, but allowing php 8.3 will require us to stay at 8.3 level in the code and doesn't provide any real benefit. Who benefits from 8.3 support? Are there any developers working on this project who must use their systems php version and cannot update it to 8.4?

The only people this would possibly concern are people who use the main branch of this project in one of their deployment routines. Which is a bad idea, so that's not something we should consider.

crazywhalecc:
I know this is a rather inconsistent decision, but I don't want to cause a version upgrade to break the workflow of most downstream projects.

One of my private project, FrankenPHP, Herd, etc. seem to be built using 8.3. If it is forced to limit to 8.4, related projects will also need to update the compilation script immediately, or completely use `spc` binary for building.

For PHP 8.3, I think the following measures can be taken to gradually transition:

- Prompt the version that is about to be deprecated at startup
- Create a Pull Request for downstream updates at the same time
- Write the upcoming destructive updates in the update log (for spc source mode)

DubbleClick:
> FrankenPHP, Herd, and some project forks etc. seem to be built using 8.3. If it is forced to limit to 8.4, related projects will also need to update the compilation script immediately, or completely use spc binary for building.

Yes and no at the same time. They all require specific commits of this repository, so nothing would change immediately. They would continue to work just fine. Only when they update the SPC tag they use, they would also have to update their jobs from 8.3 to 8.4.

> Prompt the version that is about to be deprecated at startup

Yes, that's a good idea.

> Create a Pull Request for downstream updates at the same time

I don't think that's reasonable. We don't know all the downstream project and we don't know when they even plan to update the SPC release they use.

> Write the upcoming breaking changes in the update log (for spc source mode)

Good idea.

crazywhalecc:
Updated `spc-alpine-docker` and added deprecate notice. After the PR is merged, the gnu static branch needs to be updated synchronously.

PR #616, opened 2025-03-07
Remove deprecated args for PHP 8.4

### LINKED ISSUE #614: --with-openssl-dir unrecognised
Php 8.4 doesn't understand this option anymore. Not sure about 8.3. We should probably check for version to be built and remove it in newer versions.

crazywhalecc:
## What does this PR do?

Fix #614 

## Checklist before merging

### DIFF
--- a/src/globals/test-extensions.php
+++ b/src/globals/test-extensions.php
@@ -13,8 +13,6 @@ declare(strict_types=1);
 
 // test php version
 $test_php_version = [
-    '8.1',
-    '8.2',
     '8.3',
     '8.4',
 ];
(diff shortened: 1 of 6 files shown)


src/globals/test-extensions.php
```diff
@@ -40,7 +38,7 @@
 
 // If you want to test your added extensions and libs, add below (comma separated, example `bcmath,openssl`).
 $extensions = match (PHP_OS_FAMILY) {
-    'Linux', 'Darwin' => 'gettext',
+    'Linux', 'Darwin' => 'imap,openssl,zlib,memcache',
```
DubbleClick, line 41: I would take imap out here and turn zts back on. imap will at some point be removed from php entirely, zts is the future with more and more SAPIs embracing a worker model

DubbleClick (approved):

PR #621, opened 2025-03-10
update php default to 8.4

DubbleClick:
Had to revert the PHP_CS_FIXER.. there as it fails to run. The alias works on unix anyway.

```ps
PS C:\src\static-php-cli> git commit -a -m "update php defaul to 8.4"
pre-commit: 
 - composer cs-fix -- --config=.php-cs-fixer.php --dry-run --diff... : failed
captainhook failed executing all actions, took: 1.10s

composer cs-fix -- --config=.php-cs-fixer.php --dry-run --diff {$STAGED_FILES|of-type:php}
  command output:
> PHP_CS_FIXER_IGNORE_ENV=1 php-cs-fixer fix --config=.php-cs-fixer.php --dry-run --diff src/SPC/command/DownloadCommand.php
'PHP_CS_FIXER_IGNORE_ENV' is not recognized as an internal or external command,
operable program or batch file.
Script PHP_CS_FIXER_IGNORE_ENV=1 php-cs-fixer fix handling the cs-fix event returned with error code 1
failed to execute: composer cs-fix -- --config=.php-cs-fixer.php --dry-run --diff src/SPC/command/DownloadCommand.php
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,7 +200,7 @@ Basic usage for building php with some extensions:
 ./bin/spc download --for-extensions="openssl,pcntl,mbstring,pdo_sqlite"
 # download pre-built libraries first (save time for compiling dependencies)
 ./bin/spc download --for-extensions="openssl,curl,mbstring,mbregex" --prefer-pre-built
-# download different PHP version (--with-php=x.y or --with-php=x.y.z, recommend 8.1 ~ 8.3)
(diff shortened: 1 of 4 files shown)


DubbleClick:
No particular opinion on it, but I don't think it should be part of this PR. I only committed the removal of that line because it prevented committing.

PR #622, opened 2025-03-11
bring back pr for libacl

DubbleClick:
see old pr


config/env.ini
```diff
@@ -117,8 +117,6 @@ SPC_CMD_PREFIX_PHP_BUILDCONF="./buildconf --force"
 SPC_CMD_PREFIX_PHP_CONFIGURE="./configure --prefix= --with-valgrind=no --enable-shared=no --enable-static=yes --disable-all --disable-cgi --disable-phpdbg"
 ; make command
 SPC_CMD_PREFIX_PHP_MAKE="make -j${CPU_COUNT}"
-; embed type for php, static or shared
-SPC_CMD_VAR_PHP_EMBED_TYPE="static"
```
DubbleClick, line 121: this existed twice so i removed one

bin/spc-gnu-docker
```diff
@@ -136,5 +136,6 @@ echo 'SPC_CMD_VAR_PHP_MAKE_EXTRA_LIBS="-ldl -lpthread -lm -lresolv -lutil -lrt"'
 # shellcheck disable=SC2068
 # shellcheck disable=SC2086
 # shellcheck disable=SC2090
+$DOCKER_EXECUTABLE run --rm $INTERACT -e SPC_FIX_DEPLOY_ROOT="$(pwd)" $MOUNT_LIST cwcc-spc-gnu-$SPC_USE_ARCH bin/spc $@
 
 $DOCKER_EXECUTABLE run --rm $INTERACT -e SPC_FIX_DEPLOY_ROOT="$(pwd)" --env-file /tmp/spc-gnu-docker.env $MOUNT_LIST cwcc-spc-gnu-$SPC_USE_ARCH # bin/spc $@
```
crazywhalecc, line 141: Seems I forgot remove it.

config/env.custom.ini
```diff
@@ -0,0 +1,14 @@
...
+
+[linux]
+CC=gcc
+CXX=g++
+AR=ar
+LD=ld
+SPC_DEFAULT_C_FLAGS=-fPIC
+SPC_NO_MUSL_PATH=yes
+SPC_CMD_VAR_PHP_MAKE_EXTRA_LDFLAGS_PROGRAM="-Wl,-O1 -pie"
+SPC_CMD_VAR_PHP_MAKE_EXTRA_LIBS="-ldl -lpthread -lm -lresolv -lutil"
```
crazywhalecc, line 14: We don't need this file anymore.

config/env.ini
crazywhalecc: No, this is for macOS. It's two parts for different OS.

config/source.json
```diff
@@ -36,6 +36,16 @@
             "path": "LICENSE"
         }
     },
+    "attr": {
+        "type": "git",
+        "rev": "v2.5.2",
+        "url": "https://git.savannah.nongnu.org/git/attr.git",
+        "provide-pre-built": false,
```
crazywhalecc, line 43: `"provide-pre-built": true`

config/source.json
```diff
@@ -323,6 +333,16 @@
             "path": "LICENSE"
         }
     },
+    "libacl": {
+        "type": "git",
+        "rev": "v2.3.2",
+        "url": "https://git.savannah.nongnu.org/git/acl.git",
+        "provide-pre-built": false,
```
crazywhalecc, line 340: `"provide-pre-built": true`

src/SPC/builder/LibraryBase.php
```diff
@@ -294,6 +294,16 @@ public function beforePack(): void
         // do something before pack, default do nothing. overwrite this method to do something (e.g. modify pkg-config file)
     }
 
+    public function patchBeforeConfigure(): bool
+    {
+        return false;
+    }
+
+    public function patchBeforeMake(): bool
```
crazywhalecc, line 302: I think it would be better if we add a PHPDoc to provide instructions to avoid confusion with the Extension.

crazywhalecc (changes requested):

bin/spc-gnu-docker
DubbleClick: Sorry, I'm not following. What should I do here? Remove the green addition or comment it out?

crazywhalecc:
```
$DOCKER_EXECUTABLE run --rm $INTERACT -e SPC_FIX_DEPLOY_ROOT="$(pwd)" --env-file /tmp/spc-gnu-docker.env $MOUNT_LIST cwcc-spc-gnu-$SPC_USE_ARCH bin/spc $@
```

crazywhalecc (approved):

DubbleClick:
not sure what problem ubuntu now has. I've tried building on my vm's and it runs fine.

crazywhalecc:
I can reproduce this 8.3 bug locally on my debian vm, seems it's a linking dynamic libstdc++.so bug. But I have no idea why opcache happen to get this error.

DubbleClick:
Hmm I only have alpine (3.21), RHEL (9.5) and ubuntu (2404). Could you look into this for me?

crazywhalecc:
> Hmm I only have alpine (3.21), RHEL (9.5) and ubuntu (2404). Could you look into this for me?

I'll check tomorrow night

DubbleClick:
It's correct to add -lgomp. I think I know what the issue is, hold on a second.

crazywhalecc:
Another strange thing is that PHP8.4 does not bring any other extended dependency libraries for this check, only `./conftest`

DubbleClick:
we have green CI! at least if we ignore the failed 8.4 ubuntu curl download

src/SPC/doctor/item/LinuxToolCheckList.php
```diff
@@ -26,7 +26,7 @@ class LinuxToolCheckList
 
     public const TOOLS_DEBIAN = [
         'make', 'bison', 'flex',
-        'git', 'autoconf', 'automake',
+        'git', 'autoconf', 'automake', 'autopoint',
```
crazywhalecc, line 29: What dependencies do these two packages use?

config/env.ini
```diff
@@ -45,6 +44,14 @@ SPC_CONCURRENCY=${CPU_COUNT}
...
 ; Ignore some check item for bin/spc doctor command, comma separated (e.g. SPC_SKIP_DOCTOR_CHECK_ITEMS="if homebrew has installed")
 SPC_SKIP_DOCTOR_CHECK_ITEMS=""
+; EXTENSION_DIR where the built php will look for extension when a .ini instructs to load them
+; only useful for builds targeting glibc --libc=glibc
+; default paths
+; Ubuntu/Debian: /usr/lib/php/{PHP_VERSION}/
+; RHEL: /usr/lib64/php/modules
+; Alpine: /usr/lib/php{PHP_VERSION}/modules
+; where {PHP_VERSION} is 84 for php 8.4
+EXTENSION_DIR=
```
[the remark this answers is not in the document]
DubbleClick: Yes, it sets the default directory from where the resulting php binary/fpm/embed will look to load extensions if the php.ini doesn't define extension_dir. (it's commented out and empty by default)

By default this is 
something like /usr/lib/php/php84zts20240310 which isn't very useful because it constantly changes.

Versions from package managers set this to the directories I've listed as comments.

src/SPC/doctor/item/LinuxToolCheckList.php
DubbleClick: attr needs autopoint, otherwise ./autogen.sh fails.

PR #635, opened 2025-03-13
don't require libacl and attr - use --with-libs='libacl' if you build…

DubbleClick:
… fpm with acl support

unfortunate, but I honestly do not see a good way of solving this issue, without going for an extensive rework, that is imo not justified

### DIFF
--- a/config/lib.json
+++ b/config/lib.json
@@ -33,6 +33,7 @@
             "libattr.a"
         ],
         "lib-suggests": [
+            "gettext",
             "libiconv"
         ]
     },
@@ -255,6 +256,7 @@
             "attr"
         ],
         "lib-suggests": [
+            "gettext",
             "libiconv"
         ]
     },
--- a/config/lib.json
+++ b/config/lib.json
@@ -31,6 +31,9 @@
         "source": "attr",
         "static-libs-unix": [
             "libattr.a"
+        ],
+        "lib-suggests": [
+            "libiconv"
         ]
     },
     "brotli": {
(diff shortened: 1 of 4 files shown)


crazywhalecc:
BTW, I found `attr` will look for `libiconv`:

https://github.com/crazywhalecc/static-php-cli/actions/runs/13828423309/job/38687492168?pr=634

config/lib.json
```diff
@@ -11,11 +11,6 @@
         "lib-depends": [
             "lib-base",
             "micro"
-        ],
-        "lib-depends-linux": [
-            "lib-base",
-            "libacl",
```
crazywhalecc, line 17: Adding it to `lib-suggests-linux` is more reasonable.

DubbleClick:
> BTW, I found `attr` will look for `libiconv`:
> 
> https://github.com/crazywhalecc/static-php-cli/actions/runs/13828423309/job/38687492168?pr=634

yeah can disable it with --disable-nls
I conditionally disable that by looking for libiconv now

DubbleClick:
but I think we need to move attr and libacl after libiconv if the latter is active

crazywhalecc:
An important note, if a library contains `lib-suggests`, then its source cannot set `provide-pre-built` to true. So I prefer that the library does not provide native language support at all, not to mention that it is only dependent on libacl.

DubbleClick:
Hmm okay, I can remove it.

crazywhalecc (approved):

PR #642, opened 2025-03-14
Remove SPC_NO_MUSL_PATH, remove --libc, use SPC_LIBC instead

### LINKED ISSUE #633: --libc=glibc and SPC_NO_MUSL_PATH kind of serve the same purpose
Should get rid of one. I'd favour only keeping the --libc option. What do you think?

crazywhalecc:
## What does this PR do?

Fix #633 .

@DubbleClick Okay. I will make libc option as env var and remove SPC_NO_MUSL_PATH. Sounds it's indeed unnecessary, but I found that I use `SPC_NO_MUSL_PATH` in phpunit tests, let me check it out if it works.

## Checklist before merging

### DIFF
--- a/src/SPC/builder/linux/library/icu.php
+++ b/src/SPC/builder/linux/library/icu.php
@@ -16,7 +16,7 @@ class icu extends LinuxLibraryBase
     {
         $cppflags = 'CPPFLAGS="-DU_CHARSET_IS_UTF8=1  -DU_USING_ICU_NAMESPACE=1 -DU_STATIC_IMPLEMENTATION=1 -fPIC -fPIE -fno-ident"';
         $cxxflags = 'CXXFLAGS="-std=c++17"';
-        $ldflags = $this->builder->libc !== 'glibc' ? 'LDFLAGS="-static"' : '';
+        $ldflags = getenv('SPC_LIBC') !== 'glibc' ? 'LDFLAGS="-static"' : '';
         shell()->cd($this->source_dir . '/source')
             ->exec(
                 "{$cppflags} {$cxxflags} {$ldflags} " .
--- a/src/SPC/builder/unix/library/curl.php
+++ b/src/SPC/builder/unix/library/curl.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace SPC\builder\unix\library;
 
-use SPC\builder\linux\library\LinuxLibraryBase;
 use SPC\exception\FileSystemException;
 use SPC\exception\RuntimeException;
 use SPC\store\FileSystem;
@@ -53,11 +52,10 @@ trait curl
 
         FileSystem::resetDir($this->source_dir . '/build');
 
-        $cflags = $this instanceof LinuxLibraryBase && getenv('SPC_LIBC') === 'glibc' ? '-fPIC' : '';
         // compile！
         shell()->cd($this->source_dir . '/build')
             ->setEnv([
-                'CFLAGS' => trim($this->getLibExtraCFlags() . ' ' . $cflags),
+                'CFLAGS' => $this->getLibExtraCFlags(),
                 'LDFLAGS' => $this->getLibExtraLdFlags(),
                 'LIBS' => $this->getLibExtraLibs(),
             ])
--- a/src/SPC/builder/unix/library/curl.php
+++ b/src/SPC/builder/unix/library/curl.php
@@ -53,7 +53,7 @@ trait curl
 
         FileSystem::resetDir($this->source_dir . '/build');
 
-        $cflags = $this instanceof LinuxLibraryBase && $this->builder->libc === 'glibc' ? '-fPIC' : '';
+        $cflags = $this instanceof LinuxLibraryBase && getenv('SPC_LIBC') === 'glibc' ? '-fPIC' : '';
         // compile！
         shell()->cd($this->source_dir . '/build')
             ->setEnv([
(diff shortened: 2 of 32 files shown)


src/SPC/builder/unix/library/curl.php
```diff
@@ -53,7 +53,7 @@ protected function build(): void
 
         FileSystem::resetDir($this->source_dir . '/build');
 
-        $cflags = $this instanceof LinuxLibraryBase && $this->builder->libc === 'glibc' ? '-fPIC' : '';
+        $cflags = $this instanceof LinuxLibraryBase && getenv('SPC_LIBC') === 'glibc' ? '-fPIC' : '';
```
DubbleClick, line 56: This is unnecessary. -fPIC should be in the default C flags anyhow.

src/SPC/builder/unix/library/pkgconfig.php
```diff
@@ -11,7 +11,7 @@ trait pkgconfig
     protected function build(): void
     {
         $cflags = PHP_OS_FAMILY !== 'Linux' ? "{$this->builder->arch_c_flags} -Wimplicit-function-declaration -Wno-int-conversion" : '';
-        $ldflags = !($this instanceof LinuxLibraryBase) || $this->builder->libc === 'glibc' ? '' : '--static';
+        $ldflags = !($this instanceof LinuxLibraryBase) || getenv('SPC_LIBC') === 'glibc' ? '' : '--static';
```
DubbleClick, line 14: same question as before, can't it always be --static?

src/SPC/builder/unix/library/postgresql.php
```diff
@@ -47,7 +47,7 @@ protected function build(): void
         $error_exec_cnt += $output[0] === 0 ? 0 : 1;
         if (!empty($output[1][0])) {
             $ldflags = $output[1][0];
-            $envs .= !($this instanceof LinuxLibraryBase) || $this->builder->libc === 'glibc' ? " LDFLAGS=\"{$ldflags}\" " : " LDFLAGS=\"{$ldflags} -static\" ";
+            $envs .= !($this instanceof LinuxLibraryBase) || getenv('SPC_LIBC') === 'glibc' ? " LDFLAGS=\"{$ldflags}\" " : " LDFLAGS=\"{$ldflags} -static\" ";
```
DubbleClick, line 50: same question

src/SPC/doctor/item/MacOSToolCheckList.php
```diff
@@ -40,7 +40,7 @@ public function checkBrew(): ?CheckResult
         if (($path = $this->findCommand('brew')) === null) {
             return CheckResult::fail('Homebrew is not installed', 'brew');
         }
-        if ($path !== '/opt/homebrew/bin/brew' && php_uname('m') === 'arm64') {
+        if ($path !== '/opt/homebrew/bin/brew' && getenv('SPC_ARCH') === 'arm64') {
```
DubbleClick, line 43: didn't you rename it to aarch64 earlier?

src/globals/internal-env.php
```diff
@@ -45,5 +45,6 @@
 GlobalEnvManager::putenv('SOURCE_PATH=' . SOURCE_PATH);
 GlobalEnvManager::putenv('DOWNLOAD_PATH=' . DOWNLOAD_PATH);
 GlobalEnvManager::putenv('CPU_COUNT=' . CPU_COUNT);
+GlobalEnvManager::putenv('SPC_ARCH=' . php_uname('m'));
```
DubbleClick, line 48: Why do we have SPC_ARCH, GNU_ARCH and MAC_ARCH? Isn't it just x64 vs aarch64?

crazywhalecc: `SPC_ARCH` is directly from `php_uname('m')`, and on macOS, apple sillicon, is `arm64` while gnu arch on linux is `aarch64`. And Windows will get `AMD64` on x86 platform.

crazywhalecc: Because of these different situation, spc internal always use gnu arch as prefix or suffix, but sometimes need to sync to uname (like `--host=arm64-darwin`)

src/SPC/doctor/item/MacOSToolCheckList.php
crazywhalecc: Oh, actually I should use `GNU_ARCH` here. See my comments below.

DubbleClick: I think you should check MAC_ARCH against aarch64 here, given that this is the MacOSToolCheckList.

src/SPC/builder/linux/library/icu.php
```diff
@@ -16,7 +16,7 @@ protected function build(): void
     {
         $cppflags = 'CPPFLAGS="-DU_CHARSET_IS_UTF8=1  -DU_USING_ICU_NAMESPACE=1 -DU_STATIC_IMPLEMENTATION=1 -fPIC -fPIE -fno-ident"';
         $cxxflags = 'CXXFLAGS="-std=c++17"';
-        $ldflags = $this->builder->libc !== 'glibc' ? 'LDFLAGS="-static"' : '';
+        $ldflags = getenv('SPC_LIBC') !== 'glibc' ? 'LDFLAGS="-static"' : '';
```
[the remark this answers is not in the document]
crazywhalecc: When using `-static` with `glibc` to build icu, it will get configure error: `ld: cannot find -lc`.

src/SPC/builder/unix/library/pkgconfig.php
crazywhalecc: Same as icu

DubbleClick (approved):

PR #658, opened 2025-03-19
Build freetype using CMake instead

### LINKED ISSUE #656: bin/build-frankenphp freetype: unexpected token `win32-dll'
```console
  configuration directory     ./builds/unix
  configuration rules         ./builds/unix/unix.mk

If this does not correspond to your system or settings please remove the file
`config.mk' from this directory then read the INSTALL file for help.

Otherwise, simply type `make' again to build the library,
or `make refdoc' to build the API reference (this needs Python >= 3.5).

cd builds/unix; \
        ./configure
checking build system type... x86_64-pc-linux-gnu
checking host system type... x86_64-pc-linux-gnu
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables...
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether gcc accepts -g... yes
checking for gcc option to accept ISO C89... none needed
checking how to run the C preprocessor... gcc -E
checking for pkg-config... /usr/bin/pkg-config
checking pkg-config is at least version 0.24... yes
./configure: line 3463: syntax error near unexpected token `win32-dll'
./configure: line 3463: `LT_INIT(win32-dll)'
make: *** [builds/unix/detect.mk:91: setup] Error 2
```

I've tried upgrading libtool and other shenanigans, but this still happens. Manually removing the win32-dll target doesn't solve it either because it then can't find ./build/unix/libtool (which shouldn't be there anyhow).

crazywhalecc:
## What does this PR do?

Fix #656 

## Checklist before merging

### DIFF
--- a/src/SPC/builder/unix/library/freetype.php
+++ b/src/SPC/builder/unix/library/freetype.php
@@ -18,20 +18,21 @@ trait freetype
      */
     protected function build(): void
     {
-        $extra_libs = $this->builder->getLib('libpng') ? '--with-png' : '--without-png';
-        $extra_libs .= ' ';
-        $extra_libs .= $this->builder->getLib('bzip2') ? ('--with-bzip2=' . BUILD_ROOT_PATH) : '--without-bzip2';
-        $extra_libs .= ' ';
-        $extra_libs .= $this->builder->getLib('brotli') ? ('--with-brotli=' . BUILD_ROOT_PATH) : '--without-brotli';
-        $extra_libs .= ' ';
-
-        shell()->cd($this->source_dir)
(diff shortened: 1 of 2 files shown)


DubbleClick:
Have you tried it with the centos 7 image or should I give it a spin?

crazywhalecc:
> Have you tried it with the centos 7 image or should I give it a spin?

Oops. Seems have some trouble with gnu docker and alpine docker.

DubbleClick (approved):

PR #660, opened 2025-03-19
Fix windows curl build >= 8.3

crazywhalecc:
## Checklist before merging

### DIFF
--- a/config/lib.json
+++ b/config/lib.json
@@ -68,7 +68,7 @@
             "libcurl.a"
         ],
         "static-libs-windows": [
-            "libcurl.lib"
+            "libcurl_a.lib"
         ],
         "headers": [
             "curl"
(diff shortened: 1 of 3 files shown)


DubbleClick:
Why does it have to be renamed to libcurl_a.lib?

crazywhalecc:
@DubbleClick Added this newer than 8.3:

https://github.com/php/php-src/blob/fdeadcd9ba0627cf1967220489ab6caa031e4034/ext/curl/config.w32#L33-L35

DubbleClick:
Ah they're trying to differentiate between an object containing .lib and a link layer .lib to link against the shared dll. A bit of a weird way to do that, but we'll have to follow suit I guess.

crazywhalecc:
And then I think we should add weekly extension test for windows ASAP. This problem should perhaps have been discovered earlier.

PR #661, opened 2025-03-19
enable PDO_ODBC and ODBC extension statically

yoramdelangen:
## What does this PR do?
Enabling the PHP extensions `ODBC` and `PDO_ODBC`. 

## Checklist before merging

### DIFF
--- a/config/ext.json
+++ b/config/ext.json
@@ -428,7 +428,8 @@
     },
     "odbc": {
         "support": {
-            "BSD": "wip"
+            "BSD": "wip",
+            "Windows": "wip"
         },
         "type": "builtin",
         "arg-type-unix": "custom",
@@ -504,7 +505,8 @@
     },
     "pdo_odbc": {
         "support": {
-            "BSD": "wip"
+            "BSD": "wip",
+            "Windows": "wip"
         },
         "type": "builtin",
         "arg-type": "custom",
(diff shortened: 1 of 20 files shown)


DubbleClick:
If it's an extension or dependency update, make sure adding related extensions in src/global/test-extensions.php.

other than that, looks good

yoramdelangen:
Do all extension to be added into the `globals/test-extensions.php`?. Do you want me to update the extensions docs as well?

DubbleClick:
We add the extensions currently worked on to test-extensions.

Isn't the bulk option for librariest? If it also affects extensions yes it should be added.

crazywhalecc:
Thank you for your contribution. I'll take over to add some tests and docs if you don't mind.

crazywhalecc:
I reverted bulk changes because there is no point in modifying the fixed combination here. The real `bulk` combination is set in [static-php-cli-hosted](https://github.com/static-php/static-php-cli-hosted) repo and this tests is only for easy reference.

If you think it's necessary to add odbc to example bulk combination, you can make another CI update PR in "hosted" repo and add here again.

DubbleClick:
I think you want to revert that force push x)

DubbleClick:
@yoramdelangen

crazywhalecc:
If anything else has done, I'll add it to tests later.

yoramdelangen:
it works for Windows as well, so no need for WIP in the docs. Everything should be done

PR #663, opened 2025-03-20
Add mimalloc support for macOS and Linux

### LINKED ISSUE #662: Option to link against MIMALLOC statically
We should add libmimalloc and then in the end link mimalloc.o (yes .o) before all other libs.

The option should work with both musl and glibc.

Mimalloc is *much* faster than musl libc's allocator in threaded scenarios and even faster than glibc's allocator. While a glibc targeted build can easily load mimalloc with LD_PRELOAD, it would be more convenient to offer a static linking option for it.

cc @dunglas

DubbleClick:
## What does this PR do?
add mimalloc as an optional library.
if building with --with-libs="mimalloc", it replaces the libc allocator.

Fix #662 

## Checklist before merging

### DIFF
--- a/config/lib.json
+++ b/config/lib.json
@@ -589,6 +589,15 @@
             "openssl"
         ]
     },
+    "mimalloc": {
+        "source": "mimalloc",
+        "static-libs-unix": [
+            "mimalloc.o"
+        ],
+        "static-libs-windows": [
+            "mimalloc.lib"
+        ]
+    },
     "ncurses": {
         "source": "ncurses",
         "static-libs-unix": [
--- a/config/lib.json
+++ b/config/lib.json
@@ -593,9 +593,6 @@
         "source": "mimalloc",
         "static-libs-unix": [
             "mimalloc.o"
-        ],
-        "static-libs-windows": [
-            "mimalloc.lib"
         ]
     },
     "ncurses": {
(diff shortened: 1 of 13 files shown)


crazywhalecc:
Is it linux-only feature?

DubbleClick:
no, this works on macos too.

on windows it would work too, but there's no benefit to it. fully static windows builds can still preload the shared mimalloc library if someone wants, but msvc's allocator is fast anyway.

DubbleClick:
this will majorly simplify frankenphp's mimalloc build (can get rid of the whole patch code and only need to set buildroot/lib/mimalloc.o as the first "library" in CGO_LDFLAGS.

it's also nice for static-php-cli compiling cli/fpm because musl's allocator is awfully slow in threaded scenarios

crazywhalecc (approved):

crazywhalecc:
Ready to merge?

DubbleClick:
in theory yes, chinese docs are missing and some php 8.1 tests seem to fail

DubbleClick:
never mind, chinese dovs are not missing :)

DubbleClick:
> Ready to merge?

yes ready to merge now

crazywhalecc (approved):

PR #671, opened 2025-03-23
fix phpize using wrong paths

DubbleClick:
## Checklist before merging


crazywhalecc (approved):

DubbleClick:
stumbled across this when building xdebug for the frankenphp docs...

this would actually let "static"-php-"cli" compile dynamic extensions per the users choice as well.

Issue #672, opened 2025-03-23
Going full circle back to dynamic - xdebug support!


DubbleClick:
With the glibc support (or embed=dynamic with musl) support merged in, we can actually start offering dynamic-only extensions such as xdebug now.

I'll create a proof of concept for fun. We don't have to merge it. The only "rework" it would require is building those extensions *after* php-embed.

PR #673, opened 2025-03-23
add xdebug dynamic extension

DubbleClick:
## What does this PR do?
see  #672 

## Checklist before merging

- [x] If you changed the behavior of static-php-cli, update docs in `./docs/`.


DubbleClick:
This requires some work from your side. Technically these extensions don't have libraries but should have a build function themselves, rather than adding a library with the same name. I only went this route to get a proof of concept working quickly.

There's also no functionality to test these extensions yet. It could be done by adding a `extension=$name` to the php.ini, running `make install` in the $name source directory and then launching `buildroot/bin/php` (requires cli compilation) to check if it loads the extension.

crazywhalecc:
Interesting. If it is confirmed to work for glibc and macOS, maybe it would be better to directly support this feature in Extension.php?

For example, when I build PHP, if I pass `--with-shared-extensions=xdebug,XXX,YYY`, Builder will call the `buildShared()` method in `Extension.php` after building the cli and verifying it. But we may also need to make additional adjustments to the doc structure, or only support one or a few extensions such as xdebug, and make special notes. What do you think?

DubbleClick:
> Interesting. If it is confirmed to work for glibc and macOS, maybe it would be better to directly support this feature in Extension.php?

I cannot confirm macOS other than through github actions. I think we either use a different attribute (DynamicExt) because that way, we can distinguish which libraries/extensions need to be built before or after php. We could also support putting *both* attributes on an extension, that way it would be statically baked into php, but would also produce a .so extension to copy out.

>For example, when I build PHP, if I pass --with-shared-extensions=xdebug,XXX,YYY, Builder will call the buildShared() method in Extension.php after building the cli and verifying it. But we may also need to make additional adjustments to the doc structure, or only support one or a few extensions such as xdebug, and make special notes. What do you think?

I think it would be nice for a few extensions. Treating it as a general possibility is better than hacking something together as I did for the proof of concept. This is your project and your decision about the way it's implemented.

>Also, add "target": ["static", "shared"] for ext.json?

Yes, great idea. `bin spc build "intl,pgsql" --with-shared="pgsql,xdebug" --build-embed --build-cli` would bake pgsql into the php-cli but also produce a shared library/embed sapi, but also produce a pqsql.so that could be loaded dynamically.

Which... thinking about it... doesn't make all that much sense, I guess. But it doesn't hurt either, so why not.

crazywhalecc:
It seems that some changes broke the dependency manager and I can't compile extensions with `ext-depends` separately.

src/SPC/command/BuildPHPCommand.php
```diff
@@ -107,18 +126,22 @@ public function handle(): int
             $builder = BuilderProvider::makeBuilderByInput($this->input);
             $include_suggest_ext = $this->getOption('with-suggested-exts');
             $include_suggest_lib = $this->getOption('with-suggested-libs');
-            [$extensions, $libraries, $not_included] = DependencyUtil::getExtsAndLibs($extensions, $libraries, $include_suggest_ext, $include_suggest_lib);
+            [$extensions, $libraries, $not_included] = DependencyUtil::getExtsAndLibs(array_merge($static_extensions, $shared_extensions), $libraries, $include_suggest_ext, $include_suggest_lib);
```
crazywhalecc, line 129: I think here is not approperiate. We need to use `$extensions` as static extensions `$static_extensions`. (or separate? I don't know)

DubbleClick: Hmm, I'm not sure why this wouldn't be fine. My thought here is that we need to compile all required extensions dependencies, no matter if shared or static.

crazywhalecc: I've added a filter for `$static_extensions` to solve this problem.

crazywhalecc:
Everything looks good to me now, but I haven't fully tested all we added features.

src/SPC/store/Config.php
```diff
@@ -137,7 +137,7 @@ public static function getExtTarget(string $name): ?array
         if (!isset(self::$ext[$name])) {
             throw new WrongUsageException('ext [' . $name . '] is not supported yet');
         }
-        return self::$ext[$name]['target'] ?? ['static'];
+        return self::$ext[$name]['target'] ?? ['static', 'shared'];
```
DubbleClick, line 140: this makes every extension buildable as shared extension by default. We'll have to write a test suite for shared extensions then

crazywhalecc: What if we just add existing `ext-tests/xxx.php` for additional shared extension test? I randomly tested several extensions and they works quite well.

DubbleClick: Yeah, I wouldn't suspect any extensions to have an issue with this as php is generally much more supportive of glibc linux with shared linking, but we should still make sure.

We don't need to add tests under ext-tests I think. but the test-extensions.php should also test building shared extensions.

crazywhalecc (approved):

DubbleClick:
🥇

PR #681, opened 2025-03-29
Remove openmp support for imagemagick

### LINKED ISSUE #678: libgomp dependency in imagick
this is a problem in glibc builds because libgomp is apparently not preinstalled on ubuntu arm64 minimal. The authors of the imagick library recommend disabling it anyway, though: 

https://github.com/Imagick/imagick?tab=readme-ov-file#openmp

crazywhalecc:
## What does this PR do?

Fix #678 

## Checklist before merging

### DIFF
--- a/src/SPC/builder/extension/imagick.php
+++ b/src/SPC/builder/extension/imagick.php
@@ -5,23 +5,11 @@ declare(strict_types=1);
 namespace SPC\builder\extension;
 
 use SPC\builder\Extension;
-use SPC\builder\linux\LinuxBuilder;
 use SPC\util\CustomExt;
 
 #[CustomExt('imagick')]
 class imagick extends Extension
 {
-    public function patchBeforeMake(): bool
-    {
-        // imagick may call omp_pause_all which requires -lgomp
-        $extra_libs = getenv('SPC_EXTRA_LIBS') ?: '';
-        if ($this->builder instanceof LinuxBuilder) {
-            $extra_libs .= (empty($extra_libs) ? '' : ' ') . '-lgomp ';
-        }
-        f_putenv('SPC_EXTRA_LIBS=' . $extra_libs);
-        return true;
-    }
-
     public function getUnixConfigureArg(): string
     {
         return '--with-imagick=' . BUILD_ROOT_PATH;
--- a/src/SPC/builder/extension/imagick.php
+++ b/src/SPC/builder/extension/imagick.php
@@ -5,25 +5,25 @@ declare(strict_types=1);
 namespace SPC\builder\extension;
 
 use SPC\builder\Extension;
-use SPC\exception\FileSystemException;
-use SPC\store\FileSystem;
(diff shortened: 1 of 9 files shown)


crazywhalecc:
Looks good to me and main branch of frankenphp.

src/SPC/builder/extension/imagick.php
```diff
@@ -5,20 +5,16 @@
...
+    public function patchBeforeBuildconf(): bool
     {
-        // imagick may call omp_pause_all which requires -lgomp
-        $extra_libs = getenv('SPC_EXTRA_LIBS') ?: '';
-        if ($this->builder instanceof LinuxBuilder) {
-            $extra_libs .= (empty($extra_libs) ? '' : ' ') . '-lgomp ';
-        }
-        f_putenv('SPC_EXTRA_LIBS=' . $extra_libs);
+        // destroy imagick build conf to avoid libgomp build error
+        FileSystem::replaceFileStr(SOURCE_PATH . '/php-src/ext/imagick/config.m4', '#include <omp.h>', '#include <NEVER_INCLUDE_OMP_H>');
```
DubbleClick, line 17: what is `NEVER_INCLUDE_OMP_H`?

all the omp stuff is guarded by #ifdef guards, I would instead patch HAVE_OMP_PAUSE_RESOURCE_ALL to false, which can be done by passing the ac_... variable to false

src/SPC/builder/unix/library/imagemagick.php
```diff
@@ -19,7 +19,7 @@ trait imagemagick
     protected function build(): void
     {
         // TODO: imagemagick build with bzip2 failed with bugs, we need to fix it in the future
-        $extra = '--without-jxl --without-x --enable-openmp --without-bzlib ';
+        $extra = '--without-jxl --without-x --disable-openmp --without-bzlib ';
```
DubbleClick, line 22: bzip2 worked just fine in my testing with php 8.1 and 8.4

src/SPC/builder/extension/imagick.php
DubbleClick: also please leave a comment why we do this instead of linking against -lgomp, because that's really down to poor prebuilt libgomp support and can be reverted once we produce our own gcc toolset for building.

DubbleClick:
does this patch work for imagick.so as well? I think if we choose to match something rather than pass the ac_has_omp_pause_resource_all=op to php build, we should directly patch the ext-imagick/config.m4 from where it originates

crazywhalecc:
I tried but it still dont generate `HAVE_OMP_PAUSE_RESOURCE_ALL 0`. The safest way (maybe?) is to define it manually.

DubbleClick:
Okay I'll test and merge it later today, thank you

DubbleClick (approved):

DubbleClick:
@crazywhalecc this is good to be merged from my side

PR #685, opened 2025-03-30
Downloader enhancement

### LINKED ISSUE #682: Downloader enhancement
It's time to do some optimization on the Downloader component.

- Differentiate pre-built packages for different architectures (to prevent mistakenly using libraries for different operating systems), which is useful for my own development on Mac.
- Provide `--all-from-zip` instead of `--from-zip` command to automatically download the daily cache of all library source code from GH to reduce download errors.
- Redundant git clone output should not be displayed when `--debug` is not used.
- Add download address to info log.

crazywhalecc:
## What does this PR do?

Fix #682 

- Add match pattern for different OS
- Add pre-built lock for specific OS and libc
- Quiet console output for non --debug mode

## Checklist before merging


DubbleClick:
I think this may not be as easy as os-arch-libc. Libraries built with glibc 2.17 may not be usable in compilation with glibc >=2.18.

crazywhalecc:
> I think this may not be as easy as os-arch-libc. Libraries built with glibc 2.17 may not be usable in compilation with glibc >=2.18.

We assume all developers build glibc using spc-gnu-docker or CentOS 7 just like described in the documentation. 

But what you said makes sense. Maybe we should add an obvious warning or add a libc version suffix?

DubbleClick:
Yes, I think os-arch-libc-libcversion should suffice. Since it wouldn't match on non 2.17, it would simply reuse the current functionality of falling back to source download. Good idea.

crazywhalecc:
Another question is: `spc-alpine-docker` is using `alpine:edge` as base image. When we use musl, the alpine-based musl version may be upgraded at any time, which will cause confusion.

DubbleClick:
I think musl *tries* to stay forward compatible, but doesn't guarantee it. Since musl is moving very slowly, I think we could add the version string there too.

src/SPC/builder/linux/SystemUtil.php
```diff
@@ -202,7 +202,6 @@ public static function getLibcVersionIfExists(): ?string
             }
             return null;
         }
-        /*
         // for musl, disabled temporarily as musl has better compatibility between different patch version?
```
DubbleClick, line 205: remove the comment too

crazywhalecc:
My IP has been banned by GitHub from accessing the API... But everything seems to be working fine so far, including automatic packaging of pre-builts.

https://github.com/static-php/static-php-cli-hosted/actions/runs/14156711781

.github/workflows/build-unix.yml
```diff
@@ -77,9 +79,19 @@ jobs:
...
+              ;;
+            linux-x86_64-glibc)
+              DOWN_CMD="./bin/spc-gnu-docker download"
+              BUILD_CMD="./bin/spc-gnu-docker build"
+              RUNS_ON="ubuntu-22.04"
+              ;;
+            linux-x86_64-glibc)
+              DOWN_CMD="./bin/spc-gnu-docker download"
+              BUILD_CMD="./bin/spc-gnu-docker build"
+              RUNS_ON="ubuntu-22.04-arm"
```
DubbleClick, line 94: typo should be linux-aarch64-glibc

config/pre-built.json
```diff
@@ -1,6 +1,6 @@
 {
     "repo": "static-php/static-php-cli-hosted",
     "prefer-stable": true,
-    "match-pattern": "{name}-{arch}-{os}.txz",
-    "suffix": "txz"
+    "match-pattern-linux": "{name}-{arch}-{os}-{libc}-{libcver}.txz",
+    "match-pattern": "{name}-{arch}-{os}.txz"
 }
```
DubbleClick, line 6: \n please

src/globals/defines.php
```diff
@@ -41,8 +41,9 @@
 ];
 
 // spc lock type
-const SPC_LOCK_SOURCE = 1;      // lock source
-const SPC_LOCK_PRE_BUILT = 2;   // lock pre-built
+const SPC_DOWN_SOURCE = 1;      // lock source
+const SPC_DOWN_PRE_BUILT = 2;   // lock pre-built
+const SPC_DOWN_PACKAGE = 3; // lock as package
```
DubbleClick, line 46: I think we should spell these out. I wasn't sure what DOWN referred to immediately.

src/globals/test-extensions.php
```diff
@@ -33,14 +35,14 @@
 $no_strip = false;
 
 // compress with upx
-$upx = false;
+$upx = true;
```
DubbleClick, line 38: this should still default to false imo

DubbleClick (changes requested):

src/globals/test-extensions.php
```diff
@@ -33,14 +35,14 @@
...
 
 // prefer downloading pre-built packages to speed up the build process
-$prefer_pre_built = false;
+$prefer_pre_built = true;
```
DubbleClick, line 41: after some consideration, I believe this should also be false by default. I constantly switch between local musl, local glibc, alpine docker and gnu docker builds

src/globals/test-extensions.php
crazywhalecc: Just a test, no need to be sensitive to the file. We can go back in next PR.

src/globals/test-extensions.php
crazywhalecc: Same as above

DubbleClick (approved):

Issue #687, opened 2025-03-31
Using in vendored mode with additional internal extensions


yoramdelangen:
I want to start using the project in vendored mode and add our own custom PHP extension. This means we should add additional 'sources' and 'ext' items to have a streamlined process with the project.

Is this already possible to achief? If so could you guide me a little or point me in a direction?

After digging around I think we need to add additional configuration into the `BuilderPovider` instance. I coulnd't find a proper way yet to add the addional configuration yet.

DubbleClick:
Do you want to compile your custom extensions in statically or dynamically?

crazywhalecc:
If you mean adding an extension, you can fork the project and modify the configuration file to add it.

Currently there is no other way to add extensions dynamically, because there is no ABI specifically designed for vendor mode or spc mode.

DubbleClick:
I think the simplest solution here would be to use the spc-gnu-docker and load the extensions dynamically.

yoramdelangen:
Statically would be preferable, but currently I have working version dynamically. Both is fine at this moment. Forking the project is currently the way I have done it as well, but found it very inconvinient keeping the fork up-to-date with upstream changes. I would like to have composer resolve that part for me instead of my own fork.

Let me see if i can conjure up some very minimalisc changes to "extend" current configuration. 

I was thinking about something in these lines:
- Use the project in "vendor" mode, via `composer require crazywhalecc/static-php-cli`.
- As explained in the [vendor library mode](https://static-php.dev/en/develop/structure.html#vendor-library-mode-vendor) documentation.
- Load up customizations ontop of the generated builder instance.

crazywhalecc:
It's a possible approach, I'll think about it.

crazywhalecc:
I have always been very conservative about APIs. It would be difficult to keep the API unchanged if some changes were made in the future.

When I was designing 2.0.0 in the early days, I planned to use a process-oriented single-file approach to declare extensions and libraries, which might be something like this:

```php
<?php

$lib = spc_register_library('testlib', config: ['static-libs-unix' => ['libtestlib.a']], source: ['type' => 'local', 'path' => 'testlib']);

$lib->onLinuxBuild(function ($lib) {
    shell()->cd($lib->getSourceDir())
        ->exec('./configure --enable-static --disable-shared --prefix=' . BUILD_ROOT_PATH)
        ->exec('make -j' . $lib->getBuilder()->concurrency)
        ->exec('make install');
});

[...]

PR #691, opened 2025-04-04
Vendor mode in a custom project, with custom extensions

yoramdelangen:
## What does this PR do?
Minor changes for allowing to use the library in a custom project instead of forking and adjusting code within the "project base".

Thread of the question can be found in #687.

Added the following options to adjust:
- Adding a custom command
- Allow for custom changes for `BuilderProvider` after initializing.
- By adding the following tree/namespace you can add custom extensions project-based without forking:

```bash
➜  poc-vendored-php-static-cli git:(main) tree --gitignore
.
├── bin
│   └── spc
├── composer.json
├── composer.lock
├── config
│   ├── ext.json
│   ├── lib.json
│   └── source.json
└── src
    ├── ConsoleApplication.php
    ├── MyCommand.php
    └── builder
        ├── extension
        │   └── snowflake.php
        ├── linux
        │   └── library
        │       └── snowflake.php
        ├── macos
        │   └── library
        │       └── snowflake.php
        └── unix
            └── library
                └── snowflake.php
```

Project example can be found here. In the example I have added `pdo_snowflake` extension, because I need it, and it gives a good example. 

Compling works, only the testing part fails at the moment due to PHP not using the `pdo_snowflake.so` file.

## Checklist before merging

### DIFF
--- a/src/SPC/builder/BuilderProvider.php
+++ b/src/SPC/builder/BuilderProvider.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace SPC\builder;
 
+use Closure;
 use SPC\builder\freebsd\BSDBuilder;
 use SPC\builder\linux\LinuxBuilder;
 use SPC\builder\macos\MacOSBuilder;
@@ -19,6 +20,15 @@ use Symfony\Component\Console\Input\InputInterface;
 class BuilderProvider
 {
     private static ?BuilderBase $builder = null;
+
+    private static ?Closure $customizeBuilder = null;
+    /**
+     * @param Closure(BuilderBase): void $callback
+     */
+    public static function customize(Closure $callback): void
+    {
+        self::$customizeBuilder = $callback;
+    }
 
     /**
      * @throws FileSystemException
@@ -36,6 +46,10 @@ class BuilderProvider
             'BSD' => new BSDBuilder($input->getOptions()),
             default => throw new WrongUsageException('Current OS "' . PHP_OS_FAMILY . '" is not supported yet'),
         };
+        // allow to add/customize builder instance early
+        if (self::$customizeBuilder) {
+            call_user_func_array(self::$customizeBuilder, [&self::$builder]);
+        }
         return self::$builder;
     }
 
--- a/src/SPC/ConsoleApplication.php
+++ b/src/SPC/ConsoleApplication.php
@@ -30,7 +30,7 @@ use Symfony\Component\Console\Application;
 /**
  * static-php-cli console app entry
  */
-final class ConsoleApplication extends Application
+class ConsoleApplication extends Application
 {
     public const VERSION = '2.5.1';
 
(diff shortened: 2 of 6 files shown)


src/SPC/builder/BuilderProvider.php
yoramdelangen, line 1: These changes can be discard if customization is not really necessary

yoramdelangen: Honestly, I dont really like these changes... but usable if really needed

yoramdelangen:
I mainly aim these changes to use for extensions like snowflake and a extension we created our self, that is compiled in a similar fashion as Snowflake. In case other people want to have their extensions available for the project without forking, this would be the easiest solution in stead of having a fork that can conflict on every commit pulling from upstream.

DubbleClick:
I'll leave both of your PR's up to @crazywhalecc, he's the owner of the project and I don't see myself in a position to make decisions crucial to the project structure or workflow on my own. He'll be back from vacation soon!

I really like this PR though!

crazywhalecc:
Unfortunately I ended my vacation a long time ago, but at the same time I was extremely busy at my day job.🥲

Don't worry though, I should have time to resolve remaining issues this week.

DubbleClick:
Ah welcome back, I hope you had a good time. I'm really excited for the shared extensions, once it goes live we'll have full package support for frankenphp + frankenphp-xdebug and frankenphp-ffi in around a week.

crazywhalecc:
I'm not sure how much customization people will have for private extensions, but I think we should add skeleton commands once vendor mode is officially available.

At the same time, for `xxx.json`, I also plan to implement `xxx.custom.json`, just like `env.ini`. So this may not be a simple update.

crazywhalecc:
I will take over this PR after thinking carefully about the implementation of vendor mode. Once everything is ready, I am open to reviews from anyone.

In the meantime, I have almost finished the commands to generate skeletons of extensions and libraries: #715, which also helps with vendor mode.

crazywhalecc:
I think we can start this now. The main work can be divided into several parts:

- [Hard] Write development docs
- [Easy] Declare internal callable APIs
- [Hard] Write skeleton generation commands (WIP)

In addition, I think it may be possible to implement other binary builds in vendor mode. For example, we can reuse spc dependencies to build static python, static vim and other components. Maybe no one will need them, but it should be feasible.

As for single-file build support I mentioned before, I feel like now is not the right time and might cause maintenance chaos instead. In the future, we can try to extend based on the current vendor model, and it may be better to use SPC as a base framework to develop.

**TLDR**: For the current PR, I will make a small amount of modifications in the future to support the most basic vendor functions. After that, the vendor mode support of the skeleton command will be improved.

henderkes:
Yes, this will actually also help quite a bit with spc-packages.  I ended up rewriting a lot of sloppy logic to substitute behaviour already implemented in spc. Which is kind of fine for the scope the project has at the moment or that is planned, but I might do a proper rewrite with vendored spc mode once it exists.

crazywhalecc:
Given that v3 is to be refactored, and v3 natively supports vendor mode and external mode, the code in this PR is almost unusable, I'll close this PR. After v3 released everyone can use new vendor mode to extend abilities.

PR #693, opened 2025-04-12
explicitly state libheif -> --with-heic

DubbleClick:
## What does this PR do?

explicitly states --with-heic when using libheif (still default)

Maybe we should make libheic optional though, to see how it plays out with:
https://github.com/dunglas/frankenphp/issues/1487

### DIFF
--- a/src/SPC/builder/unix/library/imagemagick.php
+++ b/src/SPC/builder/unix/library/imagemagick.php
@@ -28,6 +28,7 @@ trait imagemagick
             'libpng' => 'png',
             'libwebp' => 'webp',
             'libxml2' => 'xml',
+            'libheif' => 'heic',
             'zlib' => 'zlib',
             'xz' => 'lzma',
             'zstd' => 'zstd',


crazywhalecc:
Is the current problem that heif cannot be used in imagick? I am not sure the purpose of doing this.

If it is to be an optional dependency, then gd should also set it as an optional dependency.

DubbleClick:
No, the problem is that we are relying on imagemagick behaving against spec qnd automatically including heic support even if it wasn't configured to. We shouldbe explicit about heic support by passing --with-heic

crazywhalecc:
Makes sense. But we should also modify lib.json.

DubbleClick:
I haven't made it optional yet, that could be another PR if we really want. Most people using imagick would probably want heif support so it's ok to keep it a dependency.

I believe the core issue about the frankenphp crash was the -march-native optimisation that we added to the static-builder dockerfiles, not the libde265 itself.

crazywhalecc (approved):

PR #696, opened 2025-04-14
don't set EXTENSION_DIR by default

DubbleClick:
## What does this PR do?

doesn't set the variable to an empty string by default, because it makes the build command look ugly:

image


crazywhalecc (approved):

DubbleClick:
it's not a bug, just a small pet peeve 😅

PR #701, opened 2025-04-18
add mirror sources to savannah downloads

DubbleClick:
## What does this PR do?

the git download for savannahs git system fails very often, use tgz downloads with an optional alternative source instead

### DIFF
--- a/config/source.json
+++ b/config/source.json
@@ -37,9 +37,12 @@
         }
     },
     "attr": {
-        "type": "git",
-        "rev": "v2.5.2",
-        "url": "https://git.savannah.nongnu.org/git/attr.git",
+        "alt": {
+            "type": "url",
+            "url": "https://mirror.souseiseki.middlendian.com/nongnu/attr/attr-2.5.2.tar.gz"
+        },
+        "type": "url",
+        "url": "https://download.savannah.nongnu.org/releases/attr/attr-2.5.2.tar.gz",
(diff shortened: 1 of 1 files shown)


config/source.json
```diff
@@ -37,9 +37,12 @@
         }
     },
     "attr": {
-        "type": "git",
-        "rev": "v2.5.2",
-        "url": "https://git.savannah.nongnu.org/git/attr.git",
+        "alt": {
+            "type": "url",
+            "url": "https://mirror.souseiseki.middlendian.com/nongnu/attr/attr-2.5.2.tar.gz"
```
DubbleClick, line 42: yes, this is an official mirror linked to from savannahs download page

crazywhalecc: Move original downloads to `alt` will be good. We dont' have `alt` download option yet.

DubbleClick: I know, preparing for the future. Their downloads page seems more stable than the git. I think it's fine the way it is.

crazywhalecc (approved):

crazywhalecc:
Since downloading is the most common problem and this bug is uncontrollable, I think we should regularly back up a mirror package of each source to dl.static-php.dev, and if the source address fails to download, fallback to the mirror site by default (use alt for sites that provide alt), and add a `--disable-alt`. @DubbleClick

PR #713, opened 2025-04-24
Add build from config support (craft command)

### LINKED ISSUE #702: Better build UX
When I was modifying test-extensions.php today, I suddenly realized that the previous [simplified UX issue](https://github.com/crazywhalecc/static-php-cli/issues/187) seems to be solved by specifying a standard parameter input interface.

In other words, the parameters of the three build-related commands are expressed in the form of build-suite.yml or config.mak or other files, so that the desired results can be built with one command, and the command calls the build parameters internally. At the same time, it can also provide a direction for the future [setup-static-php](https://github.com/static-php/setup-static-php).

crazywhalecc:
## What does this PR do?

Fix #702 

## Checklist before merging


.editorconfig
```diff
@@ -0,0 +1,15 @@
...
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false
+
+[*.{yml,yaml, vue}]
+indent_size = 2
```
DubbleClick, line 15: ❤️

DubbleClick (commented):
I'm sadly not familiar with craft so I can't give an accurate review yet. If you don't mind waiting, I will have time on the weekend to test and give a more thought-through review.

henderkes:
Seems all good so far, except that for some reason I'm running into 60 second timeouts when running `bin/spc craft`. I doubt that's related to this PR though.

crazywhalecc:
> Seems all good so far, except that for some reason I'm running into 60 second timeouts when running bin/spc craft. I doubt that's related to this PR though.

@henderkes I introduced symfony/process to call subcommand,but in theory subcommands should not cause unexpected timeouts.

Either that or having to re-download php-src every time is causing this problem.

henderkes:
image

src/SPC/command/CraftCommand.php
```diff
@@ -0,0 +1,164 @@
...
+        }
+        $prefix = PHP_SAPI === 'cli' ? [PHP_BINARY, $argv[0]] : [$argv[0]];
+
+        $process = new Process([...$prefix, $cmd, '--no-motd', ...$args], timeout: null);
+        $this->log("Running: {$process->getCommandLine()}", true);
+
+        if (PHP_OS_FAMILY === 'Windows') {
+            sapi_windows_set_ctrl_handler(function () use ($process) {
+                $process->signal(SIGINT);
+            });
```
henderkes, line 152: SIGINT is defined in ext-pcntl
$process->signal sends a POSIX signal. I don't think Windows has POSIX signals, or signals at all?

crazywhalecc: Good catch. Let me look for a solution for Windows.

crazywhalecc: Windows does not support signals, but `$process->signal()` is supported on Windows. When it detects that the directory separator is `\`, it calls taskkill.

henderkes (approved):

Issue #686, opened 2025-03-31
create our own gcc toolchain for compiling spc-gnu-docker


henderkes:
To allow statically linking against libgomp, libstdc++ and so on in -static-pie builds.

Luckily this was super easy to do inside the spc-gnu-docker image itself:

```console
yum install -y wget
wget https://ftp.gwdg.de/pub/misc/gcc/releases/gcc-14.2.0/gcc-14.2.0.tar.gz
tar xzf gcc-14.2.0.tar.gz
cd gcc-14.2.0
./contrib/download_prerequisites
cd ..
mkdir objdir
cd objdir
$PWD/../gcc-14.2.0/configure   --prefix=$HOME/GCC-14   --enable-languages=c,c++,go   --enable-default-pie   --enable-default-ssp   --enable-shared   --enable-static   --enable-threads=posix   CFLAGS="-O2 -fPIC -fPIE -g0 -static-libgcc"   CXXFLAGS="-O2 -fPIC -fPIE -g0 -static-libgcc -static-libstdc++" --disable-multilib
make -j4
make install
```

DubbleClick:
This might actually be a lot more complex because apparently with gcc 11 there was a glibcxx versioning bump in the stdc++ init code, requiring a more recent glibcxx version than installed by default.

I'm trying to patch the libstdc++ gnu.ver to possibly get around that, but we'll have to see how much success I have with it.

If that doesn't succeed we'd have to fumble around with LD_PRELOAD path or build a gcc 10.

DubbleClick:
Okay I gave up on gcc 11+.

```console
../gcc-10.2.0/configure  \
  --prefix=/opt/gcc10 \
  --enable-languages=c,c++ \
  --disable-multilib  \
  --enable-default-pie \
  --enable-default-ssp  \
  --enable-shared  \
  --enable-static \
  --enable-threads=posix  \
  --without-system-zlib \
  --with-default-libstdcxx-abi=gcc4-compatible  \
  CFLAGS="-O2 -fPIC -fPIE -g0"  \
  CXXFLAGS="-O2 -fPIC -fPIE -g0"
```

Works fine.

crazywhalecc:
> Luckily this was super easy to do inside the spc-gnu-docker image itself

But it builds too slow. My M3 Pro mac with arm docker needs at least 15min. Publishing pre-built content is more reasonable.

DubbleClick:
[...]

PR #727, opened 2025-05-15
gnu gcc builder script #686

henderkes:
## What does this PR do?

#686

### DIFF
--- a/bin/spc-gnu-build-gcc
+++ b/bin/spc-gnu-build-gcc
@@ -1,0 +1,195 @@
+#!/usr/bin/env bash
+
+# This file is using docker to build gcc and binutils and copy them to the host
+set -e
+
+# Set default GCC version if not provided
+if [ -z "$GCC_VERSION" ]; then
+    GCC_VERSION="10.5.0"
+fi
+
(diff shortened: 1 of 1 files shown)


henderkes:
And the second problem I see is that gcc is a massive download. And we need to download the full gcc tree to build libgomp (even though we don't need to build full gcc, just building libgomp is ok).

henderkes:
It's doable that way, but I think using our own gcc toolchain, like we do for musl, is the better choice. that way we can explicitly support 1 compiler.

crazywhalecc:
It looks like we should add it to `static-php-cli-hosted` and docs. Maybe it would be better to change this PR to update the documentation part and implement the main script part in static-php-cli-hosted.

henderkes:
Yes, this has no place being in spc, it should be treated the same way as the musl toolchain.

It would be nice if we could get gcc 15 to work instead, anyhow. I did compile it and successfully use it on my rhel 10 host, but had issues with >=11 in the centos 7 docker when I first tackled this.

henderkes:
looking to switch to zig for better cross compilation instead

PR #731, opened 2025-05-20
Fix GHA token missing bug

### LINKED ISSUE #730: actions upx download fails with 403
probably not passing the github token correctly in the curl command

crazywhalecc:
## What does this PR do?

Fix #730 

## Checklist before merging

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -186,11 +186,12 @@ jobs:
       - name: "Prepare UPX for Linux"
         if: ${{ startsWith(matrix.os, 'ubuntu-') }}
         run: |
+          export GITHUB_TOKEN=${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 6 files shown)


henderkes (approved):

henderkes:
maybe try explicitly setting it in the command? instead of only in the env

crazywhalecc:
> maybe try explicitly setting it in the command? instead of only in the env

@henderkes Ah, the bug is actually very simple: `passthru` does not let the environment variable pass through....

crazywhalecc:
I kind of hate docker. The reason is that env can't penetrate into `bin/spc-alpine-docker`.

henderkes:
> I kind of hate docker. The reason is that env can't pass into `bin/spc-alpine-docker`.

I hate docker, too. We're in the minority, though 😢

henderkes:
You can work around it by adding an argument GITHUB_TOKEN and then setting ENV GITHUB_TOKEN $GITHUB_TOKEN

PR #733, opened 2025-05-26
Fix CraftCommand ignoring build-options during build

### LINKED ISSUE #732: CraftCommand does not forward build-options to build command
Given the following `craft.yml`:
```yaml
php-version: 8.4
extensions: "curl"
sapi: cli
build-options:
  with-clean: false
  prefer-pre-built: true
  with-config-file-path: /custom/config/path
```

I'd expect the following build process to be executed:
```sh
'/usr/local/bin/php' 'bin/spc' 'build' '--no-motd' 'curl' '--with-libs=' '--build-cli' '--with-suggested-libs' '--with-config-file-path="/custom/config/path"'
```

But the actual process looks like this:
```sh
'/usr/local/bin/php' 'bin/spc' 'build' '--no-motd' 'curl' '--with-libs=' '--build-cli'
```

I'll shortly push a fix.

vworldat:
## What does this PR do?

At the moment, it's impossible to pass build-options when using `spc craft` (see #732). This PR fixes that, allowing the following use cases:

- boolean `true` values such as `--with-suggested-libs` are passed without parameters, assuming `InputOption::VALUE_NONE` behaviour.
- boolean `false` values (e.g. when setting `with-suggested-libs: false` in `yaml`) are omitted since Symfony console treats those values as opt-in only.
- array values are treated as "add this option multiple times". Tested with `with-hardcoded-ini`.
- everything else is passed as `--param=value`.

Works fine locally both with `spc` as well as `spc-alpine-docker`.

## Checklist before merging

Closes #732

### DIFF
--- a/src/SPC/command/CraftCommand.php
+++ b/src/SPC/command/CraftCommand.php
@@ -79,20 +79,7 @@ class CraftCommand extends BaseCommand
                     $craft['download-options']['ignore-cache-sources'] .= ',php-src';
                 }
             }
-            $download_options = $craft['download-options'];
-            foreach ($download_options as $option => $val) {
-                if (is_bool($val) && $val || is_null($val)) {
-                    $args[] = "--{$option}";
-                } elseif (is_string($val)) {
-                    $args[] = "--{$option}={$val}";
-                } elseif (is_array($val)) {
-                    foreach ($val as $v) {
-                        if (is_string($v)) {
-                            $args[] = "--{$option}={$v}";
-                        }
-                    }
-                }
-            }
+            $this->optionsToArguments($craft['download-options'], $args);
             $retcode = $this->runCommand('download', ...$args);
             if ($retcode !== 0) {
                 $this->output->writeln('<error>craft download failed</error>');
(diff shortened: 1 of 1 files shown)


henderkes:
looks good to me

src/SPC/command/CraftCommand.php
```diff
@@ -104,6 +104,25 @@ public function handle(): int
...
+                if (is_array($value)) {
+                    foreach ($value as $subVal) {
+                        $args[] = "--{$option}={$subVal}";
+                    }
+
+                    continue;
+                }
+
+                $args[] = "--{$option}={$value}";
+            }
```
crazywhalecc, line 125: https://github.com/crazywhalecc/static-php-cli/blob/ca3d6d14d0b8df5ad04b881fbdaa50f6ad661a3f/src/SPC/command/CraftCommand.php#L83-L95

Seems we could extract this option parsing in a separate function, just re-use this section.

crazywhalecc (changes requested):

PR #736, opened 2025-05-31
libgomp support instead of custom built gcc toolchain?

henderkes:
## What does this PR do?
alternative to building our own gcc

### DIFF
--- a/config/lib.json
+++ b/config/lib.json
@@ -148,6 +148,17 @@
             "IOKit"
         ]
     },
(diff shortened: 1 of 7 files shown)


henderkes:
let's remove openmp support entirely, system packages distribute it disabled, the ini disables it by default and the authors recommend disabling it. WIth a distinction between libgomp, libomp and the non-existent usable static libraries for them, it's just not worth it.

PR #739, opened 2025-06-04
force curl to respect SSL_CERT_FILE

henderkes:
## What does this PR do?

There has been a problem in recent versions where SSL_CERT_FILE, SSL_CERT_DIR and CURL_CA_BUNDLE are not respected.
Even though it's not documented, it seems like it's no longer possible to let openssl handle the location unless the paths are disabled in the curl build.

For what it's worth, it still finds the ca bundle even if the environment variables are not set, I believe due to openssl's logic around it.

### DIFF
--- a/src/SPC/builder/unix/library/curl.php
+++ b/src/SPC/builder/unix/library/curl.php
@@ -18,7 +18,15 @@ trait curl
     {
         $extra = '';
         // lib:openssl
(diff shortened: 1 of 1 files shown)


henderkes:
This should fix problems like https://github.com/dunglas/frankenphp/issues/1610 and https://github.com/dunglas/frankenphp/issues/1627.

dunglas (approved):

crazywhalecc (approved):

PR #741, opened 2025-06-04
force minimum version 3.5 for cmake policies in freetype

henderkes:
## Checklist before merging


henderkes:
or should we specify 3.12 here?

henderkes:
We cannot merge this. The option was only added in 4.0, it may fail on older versions:

https://cmake.org/cmake/help/latest/variable/CMAKE_POLICY_VERSION_MINIMUM.html

henderkes:
Should be okay now, doesn't change anything for cmake < 4.0

PR #743, opened 2025-06-04
Feature: add http/3 support to curl.

henderkes:
## What does this PR do?
work in progress, adds experimental http3 support to libcurl

## Checklist before merging

### DIFF
--- a/src/SPC/builder/unix/library/nghttp3.php
+++ b/src/SPC/builder/unix/library/nghttp3.php
@@ -1,4 +1,4 @@
-﻿<?php
+<?php
 
 declare(strict_types=1);
 
--- a/src/SPC/builder/unix/library/nghttp3.php
+++ b/src/SPC/builder/unix/library/nghttp3.php
@@ -26,7 +26,7 @@ trait nghttp3
             ->setEnv([
                 'CFLAGS' => $this->getLibExtraCFlags(),
                 'LDFLAGS' => $this->getLibExtraLdFlags(),
-                'LIBS' => $this->getLibExtraLibs()
+                'LIBS' => $this->getLibExtraLibs(),
             ])
             ->execWithEnv(
                 './configure ' .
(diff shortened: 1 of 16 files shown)


henderkes:
```php
<?php

$ch = curl_init('https://www.google.de');
curl_setopt($ch, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_3);
curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
$response = curl_exec($ch);

$proto = curl_getinfo($ch, CURLINFO_HTTP_VERSION);

switch ($proto) {
    case CURL_HTTP_VERSION_3:
        echo "HTTP/3 used\n";
        break;
    case CURL_HTTP_VERSION_2:
        echo "Fell back to HTTP/2\n";
        break;
    case CURL_HTTP_VERSION_1_1:
        echo "Fell back to HTTP/1.1\n";
        break;
    default:
        echo "Unknown HTTP version\n";
}

curl_close($ch);
```

```console
[m@M-TH static-php-cli]$ buildroot/bin/php ssl.php
HTTP/3 used
```

src/SPC/builder/unix/library/nghttp3.php
```diff
@@ -0,0 +1,45 @@
...
+                '--with-pic ' .
+                '--enable-lib-only ' .
+                $args . ' ' .
+                '--prefix='
+            )
+            ->execWithEnv('make clean')
+            ->execWithEnv("make -j{$this->builder->concurrency}")
+            ->execWithEnv('make install DESTDIR=' . BUILD_ROOT_PATH);
+        $this->patchPkgconfPrefix(['libnghttp3.pc']);
+    }
```
henderkes, line 44: apart from the available AutoconfArgs, all three libraries use the exact same code to build

Maybe we should start thinking about defining the method once somewhere and only passing the available options to it.

PR #726, opened 2025-05-15
Fix PIC errors when building libphp.so | extensive extension building rework

henderkes:
## What does this PR do?

fix building of shared libphp.so with icu and a few other extensions

closes https://github.com/crazywhalecc/static-php-cli/issues/738
closes https://github.com/crazywhalecc/static-php-cli/issues/735
closes https://github.com/crazywhalecc/static-php-cli/issues/725
closes https://github.com/crazywhalecc/static-php-cli/issues/720
closes https://github.com/crazywhalecc/static-php-cli/issues/749
closes https://github.com/crazywhalecc/static-php-cli/issues/744
enables https://github.com/crazywhalecc/static-php-cli/issues/724

## Checklist before merging

> If your PR involves the changes mentioned below and completed the action, please tick the corresponding option.
> If a modification is not involved, please skip it directly.

- [x] If you modified `*.php`, run `composer cs-fix` at local machine.

PR #750, opened 2025-06-08
Feat: UnixCMakeExecutor

crazywhalecc:
## What does this PR do?

Replace all cmake builder to cmake executor. After refactoring we can control all cmake build process instead of updating every single file every time. This is one of prerequisites for #726 . 

After this PR is merged, it will continue to:

- UnixAutoconfExecutor.
- New `setDefaultEnv()` for UnixShell to replace the redundant `setEnv(['XXXFLAGS'])`, remove redundant `execWithEnv()`.
- WindowsCMakeExecutor.

## Checklist before merging

- [ ] Test all cmake libraries.

### DIFF
--- a/src/SPC/builder/macos/library/glfw.php
+++ b/src/SPC/builder/macos/library/glfw.php
@@ -19,7 +19,7 @@ class glfw extends MacOSLibraryBase
     protected function build(): void
     {
         UnixCMakeExecutor::create($this)
-            ->setCMakeBuildDir("{$this->source_dir}/vendor/glfw")
+            ->setBuildDir("{$this->source_dir}/vendor/glfw")
             ->setReset(false)
             ->addConfigureArgs(
                 '-DGLFW_BUILD_EXAMPLES=OFF',
--- a/src/SPC/builder/unix/library/brotli.php
+++ b/src/SPC/builder/unix/library/brotli.php
@@ -18,7 +18,7 @@ trait brotli
     protected function build(): void
     {
         UnixCMakeExecutor::create($this)
-            ->setCMakeBuildDir("{$this->getSourceDir()}/build-dir")
+            ->setBuildDir("{$this->getSourceDir()}/build-dir")
             ->build();
 
         $this->patchPkgconfPrefix(['libbrotlicommon.pc', 'libbrotlidec.pc', 'libbrotlienc.pc']);
--- a/src/SPC/builder/unix/library/libaom.php
+++ b/src/SPC/builder/unix/library/libaom.php
@@ -17,7 +17,7 @@ trait libaom
     protected function build(): void
     {
         UnixCMakeExecutor::create($this)
-            ->setCMakeBuildDir("{$this->source_dir}/builddir")
+            ->setBuildDir("{$this->source_dir}/builddir")
             ->addConfigureArgs('-DAOM_TARGET_GPU=generic')
             ->build();
         $this->patchPkgconfPrefix(['aom.pc']);
--- a/src/SPC/builder/unix/library/snappy.php
+++ b/src/SPC/builder/unix/library/snappy.php
@@ -17,7 +17,7 @@ trait snappy
     protected function build(): void
     {
         UnixCMakeExecutor::create($this)
-            ->setCMakeBuildDir("{$this->source_dir}/cmake/build")
+            ->setBuildDir("{$this->source_dir}/cmake/build")
             ->addConfigureArgs(
                 '-DSNAPPY_BUILD_TESTS=OFF',
                 '-DSNAPPY_BUILD_BENCHMARKS=OFF',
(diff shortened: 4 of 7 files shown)


src/SPC/builder/macos/library/glfw.php
```diff
@@ -17,11 +18,14 @@ class glfw extends MacOSLibraryBase
      */
     protected function build(): void
     {
-        // compile！
-        shell()->cd(SOURCE_PATH . '/ext-glfw/vendor/glfw')
-            ->exec("cmake . {$this->builder->makeCmakeArgs()} -DBUILD_SHARED_LIBS=OFF -DGLFW_BUILD_EXAMPLES=OFF -DGLFW_BUILD_TESTS=OFF")
-            ->exec("make -j{$this->builder->concurrency}")
-            ->exec('make install');
+        UnixCMakeExecutor::create($this)
+            ->setCMakeBuildDir("{$this->source_dir}/vendor/glfw")
```
henderkes: Rename to just `setBuildDir`. We don't need to specify cmake build dir in the cmake executor.

src/SPC/builder/macos/library/glfw.php
```diff
@@ -17,11 +18,14 @@ class glfw extends MacOSLibraryBase
...
+            ->setReset(false)
+            ->addConfigureArgs(
+                '-DGLFW_BUILD_EXAMPLES=OFF',
+                '-DGLFW_BUILD_TESTS=OFF',
+            )
+            ->build('.');
```
henderkes, line 28: Why .?

src/SPC/builder/unix/library/libzip.php
```diff
@@ -16,35 +16,21 @@ trait libzip
...
-                '-DBUILD_DOC=OFF ' .
-                '-DBUILD_EXAMPLES=OFF ' .
-                '-DBUILD_REGRESS=OFF ' .
-                '-DBUILD_TOOLS=OFF ' .
-                $extra .
-                '..'
+        UnixCMakeExecutor::create($this)
+            ->optionalLib('bzip2', ...cmake_boolean_args('ENABLE_BZIP2'))
+            ->optionalLib('xz', ...cmake_boolean_args('ENABLE_LZMA'))
+            ->optionalLib('openssl', ...cmake_boolean_args('ENABLE_OPENSSL'))
```
henderkes, line 22:
```suggestion
            ->optionalLib('openssl', ...cmake_boolean_args('ENABLE_OPENSSL'))
            ->optionalLib('zstd', ...cmake_boolean_args('ENABLE_ZSTD'))
```

src/SPC/builder/unix/library/libzip.php
```diff
@@ -16,35 +16,21 @@ trait libzip
...
+            ->addConfigureArgs(
+                '-DENABLE_GNUTLS=OFF',
+                '-DENABLE_MBEDTLS=OFF',
+                '-DBUILD_SHARED_LIBS=OFF',
+                '-DBUILD_DOC=OFF',
+                '-DBUILD_EXAMPLES=OFF',
+                '-DBUILD_REGRESS=OFF',
+                '-DBUILD_TOOLS=OFF',
+                '-DENABLE_ZSTD=OFF',
```
henderkes:
```suggestion
```

henderkes (changes requested):

henderkes:
Could you also change execWithEnv to put the environment silently in prod mode and only print the executed command without environment flags? For debug mode I think we should print *all* environment flags we use, so pkg_config and pkg_config_path as well.

src/SPC/builder/macos/library/glfw.php
crazywhalecc: I haven't tested yet. The previous version is using `.`.

src/SPC/builder/macos/library/glfw.php
crazywhalecc: Makes sense.

crazywhalecc:
> Could you also change execWithEnv to put the environment silently in prod mode and only print the executed command without environment flags? For debug mode I think we should print _all_ environment flags we use, so pkg_config and pkg_config_path as well.

I will do it next PR.

henderkes:
Could you base the next PR off of my PR by chance? This one will have quite a few conflicts, but the next will have even more.

crazywhalecc:
> Could you base the next PR off of my PR by chance? This one will have quite a few conflicts, but the next will have even more.

I plan to start resolving #726 conflicts after the next setEnv PR. This will make it clearer how we should modify the giant PR.

I've already done part of it, and it seems like basing the next PR on it will lead to even more confusion. Maybe breaking up parts of #726 into smaller PRs and merging them into the main branch first would be clearer.

henderkes:
Yeah that's fair.

henderkes (approved):

PR #751, opened 2025-06-09
Extract default build env to unix shell

crazywhalecc:
## What does this PR do?

Extract default build env to unix shell. Simplify shell command execute.

## Checklist before merging

### DIFF
--- a/src/SPC/builder/linux/library/libffi.php
+++ b/src/SPC/builder/linux/library/libffi.php
@@ -21,8 +21,8 @@ class libffi extends LinuxLibraryBase
         $arch = getenv('SPC_ARCH');
 
         shell()->cd($this->source_dir)
-            ->setEnv(['CFLAGS' => $this->getLibExtraCFlags(), 'LDFLAGS' => $this->getLibExtraLdFlags(), 'LIBS' => $this->getLibExtraLibs()])
-            ->execWithEnv(
+            ->initLibBuildEnv($this)
+            ->exec(
                 './configure ' .
                 '--enable-static ' .
                 '--disable-shared ' .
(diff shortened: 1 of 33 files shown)


src/SPC/builder/linux/library/libffi.php
```diff
@@ -21,8 +21,8 @@ public function build(): void
         $arch = getenv('SPC_ARCH');
 
         shell()->cd($this->source_dir)
-            ->setEnv(['CFLAGS' => $this->getLibExtraCFlags(), 'LDFLAGS' => $this->getLibExtraLdFlags(), 'LIBS' => $this->getLibExtraLibs()])
-            ->execWithEnv(
+            ->initLibBuildEnv($this)
```
henderkes: initializeEnv

it's the same between libraries and extensions.

src/SPC/util/UnixShell.php
```diff
@@ -48,7 +51,41 @@ public function exec(string $cmd): UnixShell
...
+
+    /**
+     * Init the environment variable that common build will be used.
+     *
+     * @param BSDLibraryBase|LinuxLibraryBase|MacOSLibraryBase $library Library class
+     */
+    public function initLibBuildEnv(BSDLibraryBase|LinuxLibraryBase|MacOSLibraryBase $library): UnixShell
+    {
+        $this->setEnv([
+            'CFLAGS' => $library->getLibExtraCFlags(),
```
henderkes, line 70: We should add CXXFLAGS too. It won't break anything in C-ony libraries.

henderkes:
Hahahaha, the tests are running into relocation errors. Still blows my mind that non-relocatable code is even possible on Unix and that -fPIC isn't enough.

Once you patch up one, the next one will pop up. That's why the other PR is so big.

image

crazywhalecc:
> Hahahaha, the tests are running into relocation errors. Still blows my mind that non-relocatable code is even possible on Unix and that -fPIC isn't enough.

Seems to just be a problem caused by not placing the environment variables into the command correctly. Everything looks good to me.

henderkes:
Yeah, the whack-a-mole is only a thing when building shared libraries.

PR #754, opened 2025-06-09
Feat/autoconf executor

crazywhalecc:
## What does this PR do?

Add UnixAutoconfExecutor.

## Checklist before merging

- [ ] Test all autoconf-based libraries.

### DIFF
--- a/src/SPC/builder/unix/library/curl.php
+++ b/src/SPC/builder/unix/library/curl.php
@@ -6,7 +6,6 @@ namespace SPC\builder\unix\library;
 
 use SPC\exception\FileSystemException;
 use SPC\exception\RuntimeException;
-use SPC\store\FileSystem;
 use SPC\util\executor\UnixCMakeExecutor;
 
 trait curl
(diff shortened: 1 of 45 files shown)


src/SPC/builder/unix/library/gettext.php
```diff
@@ -4,36 +4,34 @@
...
+            $autoconf->addConfigureArgs('--disable-threads');
+        }
 
-        shell()->cd($this->source_dir)->initializeEnv($this)
-            ->appendEnv(['CFLAGS' => $cflags, 'LDFLAGS' => $ldflags])
-            ->exec(
-                './configure ' .
-                '--enable-static ' .
-                '--disable-shared ' .
-                '--enable-pic ' .
```
henderkes, line 25: We need to make sure to keep this everywhere. --with-pic too. Sadly there's no general standard. Or we just pass both by default and ignore the "unknown option" warnings?

henderkes (changes requested):

henderkes:
did you mean to change the target branch to main? this branch now has a mix of the other pr and this, I think it's better to merge into icurel branch again, also because that's where we can really test if the with-pic and co really worked. Since we don't build shared libs in main branch, we would never run into the relocation errors.

crazywhalecc:
> did you mean to change the target branch to main? this branch now has a mix of the other pr and this, I think it's better to merge into icurel branch again, also because that's where we can really test if the with-pic and co really worked. Since we don't build shared libs in main branch, we would never run into the relocation errors.

I change the base branch temporarily, only the master branch will trigger CI.

crazywhalecc:
@henderkes I may have found a bug, introduced three weeks ago. It may have caused a lot of your previous problems with missing libbrotlienc. 🫢

https://github.com/crazywhalecc/static-php-cli/blob/48fc0e5b162fc5032a418471f1c7617bfdab3074/src/SPC/builder/unix/library/brotli.php#L27

henderkes:
Oh yeah, good catch!

src/SPC/builder/unix/library/librdkafka.php
```diff
@@ -25,6 +25,10 @@ function ($lib) {
                 },
                 '--disable-zstd'
             )
+            ->ignoreConfigureArgs(
+                '--with-pic',
+                '--enable-pic',
+            )
```
henderkes: Does this lead to an error? it should only be a warning, which would be fine. There's no point manually specifying ->ignoreConfigureArgs everywhere

crazywhalecc: I haven't looked deeper at librdkafka's autoconf, but it doesn't seem to recognize --disable-option-warning or a similar option. So I had to exclude it manually, and currently CI shows only it and zlib as errors.

henderkes: Ah you're right, it actually errors out instead of printing a warning. How stupid.

henderkes: We should call that ->removeConfigureArgs then, however.

src/SPC/builder/unix/library/curl.php
```diff
@@ -37,6 +38,26 @@ protected function build(): void
...
+        FileSystem::replaceFileStr(
+            BUILD_LIB_PATH . '/pkgconfig/libcurl.pc',
+            'libbrotlienc,libbrotlidec,libbrotlicommon',
+            'libbrotlidec,libbrotlicommon'
+        );
+        FileSystem::replaceFileStr(
+            BUILD_LIB_PATH . '/pkgconfig/libcurl.pc',
+            'libbrotlidec,libbrotlicommon',
+            'libbrotlienc,libbrotlidec,libbrotlicommon'
+        );
```
crazywhalecc: I guess we don't need it anymore.

henderkes: yeah let's see what happens when we remove it

henderkes (approved):

Issue #759, opened 2025-06-11
The requested URL returned error: 403


lyrixx:
Hello, We have a permanent error in [castor run](https://github.com/jolicode/castor/actions/runs/15584119461/job/43886032160)

I made a simple reproducer

```console
>/tmp/scp curl -fsSL -o spc https://dl.static-php.dev/static-php-cli/spc-bin/nightly/spc-linux-x86_64

>/tmp/scp chmod +x ./spc

>/tmp/scp ./spc --version
static-php-cli 2.5.2

>/tmp/scp ./spc 'download' '--for-extensions=mbstring,phar,posix,tokenizer,curl,filter,openssl' '--with-php=8.3' '--prefer-pre-built'
     _        _   _                 _
 ___| |_ __ _| |_(_) ___      _ __ | |__  _ __
/ __| __/ _` | __| |/ __|____| '_ \| '_ \| '_ \
\__ \ || (_| | |_| | (_|_____| |_) | | | | |_) |
|___/\__\__,_|\__|_|\___|    | .__/|_| |_| .__/   v2.5.2
                             |_|         |_|
curl: (22) The requested URL returned error: 403
[12:07:12] [E] failed http fetch
```

1. It would be nice to know what URL returns a 403
    * I edited SPC, to get the URL, and retry it locally. I was rate limited 
2. And how to fix that
    * I [put](https://github.com/jolicode/castor/pull/647) a token in my ci.yaml, I think it could be cool to add it to the doc

Thanks

PR #760, opened 2025-06-11
Better exception

lyrixx:
refs : https://github.com/crazywhalecc/static-php-cli/issues/759


lyrixx:
Side note : I noticed you "implemented" your own logging system, http client, process runner.
I think it would have been better to rely more on symfony.
For example, I run the command `-vvv` to get all logs. But it didn't work since I had to use `--debug`. 
Another exemple is the process handler, We could have much more information when a process failed

henderkes:
That's a great idea, we will tackle it later.

PR #763, opened 2025-06-12
update documentation for glibc

henderkes:
## What does this PR do?

updates notes for glibc compilation


henderkes:
Please check if the chinese translation is acceptable. I have no choice but to trust translators.

PR #764, opened 2025-06-12
remove incorrect space after --enable-yac

henderkes:
## What does this PR do?

fix shared build error for yac


henderkes:
@crazywhalecc this one snuck in at the end, we should maybe update the 1.6.0 release to include it

Issue #765, opened 2025-06-12
ignore-cache-sources: php-src


henderkes:
While this does download the new version, when it builds, it still uses the already extracted, old version.

crazywhalecc:
This is expected. If you want to ensure a clean rebuild when debugging locally, you should use the `--with-clean` option.

henderkes:
`--with-clean` removes all the built libraries as well, though. I think when the php version that was last downloaded changes, a php-src + extension recompilation should automatically be triggered.

PR #767, opened 2025-06-12
patchelf and rename -release tagged extensions from php configure build

### LINKED ISSUE #771: FrankenPHP is now an official php SAPI
Should we include building it? It's quite simple actually, we only need go, xcaddy (and embed).

henderkes:
## What does this PR do?

the release option also affects shared extensions, which is unwanted, patchelf their soname back and rename them

Closes #771


src/SPC/doctor/item/LinuxToolCheckList.php
```diff
@@ -38,6 +40,7 @@ class LinuxToolCheckList
         'tar', 'unzip', 'gzip', 'gcc',
         'bzip2', 'cmake', 'patch', 'which',
         'xz', 'libtool', 'gettext-devel',
+        'perl', 'patchelf',
```
henderkes, line 43: I ran into this error every single time I set up a RHEL installation. perl is needed for perl-FindBin on rhel, which a lot of configure scripts from libraries need

src/SPC/builder/linux/LinuxBuilder.php
```diff
@@ -293,6 +293,24 @@ protected function buildEmbed(): void
...
+                    continue;
+                }
+                $name = $ext->getName();
+                $versioned = "{$name}-{$release}.so";
+                $unversioned = "{$name}.so";
+                if (is_file(BUILD_MODULES_PATH . "/{$versioned}")) {
+                    rename(BUILD_MODULES_PATH . "/{$versioned}", BUILD_MODULES_PATH . "/{$unversioned}");
+                    shell()->cd(BUILD_MODULES_PATH)
+                        ->exec(sprintf(
+                            'patchelf --set-soname %s %s',
```
henderkes, line 312: it's a bit dirty, but I don't know a better way of dealing with this, short of patching deep into the config9.m4's of shared built-in extensions

crazywhalecc:
I got build error when building shared extension:

```
bin/spc build "" --build-shared=swoole --debug --build-cli --no-strip
```

```
configure:4185: clang -I/Users/jerry/project/git-project/static-php-cli/buildroot/include -I/Users/jerry/project/git-project/static-php-cli/buildroot/include/php -I/Users/jerry/project/git-project/static-php-cli/buildroot/include/php/main -I/Users/jerry/project/git-project/static-php-cli/buildroot/include/php/TSRM -I/Users/jerry/project/git-project/static-php-cli/buildroot/include/php/Zend -I/Users/jerry/project/git-project/static-php-cli/buildroot/include/php/ext  -L/Users/jerry/project/git-project/static-php-cli/buildroot/lib conftest.c -Wl,-Bstatic -Wl,--start-group -lcares -lbrotlidec -lbrotlienc -lbrotlicommon -lz -lssl -lcrypto -lnghttp2 -lphp -lcurl -Wl,--end-group -Wl,-Bdynamic -lc -lpthread -lnghttp2 -framework CoreFoundation -framework CoreServices -framework SystemConfiguration -lresolv >&5
ld: unknown options: -Bstatic --start-group --end-group -Bdynamic 
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

src/SPC/builder/linux/LinuxBuilder.php
```diff
@@ -293,6 +293,24 @@ protected function buildEmbed(): void
...
+                            escapeshellarg($unversioned),
+                            escapeshellarg($unversioned)
+                        ));
```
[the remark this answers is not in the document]
crazywhalecc: Well, copilot should stop misleading drivers.

henderkes:
> I got build error when building shared extension:
> 
> ```
> bin/spc build "" --build-shared=swoole --debug --build-cli --no-strip
> ```
> 
> ```
> configure:4185: clang -I/Users/jerry/project/git-project/static-php-cli/buildroot/include -I/Users/jerry/project/git-project/static-php-cli/buildroot/include/php -I/Users/jerry/project/git-project/static-php-cli/buildroot/include/php/main -I/Users/jerry/project/git-project/static-php-cli/buildroot/include/php/TSRM -I/Users/jerry/project/git-project/static-php-cli/buildroot/include/php/Zend -I/Users/jerry/project/git-project/static-php-cli/buildroot/include/php/ext  -L/Users/jerry/project/git-project/static-php-cli/buildroot/lib conftest.c -Wl,-Bstatic -Wl,--start-group -lcares -lbrotlidec -lbrotlienc -lbrotlicommon -lz -lssl -lcrypto -lnghttp2 -lphp -lcurl -Wl,--end-group -Wl,-Bdynamic -lc -lpthread -lnghttp2 -framework CoreFoundation -framework CoreServices -framework SystemConfiguration -lresolv >&5
> ld: unknown options: -Bstatic --start-group --end-group -Bdynamic 
> clang: error: linker command failed with exit code 1 (use -v to see invocation)
> ```

ld64 doesn't understand a lot of the linker flags that lld/gnu-ld support. MacOS shared extension support is very much barebones and we're expecting .so files (which curiously, also get generated on macos, when I expected .dylibs). I was able to build shared extensions on MacOS by toying with linker flags a bit, but I only have a shitty hacked together docker environment and hate the MacOS keyboard scheme, so it's not a priority for me.

If you want to properly support shared extensions on mac, there's a lot of work to be done. Start group, end group for one is crucial, because by default, the linker doesn't go over previously visited archives again to find new unresolved symbols. We'd need to pass the libraries in perfect order, so that every library is only specified once, at the very last time that it needs to be referenced.

henderkes:
closes #765

crazywhalecc:
Removing `LIBS=-Wl,-Bstatic ...` seems to work fine on macOS so far. 

```
$ ls -lah buildroot/modules/swoole.so
-rwxr-xr-x  1 jerry  staff    32M Jun 13 14:08 buildroot/modules/swoole.so

$ php -d "extension=buildroot/modules/swoole.so" --ri swoole

swoole

Swoole => enabled

$ otool -L buildroot/modules/swoole.so
buildroot/modules/swoole.so:
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1351.0.0)
        /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1900.180.0)
```

henderkes:
You're welcome to remove it, it was mostly because it made testing easier in combination with --whole-archive, to force full resolution of symbols in those libraries. I think I edged it out with the new logic to find dependent libraries and the --start-group and --end-group-

src/SPC/command/BuildPHPCommand.php
```diff
@@ -158,14 +159,24 @@ public function handle(): int
...
+                $cleanPhpSrc = true;
+            }
+            if ($cleanPhpSrc) {
+                logger()->info('Cleaning previous php build due to mismatching versions...');
+                FileSystem::removeDir(SOURCE_PATH . '/php-src');
+                FileSystem::removeDir(BUILD_MODULES_PATH);
+                $binFiles = glob(BUILD_BIN_PATH . '/php*');
+                $libFiles = glob(BUILD_LIB_PATH . '/libphp*');
+                foreach ([...$binFiles, ...$libFiles] as $file) {
+                    unlink($file);
```
crazywhalecc: Using `spc switch-php-version 8.x` looks better for switching php version. It already supports cleaning `sources/php-src` function with new download. And by the way lines to remove libphp and php binaries should also be added there. 

But if you think that switching PHP versions must change sources in the build command, we should first implementing version locks for each library, rather than PHP alone.

henderkes: I wasn't aware of that command - craft also doesn't run it.

And yes, sorry, I'm kind of using the branch to update changes for the spc-packages repo. Got FrankenPHP to build manually already, but it required patching of the php-config paths. At some point we'll have to put that in spc.

I don't think version locks for libraries is worth it. For example, when rebuilding zlib, 15? extensions would have to be rebuilt as well. It's enough to force a full rebuild on library updates.

crazywhalecc: > I don't think version locks for libraries is worth it. For example, when rebuilding zlib, 15? extensions would have to be rebuilt as well. It's enough to force a full rebuild on library updates.

I have imagined that after extracting the downloaded source, the hash of the downloaded file will be placed in the `sources/xxx` directory. If the hash and the downloaded lock are changed when extracting the file during the build process, delete it and extract it again. This will work for PHP, dependencies, and extensions, but I haven't figured out the git type yet.

Also, we could remove buildroot/lib/libxxx.a at the same time. If you concern about dependency corrupted, we could apply for dependency tree.

> I wasn't aware of that command - craft also doesn't run it.

If you change the craft.yml file, you should restart the build (delete buildroot and source) instead of just deleting php-src.

henderkes: If the library dependencies don't change, there's no reason to.

Background is the following:
```cmd
bin/spp all --type=rpm --command=spc --version=8.X`
```
will build, package and create a repo for that php version. running with 8.Y afterwards doesn't need to rebuild libraries - only php-src.

henderkes: (the 8.X variable is dynamically written into a craft.yml file)

henderkes: > I have imagined that after extracting the downloaded source, the hash of the downloaded file will be placed in the sources/xxx directory. If the hash and the downloaded lock are changed when extracting the file during the build process, delete it and extract it again. This will work for PHP, dependencies, and extensions, but I haven't figured out the git type yet.

> Also, we could remove buildroot/lib/libxxx.a at the same time. If you concern about dependency corrupted, we could apply for dependency tree.

Yes, that's a very good idea. I think we should also rewrite the --ignore-cache-sources download option to check if there's an update and only download then. That way --ignore-cache-dources=* could become the default without causing much traffic.

crazywhalecc: > Yes, that's a very good idea. I think we should also rewrite the --ignore-cache-sources download option to check if there's an update and only download then. That way --ignore-cache-dources=* could become the default without causing much traffic.

You caught it. The reason I didn't implement the ignore-cache-sources as default is that we cannot or **unable** to get the version or file hash stably and constantly. 

Many files can only be downloaded to calculate the hash, and the version can only be obtained after extraction. But if we need to download first, there is no reason to change the default option.

I'll try to implement a lock file hash later, and extract the `./.spc-hash` file after decompression. For git, I'm going to use `git rev-parse HEAD` as the value. For files, just calculate the sha1 value directly.

henderkes: Maybe it would be okay to just store the last used download link and compare against that? Most of our libraries come from github anyway, so you can parse the commit hash.

src/SPC/command/BuildPHPCommand.php
```diff
@@ -158,14 +159,24 @@ public function handle(): int
...
-                if (($ver = $builder->getPHPVersionFromArchive()) !== false) {
-                    $indent_texts['PHP Version'] = $ver;
+                $cleanPhpSrc = $builder->getPHPVersion() !== $builder->getPHPVersionFromArchive();
+            } catch (RuntimeException|WrongUsageException) {
+                $cleanPhpSrc = true;
+            }
+            if ($cleanPhpSrc) {
+                logger()->info('Cleaning previous php build due to mismatching versions...');
+                FileSystem::removeDir(SOURCE_PATH . '/php-src');
+                FileSystem::removeDir(BUILD_MODULES_PATH);
```
henderkes: @crazywhalecc do we still need this after your lock changes?

crazywhalecc: No need.

src/SPC/command/BuildPHPCommand.php
```diff
@@ -319,6 +319,13 @@ private function parseRules(array $shared_extensions = []): int
             $rule |= BUILD_TARGET_EMBED;
             f_putenv('SPC_CMD_VAR_PHP_EMBED_TYPE=' . ($embed === 'static' ? 'static' : 'shared'));
         }
+        if ($this->getOption('build-frankenphp')) {
+            $rule |= BUILD_TARGET_FRANKENPHP;
```
crazywhalecc: Redundant code.

src/SPC/command/BuildPHPCommand.php
```diff
@@ -319,6 +319,13 @@ private function parseRules(array $shared_extensions = []): int
...
+            if (!$this->getOption('enable-zts')) {
+                logger()->warning('FrankenPHP requires ZTS to work with multiple threads, the --enable-zts option will be enabled automatically!');
+                $this->input->setOption('enable-zts', true);
```
crazywhalecc: I actually more prefer throwing wrong usage exception at `checkBeforeBuildPHP`. It's REAL wrong usage.

crazywhalecc:
Considering that the main branch is used by the vast majority of downstream packages, all PRs will be merged into the main branch, including mine, should be approved by at least one of you and me before being merged.

bin/spc-gnu-docker
```diff
@@ -74,6 +74,11 @@ RUN echo "source scl_source enable devtoolset-10" >> /etc/bashrc
 RUN source /etc/bashrc
 RUN yum install -y which
 
+RUN curl -fsSL -o patchelf.tgz https://github.com/NixOS/patchelf/releases/download/0.18.0/patchelf-0.18.0-x86_64.tar.gz && \
```
crazywhalecc: Fixed arch?

crazywhalecc (changes requested):

crazywhalecc:
The rest part looks good to me. I will add some documentation about frankenphp later.

henderkes:
there were a lot of issues introduced by the lock file changes, I think I'm getting behind them though

henderkes:
found the segmentation fault bug in imagick, I think it forced symbols from libphp.a into imagick.so, which then conflicted when a program that already had the symbols from libphp.a tried to load the extension

henderkes:
heh, we have a small problem with xcaddy - go install xcaddy links it against the systems libraries. when we install xcaddy on the host and then try to run bin/spc-alpine-docker, it fails with the misleading error that the xcaddy file doesn't exist

crazywhalecc:
> heh, we have a small problem with xcaddy - go install xcaddy links it against the systems libraries. when we install xcaddy on the host and then try to run bin/spc-alpine-docker, it fails with the misleading error that the xcaddy file doesn't exist

Sounds like the pkgroot and buildroot reuse in different env error. I don't currently have any good way to get spc to recognize the correct pkgroot and buildroot, since we expect everyone to compile in only one environment at one time.

henderkes:
yeah I think it's fair to ignore this for now.

henderkes:
okay: mac failed 3x to rate limiting now. doesn't Downloader::curlExec automatically use the GITHUB_TOKEN?

config/env.ini
```diff
@@ -43,7 +43,7 @@ SPC_SKIP_PHP_VERSION_CHECK="no"
 ; Ignore some check item for bin/spc doctor command, comma separated (e.g. SPC_SKIP_DOCTOR_CHECK_ITEMS="if homebrew has installed")
 SPC_SKIP_DOCTOR_CHECK_ITEMS=""
 ; extra modules that xcaddy will include in the FrankenPHP build
-SPC_CMD_VAR_FRANKENPHP_XCADDY_MODULES="--with github.com/dunglas/frankenphp/caddy --with github.com/dunglas/mercure/caddy --with github.com/dunglas/vulcain/caddy --with github.com/dunglas/caddy-cbrotli"
+SPC_CMD_VAR_FRANKENPHP_XCADDY_MODULES="--with github.com/php/frankenphp/caddy --with github.com/dunglas/mercure/caddy --with github.com/dunglas/vulcain/caddy --with github.com/dunglas/caddy-cbrotli"
```
henderkes: this doesn't work, he hasn't updated the module name yet

docs/en/guide/manual-build.md
```diff
@@ -509,6 +511,8 @@ When `bin/spc doctor` automatically repairs the Windows environment, tools such
 Here is an example of installing the tool:
 
 - Download and install UPX (Linux and Windows only): `bin/spc install-pkg upx`
+- Download and install nasm (Windows only): `bin/spc install-pkg nasm`
+- Download and install go-mod-frankenphp: `bin/spc install-pkg go-mod-frankenphp`
```
henderkes: maybe we should just call this "go-xcaddy"

src/SPC/builder/unix/UnixBuilderBase.php
```diff
@@ -317,8 +317,8 @@ protected function buildFrankenphp(): void
         $nowatcher = $this->getLib('watcher') === null ? ',nowatcher' : '';
         $xcaddyModules = getenv('SPC_CMD_VAR_FRANKENPHP_XCADDY_MODULES');
         // make it possible to build from a different frankenphp directory!
-        if (!str_contains($xcaddyModules, '--with github.com/dunglas/frankenphp')) {
-            $xcaddyModules = '--with github.com/dunglas/frankenphp ' . $xcaddyModules;
+        if (!str_contains($xcaddyModules, '--with github.com/php/frankenphp')) {
+            $xcaddyModules = '--with github.com/php/frankenphp ' . $xcaddyModules;
```
henderkes: again, not possible yet

config/env.ini
henderkes: (and I believe it will move to frankenphp.dev instead, although I'm strongly in favour of github.com/php/frankenphp) @dunglas

crazywhalecc:
Sorry, but I prefer that whoever made the changes should fix them. 🫣 Just now my local has been replaced globally, and the your commit pushed is missing the part in the docs and UnixBuilderBase.

henderkes:
Oh, sorry! On the bright side, frankenphp built with musl now links fully statically again, meaning it can also run on glibc.

henderkes:
> Sorry, but I prefer that whoever made the changes should fix them. 🫣 Just now my local has been replaced globally, and the your commit pushed is missing the part in the docs and UnixBuilderBase.

That's by design, though. We keep php/frankenphp where it's possible, it's only for the xcaddy command where it must be called dunglas/frankenphp.

src/SPC/builder/unix/UnixBuilderBase.php
```diff
@@ -277,4 +294,70 @@ protected function patchPhpScripts(): void
...
+        $arch = arch2gnu(php_uname('m'));
+
+        // define executables for go and xcaddy
+        $xcaddy_exec = PKG_ROOT_PATH . "/go-xcaddy-{$arch}-{$os}/bin/xcaddy";
+
+        $nobrotli = $this->getLib('brotli') === null ? ',nobrotli' : '';
+        $nowatcher = $this->getLib('watcher') === null ? ',nowatcher' : '';
+        $xcaddyModules = getenv('SPC_CMD_VAR_FRANKENPHP_XCADDY_MODULES');
+        // make it possible to build from a different frankenphp directory!
+        if (!str_contains($xcaddyModules, '--with github.com/php/frankenphp')) {
```
crazywhalecc: This seems to be dunglas/frankenphp.

docs/zh/guide/manual-build.md
```diff
@@ -457,6 +459,8 @@ bin/spc dev:sort-config ext
 下面是安装工具的示例：
 
 - 下载安装 UPX（仅限 Linux 和 Windows）: `bin/spc install-pkg upx`
+- 下载安装 nasm（仅限 Windows）: `bin/spc install-pkg nasm`
+- 下载安装 go-mod-frankenphp: `bin/spc install-pkg go-mod-frankenphp`
```
crazywhalecc: Just missing replacement for docs.

crazywhalecc (changes requested):

henderkes:
I'll let a few more static extension tests run before I merge this

henderkes:
everything builds correctly except the opcache jit error. I thought php-src already fixed that O.o

crazywhalecc:
I don't think PHP will be fixed immediately in versions other than the main branch. Also, if necessary, I would use `gh cache -R crazywhalecc/static-php-cli delete --all` to clear the cache to confirm that the latest source is valid.

henderkes:
oh you're right, it was tagged, but for 8.4.9 https://github.com/php/php-src/commit/7a0beb4867259522c865b035d094e89215c38691

Feel free to merge this then.

PR #768, opened 2025-06-13
Add commit tests

crazywhalecc:
## Checklist before merging


docs/deps-craft-yml.md
```diff
@@ -42,6 +42,9 @@ build-options:
   # Set micro SAPI as win32 mode, without this, micro SAPI will be compiled as a console application (only for Windows, default: false)
   enable-micro-win32: false
 
+# Build options for shared extensions (same as `build-shared` command options, all options are optional)
+shared-extensions: [ ]
```
henderkes, line 46: fyi we can't use build-shared as a build option because it will generate:

`--build-shared=amqp --build-shared=apcu --build-shared=bcmath ...`

henderkes (changes requested):

henderkes:
I'm on a short weekend trip, I can leave a proper review tomorrow evening

crazywhalecc:
> I'm on a short weekend trip, I can leave a proper review tomorrow evening

Ah no problem. I'll leave tomorrow, too. Enjoy your weekend!

.github/workflows/commit-tests.yml
```diff
@@ -0,0 +1,137 @@
...
+            BASH_CONTENT=$(echo "$COMMIT_MESSAGE" | sed -nz 's/.*{bash}\(.*\){\/bash}.*/\1/p')
+            echo "Bash content: $BASH_CONTENT"
+            # set the output variable
+            echo "bash<<EOF" >> $GITHUB_OUTPUT
+            echo "$BASH_CONTENT" >> $GITHUB_OUTPUT
+            echo "EOF" >> $GITHUB_OUTPUT
+          fi
+
+          # parse spc_prefix from commit message, e.g. [spc_prefix:bin/spc-gnu-docker], default: bin/spc
+          if [[ "$COMMIT_MESSAGE" =~ \[spc_prefix:([^\]]+)\] ]]; then
```
henderkes: I think this is too difficult to actually use. To test with everything we'd need to define "craft bash spc_prefix=spc-gnu-docker spc_prefix`?

henderkes: don't you think it would be enough to parse for "test" or "fix" in the message and run all tests, unless otherwise defined?

crazywhalecc: To be honest, I also feel that it is a bit complicated, but I can't find a balanced solution.

I also thought about combining ext-matrix-test with test-extensions.php. When the message contains "[matrix:8.4:macos]", test all extensions on the 8.4 build of mac, but still need to define a set of simple and easy-to-use tag rules.

Directly implanting craft is the easiest way to use the ready-made solution without defining the syntax. But if I don't consider the syntax at all, I can't think of a more optimized solution to dynamically trigger the build.

crazywhalecc: The most effective solution might be connect AI, let AI determine what needs to be tested for this commit or this PR 😈

henderkes: Hahaha, surely that would go well!

> "[matrix:8.4:macos]"

Again, to complicated for my liking. May be fine for us, but it's hard to eventually tell people how to properly test their code.

We don't need to be that selective about testing, I think. Just trigger *all* tests when the commit message contains "fix" or "test". The whole matrix with all extensions/libraries currently defined in test-extensions.

crazywhalecc:
Looks good to me now. I also updated the PR template.

.github/workflows/ext-matrix-tests.yml
```diff
@@ -1,16 +1,14 @@
...
-    branches: [ "main" ]
-    paths:
-      - '.github/workflows/ext-matrix-tests.yml'
+  push:
 
 jobs:
   test:
     name: "${{ matrix.extension }} (PHP ${{ matrix.php-version }} on ${{ matrix.operating-system }})"
     runs-on: ${{ matrix.operating-system }}
+    if: contains(github.event.head_commit.message, 'extension test')
```
henderkes: `extension test` or `test extensions` please.

or just `test`

henderkes (approved):

.github/workflows/ext-matrix-tests.yml
henderkes: I really think we should just make it `test`. We're already going from running tests on every commit to just a few. If we accidentally end up testing a few times more, that's no problem.

PR #769, opened 2025-06-14
Add source hash comparator & refactor download lock

crazywhalecc:
## What does this PR do?

- Add download type `local`
- After downloading different version, extraction will be triggered if hash does not match
- Change `dir` type to `git`

## Checklist before merging

### DIFF
--- a/src/SPC/store/Downloader.php
+++ b/src/SPC/store/Downloader.php
@@ -208,7 +208,7 @@ class Downloader
         if ($download_as === SPC_DOWNLOAD_PRE_BUILT) {
             $name = self::getPreBuiltLockName($name);
         }
-        self::lockSource($name, ['source_type' => 'archive', 'filename' => $filename, 'move_path' => $move_path, 'lock_as' => $download_as]);
+        self::lockSource($name, ['source_type' => SPC_SOURCE_ARCHIVE, 'filename' => $filename, 'move_path' => $move_path, 'lock_as' => $download_as]);
     }
 
     /**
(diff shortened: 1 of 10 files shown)


src/SPC/store/Downloader.php
```diff
@@ -371,6 +374,16 @@ public static function downloadPackage(string $name, ?array $pkg = null, bool $f
...
                     break;
+                case 'local':
+                    // Local directory, do nothing, just lock it
+                    logger()->debug("Locking local source {$name}");
+                    self::lockSource($name, [
+                        'source_type' => SPC_SOURCE_LOCAL,
+                        'dirname' => $pkg['dirname'],
+                        'move_path' => $pkg['extract'] ?? null,
+                        'lock_as' => SPC_DOWNLOAD_PACKAGE,
+                    ]);
```
henderkes, line 385: what's the point in this? people adding their own local extensions?

tests/bootstrap.php
```diff
@@ -2,5 +2,7 @@
 
 declare(strict_types=1);
 
+putenv('SPC_IGNORE_BAD_HASH=yes');
```
henderkes: should probably but this somewhere else

henderkes:
looks good generally, just a nitpick and a question

src/SPC/store/Downloader.php
crazywhalecc: Yeah. For someone who want to use spc in vendor mode or private project.

henderkes (approved):

PR #772, opened 2025-06-18
Sapi/frankenphp

henderkes:
## What does this PR do?
adds support for the FrankenPHP sapi (linux, macos, bsd)

## Checklist before merging


src/SPC/builder/traits/UnixGoCheckTrait.php
```diff
@@ -0,0 +1,83 @@
...
+        $gobin = getenv('GOBIN') ?: (getenv('HOME') . '/go/bin');
+        putenv("GOBIN={$gobin}");
+
+        $paths[] = $gobin;
+
+        if ($this->findCommand('xcaddy', $paths) === null) {
+            shell(true)->exec('go install github.com/caddyserver/xcaddy/cmd/xcaddy@latest');
+        }
+
+        return CheckResult::ok();
```
henderkes: the go install locations are what I'm least satisfied with. I don't like leaving a `go` folder under $HOME, but also can't think of a better way

src/SPC/builder/unix/UnixBuilderBase.php
```diff
@@ -277,4 +277,36 @@ protected function patchPhpScripts(): void
...
+            'LD_LIBRARY_PATH' => BUILD_LIB_PATH,
+        ];
+        shell()->cd(BUILD_BIN_PATH)
+            ->setEnv($env)
+            ->exec(
+                'xcaddy build ' .
+                '--output frankenphp ' .
+                '--with github.com/dunglas/frankenphp/caddy ' .
+                getenv('SPC_CMD_VAR_FRANKENPHP_XCADDY_MODULES')
+            );
```
henderkes: pretty simple really, just needs xcaddy

henderkes:
Hahaha, I found a frankenphp bug. The php-server directive unconditionally enables brotli encoding.

henderkes:
```console
bin/spc-gnu-docker build "bcmath" --build-frankenphp --enable-zts
[05:05:58] [I]    Build complete, used 76.212 s !
[05:05:58] [I]
[05:05:58] [I] License path (host system): /home/m/static-php-cli/buildroot/license/
[m@M-TH static-php-cli]$ buildroot/bin/frankenphp php-server
Error: module not registered: http.encoders.br
[m@M-TH static-php-cli]$ buildroot/bin/frankenphp run
2025/06/18 05:06:23.724 INFO    maxprocs: Leaving GOMAXPROCS=24: CPU quota undefined
2025/06/18 05:06:23.724 INFO    GOMEMLIMIT is updated   {"package": "github.com/KimMachineGun/automemlimit/memlimit", "GOMEMLIMIT": 14982012518, "previous": 9223372036854775807}
2025/06/18 05:06:23.731 INFO    admin   admin endpoint started  {"address": "localhost:2019", "enforce_origin": false, "origins": ["//localhost:2019", "//[::1]:2019", "//127.0.0.1:2019"]}
2025/06/18 05:06:23.732 INFO    serving initial configuration
```

henderkes:
@dunglas this might be interesting to you. Super simple way to build frankenphp.

Static build:
```
bin/spc-gnu-docker build "brotli" --with-libs="watcher" --build-frankenphp --enable-zts
```

Shared build:
```
bin/spc-gnu-docker build "brotli" --with-libs="watcher" --build-shared=shared --build-frankenphp --enable-zts
```

crazywhalecc:
Regardless of the weird upstream and downstream dependency issues, I think there are a few places that need to be changed.

1. Using go and caddy as doctor's pills is a bit overdiagnosed. People who use spc in general may not build it. We should install go and xcaddy as pkg.
2. Golang's dependencies should also be downloaded in advance during the download phase, but we currently don't have a corresponding pull mechanism, so we may need to reconsider.
3. It seems that embed SAPI is not built before building FrankenPHP.
4. SPCConfigUtil should be used because macOS and other platforms may lack frameworks.

henderkes:
1.) definitely possible, but it'll lead to them being downloaded multiple times on the same OS. I placed them in /usr/local/go just like we place the musl toolchain in /usr/local/musl
2.) I don't think that's possible
3.) It is, BUILD_FRANKENPHP = BUILD_EMBED | 16;
4.) I'm not sure what you mean. Could you do the change you're thinking of?

henderkes:
For 1.) I'm still not sure what the best course of action is. We could make `--build-frankenphp` add 'go,xcaddy' to `--with-libs`, but then we'd have to cross check if they're already installed on the OS. Every time we clear buildroot/, they disappear again and need to be re-downloaded, even though there may not be any reason to. We'd also have to, again, conditionally, override the GOROOT if it's set.

crazywhalecc:
I'll show you some cases for Q1 and Q4 later. And Q2 is not impossible: `go mod vendor` looks good to me.

config/env.ini
```diff
@@ -42,6 +42,9 @@ SPC_CONCURRENCY=${CPU_COUNT}
 SPC_SKIP_PHP_VERSION_CHECK="no"
 ; Ignore some check item for bin/spc doctor command, comma separated (e.g. SPC_SKIP_DOCTOR_CHECK_ITEMS="if homebrew has installed")
 SPC_SKIP_DOCTOR_CHECK_ITEMS=""
+; extra modules that xcaddy will include in the FrankenPHP build
+SPC_CMD_VAR_FRANKENPHP_XCADDY_MODULES="--with github.com/dunglas/frankenphp/caddy --with github.com/dunglas/mercure/caddy --with github.com/dunglas/vulcain/caddy --with github.com/dunglas/caddy-cbrotli"
```
crazywhalecc, line 46: Are these modules required all the time? If so, it would be better to add them as dependencies directly.

crazywhalecc:
Oh, we merge into the docs branch?

config/env.ini
henderkes: They are used in the official frankenphp build, so I set it as an overridable option.

I personally build without vulcain but with caddy cache and supervisor.

henderkes:
> I'm still looking for ways to separate downloading and building, but for now it looks almost impossible to use xcaddy build instead of go build.

Yeah, like I said earlier, I don't think there's a good way. It will have to do some downloading every time we build (it also updates modules every time), the only way around that would be incredibly tedious by not going with xcaddy at all.

henderkes:
> Oh, we merge into the docs branch?

yes, it needs some changes I've made there for the packaging to even work properly

config/env.ini
henderkes: (also, again, can't download them earlier as it's part of xcaddy's process)

crazywhalecc:
> yes, it needs some changes I've made there for the packaging to even work properly

But I haven't fully reviewed the docs branch yet, and it seems there are some unfinished parts.

henderkes:
> and it seems there are some unfinished parts.

I don't think so. These changes here more or less complete it.

This branch was based off of the docs branch in the first place, it has all of its changes in here.

PR #775, opened 2025-06-18
Sapi/frankenphp prerequisites

crazywhalecc:
## What does this PR do?

Some framework updates and refactoring frankenphp deps download method.

## Checklist before merging

  - [X] `composer analyse`
  - [X] `composer test`


src/SPC/builder/BuilderBase.php
```diff
@@ -510,6 +513,26 @@ public function emitPatchPoint(string $point_name): void
         }
     }
 
+    public function checkBeforeBuildPHP(int $rule): void
+    {
+        if (($rule & BUILD_TARGET_FRANKENPHP) === BUILD_TARGET_FRANKENPHP) {
+            // frankenphp only support linux and macOS
+            if (!in_array(PHP_OS_FAMILY, ['Linux', 'Darwin'])) {
```
henderkes, line 520: why not BSD?
it builds fine on FreeBSD and OpenBSD, dunglas recently merged a fix

src/SPC/builder/BuilderBase.php
```diff
@@ -510,6 +513,26 @@ public function emitPatchPoint(string $point_name): void
...
+            // frankenphp needs package go-mod-frankenphp installed
+            $pkg_dir = PKG_ROOT_PATH . '/go-mod-frankenphp-' . arch2gnu(php_uname('m')) . '-' . osfamily2shortname();
+            if (!file_exists("{$pkg_dir}/bin/go") || !file_exists("{$pkg_dir}/bin/xcaddy")) {
+                global $argv;
+                throw new WrongUsageException("FrankenPHP SAPI requires go-mod-frankenphp package, please install it first: {$argv[0]} install-pkg go-mod-frankenphp");
+            }
+            // frankenphp needs libxml2 libs
+            if (!$this->getLib('libxml2')) {
+                throw new WrongUsageException('FrankenPHP SAPI requires libxml2 library, please include `xml` extension in your build.');
+            }
```
henderkes: is this a requirement on macos? FrankenPHP itself doesn't need libxml2, it needs brotli and watcher for all the default functionality, but that's what -nowatcher and -nobrotli tags are for.

src/SPC/builder/unix/UnixBuilderBase.php
```diff
@@ -219,6 +219,19 @@ protected function sanityCheck(int $build_target): void
...
         }
+
+        // sanity check for frankenphp
+        if (($build_target & BUILD_TARGET_FRANKENPHP) === BUILD_TARGET_FRANKENPHP) {
+            logger()->info('running frankenphp sanity check');
+            $frankenphp = BUILD_BIN_PATH . '/frankenphp';
+            if (!file_exists($frankenphp)) {
+                throw new RuntimeException('FrankenPHP binary not found: ' . $frankenphp);
+            }
+            [$ret, $output] = shell()->execWithResult($frankenphp . ' -v');
```
henderkes: the idiomatic way of getting the version is `./frankenphp version`

It might also need LD_LIBRARY_PATH to be set to BUILD_LIB_DIR when embed=shared.

src/SPC/builder/unix/UnixBuilderBase.php
```diff
@@ -313,21 +329,27 @@ protected function buildFrankenphp(): void
...
         $env = [
+            'PATH' => PKG_ROOT_PATH . "/go-mod-frankenphp-{$arch}-{$os}/bin:" . getenv('PATH'),
+            'GOROOT' => PKG_ROOT_PATH . "/go-mod-frankenphp-{$arch}-{$os}",
+            'GOBIN' => PKG_ROOT_PATH . "/go-mod-frankenphp-{$arch}-{$os}/bin",
+            'GOPATH' => PKG_ROOT_PATH . '/go',
             'CGO_ENABLED' => '1',
-            'CGO_CFLAGS' => '$(php-config --includes) -I$(php-config --include-dir)/..',
-            'CGO_LDFLAGS' => '$(php-config --ldflags) -L' . BUILD_LIB_PATH . " $(php-config --libs) {$brotliLibs} {$watcherLibs} -lphp {$lrt}",
+            'CGO_CFLAGS' => $config['cflags'],
+            'CGO_LDFLAGS' => "{$config['ldflags']} {$config['libs']} {$lrt}",
```
henderkes, line 343: 👍

src/SPC/command/BuildPHPCommand.php
```diff
@@ -316,7 +319,7 @@ private function parseRules(array $shared_extensions = []): int
             $rule |= BUILD_TARGET_EMBED;
             f_putenv('SPC_CMD_VAR_PHP_EMBED_TYPE=' . ($embed === 'static' ? 'static' : 'shared'));
         }
-        $rule |= ($this->getOption('build-frankenphp') ? BUILD_TARGET_FRANKENPHP : BUILD_TARGET_NONE);
+        $rule |= ($this->getOption('build-frankenphp') ? (BUILD_TARGET_FRANKENPHP | BUILD_TARGET_EMBED) : BUILD_TARGET_NONE);
```
henderkes, line 322: I'd rather keep BUILD_TARGET_FRANKENPHP defined as BUILD_TARGET_EMBED | 16, because that way we cannot possibly forget in a different location in the future

src/SPC/store/Downloader.php
```diff
@@ -385,10 +386,13 @@ public static function downloadPackage(string $name, ?array $pkg = null, bool $f
                     ]);
                     break;
                 case 'custom':          // Custom download method, like API-based download or other
-                    $classes = FileSystem::getClassesPsr4(ROOT_DIR . '/src/SPC/store/source', 'SPC\store\source');
+                    $classes = FileSystem::getClassesPsr4(ROOT_DIR . '/src/SPC/store/pkg', 'SPC\store\pkg');
```
henderkes, line 389: 👍

src/SPC/store/pkg/GoModFrankenphp.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        $arch = match (explode('-', $name)[3]) {
+            'x86_64' => 'amd64',
+            'aarch64' => 'arm64',
+            default => throw new \InvalidArgumentException('Unsupported architecture: ' . $name),
+        };
+        $os = match (explode('-', $name)[4]) {
+            'linux' => 'linux',
+            'macos' => 'darwin',
+            default => throw new \InvalidArgumentException('Unsupported OS: ' . $name),
```
henderkes, line 32: again, go supports BSD

src/SPC/builder/BuilderBase.php
crazywhalecc: I have temporarily stopped maintaining BSDBuilder. I will add it back when I need to improve FreeBSD builds and CI.

src/SPC/builder/BuilderBase.php
crazywhalecc: Yes. When I build only one extension without dependencies, it fails to build, saying it can't find libxml-2.0. But this seems to be a problem with FrankenPHP itself. It will force add the system's homebrew and /usr/local directories as LDFLAGS, which will make the spc build unstable. I don't know much about FrankenPHP's build system, but it seems to hardcode the LDFLAGS in the go source.

EDIT: libxml2 missing seems not related to hardcoded LDFLAGS things. These are two problems.

src/SPC/command/BuildPHPCommand.php
crazywhalecc: Because frankenphp itself depends on embed, I think there is nothing wrong with declaring the dependency here. Also, for `($rule & FRANKENPHP) === FRANKENPHP)` that we often use, defining `FRANKENPHP` as `EMBED | 16` is unclear and will cause errors for this line.

src/SPC/builder/BuilderBase.php
henderkes: That seems to be macos specific then? I'm able to build without libxml

henderkes: @dunglas https://github.com/php/frankenphp/blob/main/frankenphp.go#L17

why are libxml2 and the brew paths required on MacOS? This may introduce problems with potentially linking to dynamic libraries or oldrr versions, depending on the order the linker iterates through the library paths

henderkes: this may also introduce issues on linux when a user has the php-devel package installed

is there a way to make Go ignore these linker comments?

PR #779, opened 2025-06-19
remove old extension file before building new php version

henderkes:
## What does this PR do?

I would like to integrate this into the LockFile somehow, but I'm not sure how.

please feel free to move it to the right location


crazywhalecc:
Currently LockFile is only managing source and downloads, it's not managing `buildroot`.

How about deleting the module directory here only if it is detected that this build contains a shared extension?

henderkes:
if it doesn't contain a shared extension, I don't think it's even created in the first place, so that would only add unnecessary code

PR #781, opened 2025-06-19
fix bug with only --build-embed

henderkes:
passing --build-embed sets the result to NULL, not true (and not false, like for options that are not passed)

## What does this PR do?

fixes a bug where 
`spc build "bcmath" --build-embed`

returned `Please add at least one build SAPI!`


henderkes:
looks like that one slipped past us because we don't test with only --build-embed

PR #784, opened 2025-06-19
Revert build-embed option as boolean

crazywhalecc:
## What does this PR do?

Revert build-embed option as boolean. Related: https://github.com/php/frankenphp/pull/1651#issuecomment-2988744303

This should not be a bug, but this would be misleading. Although docopt requires options to support space reading, it will cause some confusion according to our habits, especially in other SAPIs `--build-XXX` does not accept values.

But I think it is acceptable to use another option to override the default environment variable (such as `--with-embed-type=shared`, if you still feel it is necessary to override it with the command.

Otherwise, we consider it a bad practice to encourage others to put all parameters after the extensions.

## Checklist before merging

  - [ ] `composer analyse`
  - [ ] `composer test`


henderkes:
A --with-embed-type=static|shared option would be a good compromise. I'll go back to the drawing board to figure out a better alternative.

PR #792, opened 2025-06-20
Skip re-extract ext-imap source

### LINKED ISSUE #789: [CI] Extension test failed: new LockFile seems conflict with ext-imap source strategy
https://github.com/static-php/static-php-cli-hosted/actions/runs/15770358785/job/44454068593#step:8:8593

The framework wrongly removed and replaced the source from php-src/ext/imap due to it doesn't contain `./.spc-hash` file.

crazywhalecc:
## What does this PR do?

Fix #789 .

## Checklist before merging

  - [X] `composer analyse`
  - [X] `composer test`

### DIFF
--- a/src/SPC/store/SourceManager.php
+++ b/src/SPC/store/SourceManager.php
@@ -66,7 +66,7 @@ class SourceManager
             $check = LockFile::getExtractPath($lock_name, SOURCE_PATH . '/' . $source);
             // $check = $lock[$lock_name]['move_path'] === null ? (SOURCE_PATH . '/' . $source) : (SOURCE_PATH . '/' . $lock[$lock_name]['move_path']);
             if (!is_dir($check)) {
-                logger()->debug('Extracting source [' . $source . '] to ' . $check . ' ...');
+                logger()->debug("Extracting source [{$source}] to {$check} ...");
                 $filename = LockFile::getLockFullPath($lock_content);
                 FileSystem::extractSource($source, $lock_content['source_type'], $filename, $check);
                 LockFile::putLockSourceHash($lock_content, $check);
@@ -81,7 +81,14 @@ class SourceManager
 
             // when source already extracted, detect if the extracted source hash is the same as the lock file one
             if (file_exists("{$check}/.spc-hash") && FileSystem::readFile("{$check}/.spc-hash") === $hash) {
-                logger()->debug('Source [' . $source . '] already extracted in ' . $check . ', skip !');
+                logger()->debug("Source [{$source}] already extracted in {$check}, skip !");
+                continue;
+            }
+
+            // ext imap was included in php < 8.4 which we should not extract,
+            // but since it's not simple to compare php version, for now we just skip it
+            if ($source === 'ext-imap') {
+                logger()->debug("Source [ext-imap] already extracted in {$check}, skip !");
                 continue;
             }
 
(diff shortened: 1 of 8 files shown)


src/SPC/store/SourceManager.php
```diff
@@ -81,7 +81,14 @@ public static function initSource(?array $sources = null, ?array $libs = null, ?
 
             // when source already extracted, detect if the extracted source hash is the same as the lock file one
             if (file_exists("{$check}/.spc-hash") && FileSystem::readFile("{$check}/.spc-hash") === $hash) {
-                logger()->debug('Source [' . $source . '] already extracted in ' . $check . ', skip !');
+                logger()->debug("Source [{$source}] already extracted in {$check}, skip !");
+                continue;
+            }
+
+            // ext imap was embeded in php <= 8.3, if php version < 8.4, we should not extract
```
henderkes: embedded

henderkes: actually, better: "included"

henderkes:
imap still seems to fail, we should probably fix it a last time, but move it into unsupported/deprecated territory.

I remember the huge pain in the ass it was to get it working in the first place, no other library was that bad

crazywhalecc:
That's really weird, I'll have time to look into it next week. 🫠

henderkes:
I'll leave the fun of figuring out the swoole segmentation fault to you. Imap now works again.

crazywhalecc:
Seems swoole is using glibc-only feature, but it will not working with extra CFLAGS `-U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=0`.

crazywhalecc:
I think we should generate environment info report after building anything. I also encounter this problem when trying to use swoole and spc branches from May 31, although its CI was green 2 months ago.

crazywhalecc:
Okay, I found the key point. 2025-05-30 released Alpine Linux 3.22, and before it we were using `alpine:edge -> 3.21.x`. If we use specified version of Alpine like `3.21`, the build is good.

The only difference I found between 3.21 and 3.22 is the gcc version (14.2: `--disable-libsanitizer`, 14.3: `--enable-libsanitizer`).

And if I add `-fsanitize=address` to CFLAGS, it build successfully.

henderkes:
Could likely work around it by using -fno-sanitize=address (or =undefined or =thread) like we do with zig, which automatically enables them. I don't run into the issue when doing a static musl or dynamic musl build with zig.

henderkes:
Oh, on a positive note, all the bugs I've reported are tagged for 15.0 release, so we'll be able to skip the messy zig-cc workaround then. Only the performance one hasn't seen any response.

henderkes:
I'd rather address the issue since we don't know how long it will take to push 3.0. Otherwise people running the latest version will run into the issue.

crazywhalecc:
I've made alpine back to 3.21 and created an issue to swoole repo as workaround. From the most direct perspective, the update of alpine broke our build, so it is a better choice to temporarily return to the previous version.

I believe the best solution is always to fix the upstream repository, but I'm not sure which approach the swoole maintainers are willing to accept. For us, what we can do now is build it successfully.

henderkes:
Oh you're right, is this is a general issue then swoole should set the required compilation parameters in their config.m4.

PR #794, opened 2025-06-20
revert perl-IPC-Cmd

henderkes:
## What does this PR do?

add back perl-IPC-Cmd for spc-gnu-docker (not included in perl package on aarch64, apparently)


henderkes:
don't ask, I had to docker system prune -af to run into this...

PR #800, opened 2025-06-21
remove disabling of asm arithmetic

henderkes:
## What does this PR do?

why did we need this? asm goto's should be perfectly fine on linux

was this specifically for glibc 2.17?


crazywhalecc (approved):
I can't quite remember what this was for (never forget to annotate a specific patch), I'll test it locally later when I have time. But this pr is nothing wrong.

henderkes:
I haven't tested it so I don't want to merge it yet. What I found on the internet was all about it being incompatible with certain extensions like memcached on MacOS, not Linux.

henderkes:
all tests with all extensions pass both static and shared, so I think we can merge this

PR #802, opened 2025-06-22
Check lib-suggest before packing libs

crazywhalecc:
## What does this PR do?

Check lib-suggest before packing libs. This will prevent packing alternative libs when someone uses different dependency tree.

## Checklist before merging

  - [X] `composer analyse`
  - [X] `composer test`

### DIFF
--- a/src/SPC/command/dev/PackLibCommand.php
+++ b/src/SPC/command/dev/PackLibCommand.php
@@ -42,6 +42,15 @@ class PackLibCommand extends BuildCommand
 
             $builder->proveLibs($libraries);
             $builder->validateLibsAndExts();
+
+            // before pack, check if the dependency tree contains lib-suggests
+            foreach ($libraries as $lib) {
+                if (Config::getLib($lib, 'lib-suggests', []) !== []) {
+                    logger()->critical("The library {$lib} has lib-suggests, packing [{$lib_name}] is not safe, abort !");
+                    return static::FAILURE;
+                }
+            }
+
             foreach ($builder->getLibs() as $lib) {
                 if ($lib->getName() !== $lib_name) {
                     // other dependencies: install or build, both ok


src/SPC/command/dev/PackLibCommand.php
```diff
@@ -42,6 +42,15 @@ public function handle(): int
...
             $builder->proveLibs($libraries);
             $builder->validateLibsAndExts();
+
+            // before pack, check if the dependency tree contains lib-suggests
+            foreach ($libraries as $lib) {
+                if (Config::getLib($lib, 'lib-suggests', []) !== []) {
+                    logger()->critical("The library {$lib} has lib-suggests, packing [{$lib_name}] is not safe, abort !");
+                    return static::FAILURE;
+                }
+            }
```
henderkes, line 52: This is good, but not enough. I'm pretty sure that we have libraries that have features depending on other libraries that aren't advertised in lib-suggests.

crazywhalecc: Do you have any good way to grab those libraries?

henderkes: Manual labour :P

crazywhalecc: so be it 💯

henderkes: Where/how was this resolved? Did you go through all the lib to check if they have suggests correctly specified in lib.json?

crazywhalecc: Yet I think that's another issue or PR. For clearer this PR will only add a simple blocker for pack-lib workflows.

henderkes: Okay, we shouldn't forget about it though as it'll inevitably cause bugs if a suggests section isn't present.

henderkes (approved):

PR #804, opened 2025-06-22
Allow deleting pre-built or source only for del-download

crazywhalecc:
## What does this PR do?

Allow deleting pre-built or source only for del-download:

- `bin/spc del-download --pre-built-only XXX`
- `bin/spc del-download --source-only XXX`

Sometimes we need to test either source or pre-built at the same time. It's just a little function to help us simplify the download procedure. (sometimes download from another side of the earth is quite slow)

## Checklist before merging

  - [ ] `composer analyse`
  - [ ] `composer test`


henderkes (approved):
I think it's a bit overkill and I probably won't ever use it given that I never use prebuilt sources, but why not.

PR #810, opened 2025-06-28
refactor: replace SPC_LIBC with SPC_TARGET and update related logic

### LINKED ISSUE #809: Proposal: SPC_LIBC change to more meaningful vars
@henderkes 

I suddenly had a terrible idea (for now): for Linux builds, how about splitting SPC_LIBC into `SPC_TOOLCHAIN` and `SPC_TARGET`?

- SPC_TOOLCHAIN ​​uses `zig` by default after zig is implemented, and others include `alpine-native` and `musl`. Create a tool class: `SPCToolchain`. For windows and macOS: `msbuild`, `clang`.
- SPC_TARGET includes `musl-static`, `musl`, and `glibc`. We can choose according to the situation in the next major version. Tool class: `SPCTarget`. For windows and macOS: `darwin-{some-minimal-version-num}`, `windows-7|8|10|11`

So we have a table:

| SPC_TOOLCHAIN ​​| musl-static | musl | glibc |
| --- | --- | --- | --- |
| zig | yes | yes | yes |
| musl | yes | no | no |
| musl-native | yes | yes | no |
| gnu-native | no | no | yes (limited portability) |

> You may not want `alpine-native` this name, but it does exist - based on alpine-docker or Alpine Linux and can only be used in these env.
> And for musl (formerly musl-toolchain we built), it should be the default solution before our breaking changes.

The advantage is that we can avoid judging the `CC` variable on this basis, and we can expand the `SPC_TOOLCHAIN` and `libc` targets on this basis (assuming it is possible in the future). We can also add different names for Mac and Win to avoid judging whether the current system is Linux at first in general-use code.

And if we finished this, we could remove or add a toolchain support more easily, just check all the toolchain constant reference.

The disadvantage is that it does not fundamentally solve the problem of different tool chains, introduces additional variables, and causes destructive updates.

If we continue to add more duplicate CC checks on top of this, it will make it very difficult to temporarily change these values. But I don't think it is necessary, except that few of us will frequently debug SPC internals and different builds.

crazywhalecc:
## What does this PR do?

replace SPC_LIBC with SPC_TARGET and update related logic. This fixes #809 

Looks like after doing this, we can build musl libc based static-php using `bin/spc-alpine-docker` and `SPC_TARGET=musl`.

## Checklist before merging

  - [X] `composer analyse`
  - [X] `composer test`
  - [X] If adding new or fixing bugs, add commit message containing `extension test` or `test extensions` to trigger full test suite.


src/SPC/builder/extension/imagick.php
```diff
@@ -26,15 +27,15 @@ public function patchBeforeMake(): bool
 
     public function getUnixConfigureArg(bool $shared = false): string
     {
-        $disable_omp = !(getenv('SPC_LIBC') === 'glibc' && str_contains(getenv('CC'), 'devtoolset-10')) ? '' : ' ac_cv_func_omp_pause_resource_all=no';
+        $disable_omp = SPCTarget::isTarget(SPCTarget::GLIBC) ? ' ac_cv_func_omp_pause_resource_all=no' : '';
```
henderkes: removing this entirely in the zig branch, see reasoning there

henderkes: (we always disable omp)

crazywhalecc:
> I don't like this. It doesn't help us move away from the many different conditionals.

In fact, given the current situation, I don't want to get rid of the conditional statement. This just makes the judgment logic clearer. Conditional statements are difficult to avoid, so why not do it as a complete target set?

As for whether to continue to use SPC_LIBC, I think it can continue to be used, but it should not use SPC_LIBC=musl but musl-static. This is why the shared build of musl can be achieved in this PR.

src/SPC/util/GlobalEnvManager.php
```diff
@@ -80,6 +62,30 @@ public static function init(): void
...
+
+        // deprecated: convert SPC_LIBC to SPC_TARGET
+        if (getenv('SPC_LIBC') !== false) {
+            logger()->warning('SPC_LIBC is deprecated, please use SPC_TARGET instead.');
+            $target = match (getenv('SPC_LIBC')) {
+                'musl' => SPCTarget::MUSL_STATIC,
+                default => SPCTarget::GLIBC,
+            };
+            self::putenv("SPC_TARGET={$target}");
+            self::putenv('SPC_LIBC');
```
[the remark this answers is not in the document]
crazywhalecc: Actually unset is right, dude.

henderkes:
> As for whether to continue to use SPC_LIBC, I think it can continue to be used, but it should not use SPC_LIBC=musl but musl-static. This is why the shared build of musl can be achieved in this PR.

Yes, please. But keep SPC_TARGET reserved for actually specifying the target (which is only possible with zig).

crazywhalecc:
> Yes, please. But keep SPC_TARGET reserved for actually specifying the target (which is only possible with zig).

Then our understanding of SPC_TARGET may be different. The target you understand is more like a toolchain options, while the current SPC_TARGET is like the result of the build. The toolchain options should be specified in the variables or configurations at the next level of SPC_TARGET.

henderkes:
No, I'm thinking of SPC_TARGET as an actual target triple/quadruple. `x86_64-linux-gnu.2.17` for example. Or `aarch64-linux-musl -dynamic`.

It's what I want to reserve for Zig compilation, because when Zig is implemented, I want to deprecate (meaning, people have to explicitly configure their environment, rather than rely on our prebuilt toolchains or implicit behaviour) all other builds and only support Zig. It can do everything we need and we only need to support *one* platform, that behaves identical on alpine as it does on ubuntu or rhel, because it's statically compiled and completely ignores the system.

So I want SPC_LIBC's behaviour to stay the way it is now, for compatibility reasons, until 3.0.

People who want to switch to Zig now can use SPC_TARGET - which switches behaviour over to ignore SPC_LIBC and others.
With 3.0 we can then completely remove SPC_LIBC and the musl toolchain. We don't need them anymore. We don't need to check for (almost) any conditions, other than odd cases like openssl's internal target with the -clang postfix.

This gives everyone a clear "upgrade" path with ample time, but will in the end cut our maintenance effort in half. We won't even need docker anymore.

crazywhalecc:
> So I want SPC_LIBC's behaviour to stay the way it is now, for compatibility reasons, until 3.0.

That's also why I created this PR. The SPC_LIBC is not clear and SPC_TARGET can clearly describe what we are building. At the same time, this SPC_TARGET is not only for Linux, but also for other systems (if new target is needed in the future).

> I'm thinking of SPC_TARGET as an actual target triple/quadruple.

What you are talking about does not conflict with this PR. In 3.0, we can expand the parsing level of SPC_TARGET, which is not difficult based on this. But in the current state, we do not need to be too detailed for the time being, because we also don't support any way to specify musl version.

henderkes:
>What you are talking about does not conflict with this PR. In 3.0, we can expand the parsing level of SPC_TARGET, which is not difficult based on this. But in the current state, we do not need to be too detailed for the time being, because we also don't support any way to specify musl version.

Which is great in theory, but wouldn't that mean we need to keep supporting native gcc/musl toolchain/native gcc on glibc? Otherwise we change ABI if we switch over to Zig and don't have a clear upgrade path that doesn't mean switching exactly when we release 3.0.

crazywhalecc:
Since the existing toolchains and environments have been proven to be usable and partially stable and used by a lot of projects, it is not difficult to maintain them for us now. Secondly, the addition of the Zig toolchain will not make spc 3.0, because we didn't remove any ABI/API at present.

So this PR is more like a tidying up of the past messy toolchain judgments + supporting the future Zig or other toolchain more elegant with current toolchain. 

In particular, musl-toolchain will not be downloaded when SPC_TOOLCHAIN ​​is not detected as musl. The Zig toolchain can also be doctor-fixed by simply replacing a string.

henderkes:
> Since the existing toolchains and environments have been proven to be usable and partially stable and used by a lot of projects, it is not difficult to maintain them for us now.

I disagree because there are a *lot* of patches for individual toolchains. Centos 7 gcc needs a Docker container *and* individual patches to like static -lstdc++ because the shared library doesn't contain all symbols we use. GCC under Ubuntu and under RHEL behave differently with different search paths. We need the musl toolchain that we compile ourselves, which is still on gcc 13. We have separate code branches in so many places.

And *there's no reason to keep them*. Zig-cc can do everything that the other toolchains can.

crazywhalecc:
Well, I think we don't have any disagreement on removing redundant toolchains, but I think deleting them is a 3.0 thing. I just said that the "current" PRs do not need to delete any toolchains and judgments, it is not difficult to maintain for now, not to say "it is not difficult to maintain them all the time".

So let's get back to the topic, do you have any reviews about what this PR does?

henderkes:
>  I just said that the "current" PRs do not need to delete any toolchains and judgments, it is not difficult to maintain for now, not to say "it is not difficult to maintain them all the time".

Ah ok, I understand.

I think this is a good move, but I would like to keep it called SPC_LIBC for backwards compatibility. 'musl' needs to stay musl-static for compatibility reasons and we will add a musl-dynamic later.

I want SPC_TARGET to be reserved for zig builds.

crazywhalecc:
> I want SPC_TARGET to be reserved for zig builds.

What do you think if I keep continue to use SPC_LIBC as the mapping of SPC_TARGET:

- SPC_LIBC=musl -> SPC_TARGET=musl-static
- SPC_LIBC=glibc -> SPC_TARGET=glibc
- SPC_LIBC=musl-shared -> SPC_TARGET=musl

And for zig builds, I want to treat SPC_TARGET as `glibc@zig-2.17` or `glibc@2.17` for backward compatibility and forward compatibility. I don't know which symbol (or char?) is appropriate, but the specific selection symbol should always be at the end of the real "target" string. For practical purposes, zig is zig, and `@{something}` means we are telling SPC to tell the toolchain predicate a specific condition.

henderkes:
`SPC_LIBC=musl-dynamic` -> `SPC_TARGET=native-native-musl -dynamic`
`SPC_LIBC=musl` -> `SPC_TARGET=native-native-musl`
`SPC_LIBC=glibc` -> `SPC_TARGET=native-native-gnu`
`SPC_LIBC=glibc@2.17` -> not supported, use `SPC_TARGET=native-native-gnu.2.17` instead!

but keep the usage of SPC_LIBC with gcc! Only when SPC_TARGET is actually used, it should *always* use zig.

crazywhalecc:
`SPC_TARGET` is also preserved for Mach-O cross compilation, it's not linux-only or zig-exclusive. And btw zig is not native. I still insist that SPC_TARGET is a multiple choice question + labeling suffix within a range, and should not be exclusive to a certain compiler or toolchain.

henderkes:
Zig supports "native" for architecture and OS to choose what it's running on.
When you're on aarch64-linux, `native-native-gnu` and `aarch64-linux-gnu` are equivalent.

henderkes:
>SPC_TARGET is a multiple choice question + labeling suffix within a range, and should not be exclusive to a certain compiler or toolchain.

Then what's the upgrade path to future 3.0 where we drop SPC_LIBC and other toolchains? I want SPC_TARGET to be the option that uses zig now and stays stable with 3.0, so people can switch to zig now.

crazywhalecc:
> Zig supports "native" for architecture and OS to choose what it's running on. When you're on aarch64-linux, `native-native-gnu` and `aarch64-linux-gnu` are equivalent.

Oh, when I say native, I mean whether the compiler or toolchain is native, not the architecture. So the name of the toolchain in my PR is also called gcc-native when using native (or system) gcc. I have also defined a default clang-native toolchain using clang for MacOS as well, which initializes default environment variables in the ClangNativeToolchain class.

After using zig, you can directly change linux's `gcc-native` to zig without modifying other parts.

As for cross-compilation on Linux and Mac, I also think that the current toolchain does not support it, so using the suffix passed to the toolchain + mapping SPC_LIBC is still a better choice.

Let me explain our route again:

- Implement this PR (SPC_TARGET is equivalent to an enhancement of SPC_LIBC)
- Implement zig and use zig as the default build
- Implement different versions of zig (SPC_TARGET version suffix)
- Deprecate other toolchains for Linux in 3.0

henderkes:
> Let me explain our route again:
> 
> * Implement this PR (SPC_TARGET is equivalent to an enhancement of SPC_LIBC)
> * Implement zig and use zig as the default build
> * Implement different versions of zig (docking version suffix)
> * Abandon other toolchains

That route has a problem though. we cannot use zig as the default build without breaking BC.

That's why I want:

* Implement this PR based only on SPC_LIBC, where SPC_LIBC=musl and SPC_LIBC=glibc keep their original functionality (musl toolchain on ubuntu, or native gcc on alpine and native gcc on ubuntu)
* Implement zig - zig is *only* used when SPC_TARGET variable is set (ignores SPC_LIBC)
* Warning when SPC_TARGET is not used
* 3.0: drop SPC_LIBC and drop musl-toolchains and shenanigans in our code related to 2.17 gcc.
* Now people can still use musl-toolchain or gcc as much as they like with empty SPC_TARGET, but they need to configure CC, CXX, LD and PATH manually.

crazywhalecc:
> That route has a problem though. we cannot use zig as the default build without breaking BC.

Just modifying ini to `SPC_TARGET=musl-static@zig` default. And old project using `SPC_LIBC=musl` or `SPC_LIBC=glibc` will work as usual.

henderkes:
> > That route has a problem though. we cannot use zig as the default build without breaking BC.
> 
> Just modifying ini to `SPC_TARGET=musl-static@zig` default.

That's a BC break if users rely on specific functionality of musl toolchain. Currently the default is musl-toolchain. Also specifying @zig when zig is the only supported preconfigured target is pointless. We don't need different toolchains anymore.

crazywhalecc:
> That's a BC break if users rely on specific functionality of musl toolchain. Currently the default is musl-toolchain. Also specifying @zig when zig is the only supported preconfigured target is pointless. We don't need different toolchains anymore.

In 3.0, we will ignore `@zig` suffix and make zig as default target. Current in 2.x we will keep using `musl-static` with `musl` toolchain as default target. But once we have verified that zig can build a stable musl-static (at least for two or more projects using), I think we can release 3.0 soon. Just like when we used musl-toolchain to replace musl-wrapper.

I mentioned many times above that we no longer need multiple toolchains on Linux, and that does not mean that different systems must use the same toolchain. So I also abstracted the clang build of MacOS into TARGET and the corresponding clang toolchain (I will also put msvc toolchain into it). This is another reason why I submitted this PR, which is compatible with the current situation and convenient for deleting unused tool chains in the future.

henderkes:
But in the current implementation anything but @zig is not deprecated and @zig is not the default.
We need to make SPC_TARGET use zig by default, other SPC_TARGETs should be deprecated because we will remove them in 3.0.

But really, there's no reason to rename SPC_LIBC to SPC_TARGET for now at all then. Just keep using SPC_LIBC and deprecate SPC_LIBC. SPC_TARGET can be added in a new PR. I will create a PR on top of this to show you what I mean tomorrow.

henderkes:
Also, the reason why I don't want SPC_TARGET to be an enum is that *all* of the following are valid x86_64 options and can be used safely.
```c
-target x86_64-redhat-linux-gnu.2.17
-target x86_64-suse-linux-gnu.2.17 # centos 7+
-target x86_64-linux-gnu.2.17 # equals x86_64-unknown-linux-gnu.2.17 
-target x86_64-linux-gnu.2.25 # rhel 8+
-target x86_64-linux-musl # equals target x86_64-unknown-linux-musl -static 
-target x86_64-linux-musl.1.25
-target x86_64-linux-musl -dynamic
-target native-native-linux-gnu
-target native-native-redhat-linux-gnu
# and so on
```

crazywhalecc:
I'm sorry to have argued with you for so long. I understood your idea of ​​SPC_TARGET as a zig-only one from the beginning - it has little to do with the core solution of this PR. I may have overthought it a little, so I **calmly** read through the content we discussed and the zig implementation branch again, and concluded that this PR is generally reasonable, but it is indeed more reasonable to modify the SPC_TARGET part in your way. (Also sorry, I used some translations below, and I may be faster in typing some expressions in Chinese)

## Part 1 Explanation

In other words, our goal of iterating to the new version is the same, but the iteration method is slightly different. My PR does not conflict with the Zig goal, but reduces the cost of Zig integration through architectural optimization. Through the three-step strategy of "modular toolchain → add Zig → eliminate the old toolchain", we can not only ensure the stability of the current project, but also lay a solid foundation for the cross-platform Zig compilation of version 3.0.

Of course, I haven't finished the current PR. The goal is to set patch points for almost all judgment parts and place such toolchain-specific patches in the corresponding Toolchain and Target. This will make it very easy to delete the toolchain in the future. Don't worry, I will complete this, and I will do it when refactoring error handling and patch points. The reason I haven't done it now is that I am still writing new documents, and for newly added patch points, I need to write documents before introducing them, otherwise it is difficult to infer which patch points there are from the code. I know this has little to do with SPC_TARGET. This PR is more like a decoupling of the toolchain, and by the way, SPC_LIBC is mapped to SPC_TARGET, and only this conflicts with your logic.

## Part 2 SPC_TARGET Optimization in this PR

In addition, regarding the zig target, I just read the zig document. It is indeed more advantageous to use the zig format as a target and as a fill-in-the-blank question, so I think maybe the following changes can be made:

The current PR's Toolchain remains independent, but SPCTarget is changed to:

1. Add a target parsing function for the zig format so that we can determine the target and version of the build.
2. Before implementing zig, SPC_LIBC will continue to be used as a usable value and determine the musl-toolchain toolchain. When the SPC_LIBC variable is detected, the old toolchain will continue to be used (and a prompt will be prompted that SPC_LIBC will be deprecated).
3. SPC_LIBC is still retained in the configuration but @deprecated is prompted. If you need to use zig, you need to explicitly specify SPC_TARGET and delete SPC_LIBC.

## Part 3 3.0 Plan

Based on this and zig PRs, for 3.0, the following work is to be done:

1. Delete SPC_LIBC support.
2. Delete two toolchains support under Linux (gcc-native, musl).
3. Replace Docker commands with error prompts, and the prompt content is the equivalent SPC_TARGET.
4. Update the document section.
5. Build glibc by default.

henderkes:
>I understood your idea of ​​SPC_TARGET as a zig-only one from the beginning - it has little to do with the core solution of this PR.

Yes! All I want changed is that we don't use the SPC_TARGET env variable for toolchain detection because I want it to be the clear migration path to using zig. 

> Add a target parsing function for the zig format so that we can determine the target and version of the build.

I'm not sure if we really need to. The only part we would probably have to parse out is -static (implied on musl) or -dynamic. Because frankenphp needs to pass -static -static-pie to the go linker too for static builds.

> Build glibc by default.

I haven't thought about that yet. We could keep musl (static) the default as it's so simple to switch with SPC_TARGET.

We can build musl -dynamic on glibc systems (install `musl`) and I think glibc (-dynamic) on musl (install `gcompat libstd++`). I've tested a minimal example of the first.

crazywhalecc:
Good. I will make some changes to this PR tomorrow and leave the zig part. You could complete it in the zig branch after this PR merged.

henderkes:
zig branch is now ready, I can compile:

musl-dynamic >= 1.20 (maybe lower?)
musl-static (default for musl)
glibc (-dynamic, static is not possible) >= 2.17

all ZTS extensions (shared and static) except for grpc, which always fails for me, no matter the compiler

crazywhalecc:
> all ZTS extensions (shared and static) except for grpc, which always fails for me, no matter the compiler

I think we need to open a separate issue or branch to deal with grpc issues. Its official default static build method conflicts with our own dependencies. Maybe using CMake is a workaround, but it would be better to let grpc's dependencies use spc's source instead of its several gigabytes git submodule. Every time I pull it on a server behind THE firewall, I get errors over and over again, which is really annoying.

henderkes:
Yeah I'm just leaving grpc alone. When a library fails to build from source with its given instructions, I'm not really interested in spending hours to fix it.

config/env.ini
```diff
@@ -55,6 +55,8 @@ SPC_CMD_VAR_FRANKENPHP_XCADDY_MODULES="--with github.com/dunglas/frankenphp/cadd
 ; EXTENSION_DIR=
 
 [windows]
+; build target: win7-static
+SPC_TARGET=native-windows
```
henderkes, line 59: replace with SPC_LIBC

config/env.ini
```diff
@@ -109,6 +122,9 @@ SPC_CMD_VAR_PHP_MAKE_EXTRA_LDFLAGS=""
 SPC_CMD_VAR_PHP_MAKE_EXTRA_LDFLAGS_PROGRAM="-all-static -Wl,-O1 -pie"
 
 [macos]
+; build target: macho or macho (possibly we could support macho-universal in the future)
+; Currently we do not support universal and cross-compilation for macOS.
+SPC_TARGET=native-macos
```
henderkes, line 127: replace with SPC_LIBC

src/SPC/util/SPCTarget.php
```diff
@@ -0,0 +1,87 @@
...
+    {
+        $env = getenv('SPC_TARGET');
+        $libc = getenv('SPC_LIBC');
+        // if SPC_LIBC is set, it means the target is static, remove it when 3.0 is released
+        if ($libc === 'musl') {
+            return true;
+        }
+        // TODO: add zig target parser here
+        return false;
+    }
```
henderkes: isStatic

src/globals/test-extensions.php
```diff
@@ -21,15 +21,15 @@
...
-    // 'macos-14',
-    // 'macos-15',
-    // 'ubuntu-latest',
-    // 'ubuntu-22.04',
-    // 'ubuntu-24.04',
-    // 'ubuntu-22.04-arm',
-    // 'ubuntu-24.04-arm',
-    'windows-latest',
+    // 'macos-13', // bin/spc for x86_64
+    // 'macos-14',  // bin/spc for arm64
```
henderkes, line 25: I think we can remove macos-14, there's no point in having it. We only ever use 13 and 15 because they're different architectures, anyway.

config/env.ini
crazywhalecc: Why? Windows and macOS doesn't need libc judgement all the time.

src/SPC/util/SPCTarget.php
crazywhalecc: This can be changed, but there is ambiguity whether using isStaticTarget or isStatic.

src/globals/test-extensions.php
crazywhalecc: Let's just remove it in any test modification.

src/SPC/util/SPCTarget.php
henderkes: Target is already included in the classname.

SPCTarget::isStaticTarget is superfluous.

config/env.ini
henderkes: Because windows and macos will also use zig's target triple later. The SPC_TARGET variable should be the way to switch away from our current tooling and instead use zig.

henderkes: (and having SPC_TARGET=native-windows is silly, when SPC_TARGET will be something like `x86_64-windows-msvc -static` later)

crazywhalecc: So I will leave it alone and leave it for future changes. There is no need for us to argue about this now.

bin/spc-gnu-docker
```diff
@@ -107,11 +107,11 @@ ADD ./bin/setup-runtime /app/bin/setup-runtime
 ADD ./bin/spc /app/bin/spc
 RUN /app/bin/setup-runtime
 RUN /app/bin/php /app/bin/composer install --no-dev
-ENV PATH="/app/bin:/cmake/bin:$PATH"
 ENV SPC_LIBC=glibc
+ENV PATH="/app/bin:/cmake/bin:/opt/rh/devtoolset-10/root/usr/bin:\$PATH"
```
henderkes, line 111: the enable script of the toolset does this already

bin/spc-gnu-docker
```diff
@@ -107,11 +107,11 @@ ADD ./bin/setup-runtime /app/bin/setup-runtime
...
 
 ADD ./config/env.ini /app/config/env.ini
-RUN CC=gcc bin/spc doctor --auto-fix --debug
+RUN CC=gcc bin/php bin/spc doctor --auto-fix --debug
```
henderkes: hmm, why?

bin/spc-gnu-docker
crazywhalecc: Yet we need to build make and automake below.

bin/spc-gnu-docker
crazywhalecc: Ah just a mistake but did not influence anything. Because I have wrong PATH before and `/usr/bin/env` cannot find PHP at that time. The php in gnu-docker is exactly `bin/php`. It doesn’t really matter.

bin/spc-gnu-docker
henderkes: I mean the scl_enable call, not what we do in our SPC code.

henderkes:
Looks like the scl_enable isn't enough and we actually need to overwrite the path, weird.

bin/spc-gnu-docker
```diff
@@ -192,5 +187,5 @@ if [ "$SPC_DOCKER_DEBUG" = "yes" ]; then
     set -ex
     $DOCKER_EXECUTABLE run $PLATFORM_ARG --privileged --rm -it $INTERACT $ENV_LIST --env-file /tmp/spc-gnu-docker.env $MOUNT_LIST cwcc-spc-gnu-$SPC_USE_ARCH-$SPC_DOCKER_VERSION /bin/bash
 else
-    $DOCKER_EXECUTABLE run $PLATFORM_ARG --rm $INTERACT $ENV_LIST --env-file /tmp/spc-gnu-docker.env $MOUNT_LIST cwcc-spc-gnu-$SPC_USE_ARCH-$SPC_DOCKER_VERSION bin/spc $@
+    $DOCKER_EXECUTABLE run $PLATFORM_ARG --rm $INTERACT $ENV_LIST --env-file /tmp/spc-gnu-docker.env $MOUNT_LIST cwcc-spc-gnu-$SPC_USE_ARCH-$SPC_DOCKER_VERSION /app/bin/php bin/spc $@
```
henderkes:
```suggestion
    $DOCKER_EXECUTABLE run $PLATFORM_ARG --rm $INTERACT $ENV_LIST --env-file /tmp/spc-gnu-docker.env $MOUNT_LIST cwcc-spc-gnu-$SPC_USE_ARCH-$SPC_DOCKER_VERSION bin/spc $@
```

bin/spc-gnu-docker
```diff
@@ -107,8 +107,8 @@ ADD ./bin/setup-runtime /app/bin/setup-runtime
 ADD ./bin/spc /app/bin/spc
 RUN /app/bin/setup-runtime
 RUN /app/bin/php /app/bin/composer install --no-dev
-ENV PATH="/app/bin:/cmake/bin:$PATH"
 ENV SPC_LIBC=glibc
+ENV PATH="/app/bin:/cmake/bin:\$PATH"
```
henderkes:
```suggestion
ENV PATH="/app/bin:/cmake/bin:/opt/rh/devtoolset-10/root/usr/bin:\$PATH"
```

henderkes (approved):
when you fix the musl toolchain install this should be good to be merged

I just realized native-windows and native-macos are actually valid targets, so yes, nothing speaks against using them

henderkes:
no idea why the patches are failing on alpine now, I can't reproduce it on my local alpine vm

crazywhalecc:
> no idea why the patches are failing on alpine now, I can't reproduce it on my local alpine vm

Because current main branch has duplicate patch bug and #811 fixed it. The `tests` workflow will build twice (the second time is to test embed build on *nix), that's why only workflow fails and your local is good.

But that pr brings new bug for windows build and I haven't figured it out.

henderkes:
looks like it introduces a runtime mismatch, maybe one of the patches that previously fixed that isn't running through successfully?

This is good to be merged in my eyes, after micro.

Issue #818, opened 2025-07-12
Windows php.exe built by SPC cannot load shared extensions


Mr-ShiHuaYu:
![Image](https://github.com/user-attachments/assets/64b15c0c-597f-404d-b4ef-70d6a455fce5)

如图，php.exe的同目录放了一个php.ini，php版本是8.2的，8.1的也试过，在php.ini中，添加了extension，运行php -m时，一开始一直提示php8.dll错误，所以我从之前的非静态php安装目录下拷贝了一个php8.dll，放到同目录下，然后不弹窗提示php8.dll错误了。
运行php -m，显示警告，api不对，扩展也没有加载成功，我觉得是不是php8.dll的问题？还是说我的扩展的编译环境和这个静态php不一致？我想要的其实就是能够动态的加扩展，谢谢。

henderkes:
Api version mismatch happens when you compile static-php-cli with say version 8.2.39 but you put the extension dll from php 8.3.23 in to be loaded.

You need to compile php with the same php version as the extension was compiled with.

henderkes:
Oh, I didn't see the screenshot. The module version match, but the build system string carries the build system with it. It seems that the extension was built with it set to "VS16", which is Visual Studio 2019 (unknown version), while our php is set to 19.X (Visual Studio 2022) version X. Maybe there's a way to patch that? No idea.

henderkes:
https://github.com/php/php-src/blob/master/Zend/zend_extensions.h#L54
https://github.com/php/php-src/blob/a402edac1a37ee3013050d72d8a3278b179d7eaa/Zend/zend_build.h#L38
https://github.com/php/php-src/blob/a402edac1a37ee3013050d72d8a3278b179d7eaa/win32/build/confutils.js#L2961

henderkes:
https://github.com/php/php-src/blob/a402edac1a37ee3013050d72d8a3278b179d7eaa/win32/build/confutils.js#L3122

Looks like a bug (?) in php's source that doesn't default the compatibility level to VS17 if the compiler is newer than 19.5? In any case we'd need to use an older compiler version (VS 19) when targeting <8.4. Or we just patch the compatibility level to depend on the target version; rather than the users compiler, and hope nothing breaks. I wouldn't say the latter was an option if this wasn't Microsoft we're talking about, but it is, so it will most likely be fine.

Mr-ShiHuaYu:
[...]

PR #819, opened 2025-07-15
Add PHP 8.5 initial support

### LINKED ISSUE #839: Switch argon2 implementation to openssl in 8.5
https://github.com/php/frankenphp/pull/1732#issuecomment-3132993642

crazywhalecc:
## What does this PR do?

Support PHP 8.5 alpha.

- Fix #839 

## TODO

- [x] Check and fix upstream phpmicro patches (fixed in https://github.com/dixyes/phpmicro/pull/5 )
  - [X] cli_checks
  - [X] comctl32
  - [X] disable_huge_page
  - [X] macos_iconv
  - [X] phar
  - [X] static_extensions_win32
  - [X] static_opcache (PHP 8.5 doesn't need to patch opcache)
  - [X] vcruntime140
  - [X] win32
  - [X] zend_stream
- [x] Update static-php.dev docs
- [x] Update README.md
- [x] ~~Waiting upstream response https://github.com/easysoft/phpmicro/issues/23~~ https://github.com/static-php/phpmicro/pull/8/files

Tracked all unsupported things and future versions in #842 .

## Checklist before merging

  - [X] `composer analyse`
  - [X] `composer test`
  - [x] If adding new or fixing bugs, add commit message containing `extension test` or `test extensions` to trigger full test suite.

### DIFF
--- a/src/SPC/builder/extension/openssl.php
+++ b/src/SPC/builder/extension/openssl.php
@@ -28,7 +28,7 @@ class openssl extends Extension
     {
         $openssl_dir = $this->builder->getPHPVersionID() >= 80400 ? '' : ' --with-openssl-dir=' . BUILD_ROOT_PATH;
         $args = '--with-openssl=' . ($shared ? 'shared,' : '') . BUILD_ROOT_PATH . $openssl_dir;
-        if ($this->builder->getPHPVersionID() >= 80500 || !$this->builder->getOption('enable-zts')) {
+        if ($this->builder->getPHPVersionID() >= 80500 || ($this->builder->getPHPVersionID() >= 80400 && !$this->builder->getOption('enable-zts'))) {
             $args .= ' --with-openssl-argon2 OPENSSL_LIBS="-lz"';
         }
         return $args;
@@ -37,7 +37,7 @@ class openssl extends Extension
     public function getWindowsConfigureArg(bool $shared = false): string
     {
         $args = '--with-openssl';
-        if ($this->builder->getPHPVersionID() >= 80500) {
+        if ($this->builder->getPHPVersionID() >= 80500 || ($this->builder->getPHPVersionID() >= 80400 && !$this->builder->getOption('enable-zts'))) {
             $args .= ' --with-openssl-argon2';
         }
         return $args;
--- a/config/ext.json
+++ b/config/ext.json
@@ -531,9 +531,10 @@
         },
         "notes": true,
         "type": "builtin",
-        "arg-type": "with-path",
+        "arg-type": "custom",
         "lib-depends": [
-            "libargon2"
+            "libargon2",
+            "openssl"
         ]
     },
     "pcntl": {
--- a/src/SPC/builder/extension/password_argon2.php
+++ b/src/SPC/builder/extension/password_argon2.php
@@ -42,4 +42,14 @@ class password_argon2 extends Extension
         }
         return $patched;
     }
+
+    public function getConfigureArg(bool $shared = false): string
+    {
+        if ($this->builder->getLib('openssl') !== null) {
+            if ($this->builder->getPHPVersionID() >= 80500 || ($this->builder->getPHPVersionID() >= 80400 && !$this->builder->getOption('enable-zts'))) {
+                return '--without-password-argon2'; // use --with-openssl-argon2 in openssl extension instead
+            }
+        }
+        return '--with-password-argon2';
+    }
 }


henderkes:
Hahaha, I have been working on this, too.

henderkes:
We should also put a watch on https://github.com/php/php-src/pull/18660 because it might land in 8.5.

henderkes:
It was merged, opcache is always statically compiled into php 8.5.

crazywhalecc:
Looks like we have to wait alpha3 or test master branch directly.

crazywhalecc:
> We should also put a watch on https://github.com/php/php-src/pull/18660 because it might land in 8.5.

Looks like we don't need to patch opcache for PHP 8.5 at all. We only need to add version check and just ignore `opcache` extension like `json`, but the phpmicro project would have to create an empty patch file to avoid backward conflicts.

src/SPC/builder/macos/MacOSBuilder.php
```diff
@@ -97,9 +97,17 @@ public function buildPHP(int $build_target = BUILD_TARGET_NONE): void
...
+        $phpVersionID = $this->getPHPVersionID();
+        $json_74 = $phpVersionID < 80000 ? '--enable-json ' : '';
         $zts = $this->getOption('enable-zts', false) ? '--enable-zts --disable-zend-signals ' : '';
 
+        $opcache_jit = !$this->getOption('disable-opcache-jit', false);
+        if ($opcache_jit) {
+            $opcache_jit = $phpVersionID >= 80500 ? '--enable-opcache-jit ' : '';
+        } else {
+            $opcache_jit = '--disable-opcache-jit ';
+        }
```
henderkes: Revert this please, otherwise we take the ability to compile opcache jit for php < 8.5.

henderkes (changes requested):

src/SPC/builder/macos/MacOSBuilder.php
crazywhalecc: The previous opcache jit option we used (enable is `''`) is working well, but explictly passing it is more clearer.

src/SPC/builder/extension/openssl.php
```diff
@@ -26,6 +26,27 @@ public function patchBeforeMake(): bool
     public function getUnixConfigureArg(bool $shared = false): string
     {
         $openssl_dir = $this->builder->getPHPVersionID() >= 80400 ? '' : ' --with-openssl-dir=' . BUILD_ROOT_PATH;
-        return '--with-openssl=' . ($shared ? 'shared,' : '') . BUILD_ROOT_PATH . $openssl_dir;
+        $args = '--with-openssl=' . ($shared ? 'shared,' : '') . BUILD_ROOT_PATH . $openssl_dir;
+        if (
+            $this->builder->getPHPVersionID() >= 80500 &&
+            ($ver = $this->builder->getLib('openssl')->getLibVersion()) &&
+            version_compare($ver, '3.2.0', '>=')
```
henderkes: we can skip this check given our download will not just decide to revert to an old version

there are many places where we (or php) requires a certain minimum version

src/SPC/store/SourcePatcher.php
```diff
@@ -98,7 +98,9 @@ public static function patchBeforeConfigure(BuilderBase $builder): void
             }
         }
         // patch capstone
-        FileSystem::replaceFileRegex(SOURCE_PATH . '/php-src/configure', '/have_capstone="yes"/', 'have_capstone="no"');
+        if (is_unix()) {
+            FileSystem::replaceFileRegex(SOURCE_PATH . '/php-src/configure', '/have_capstone="yes"/', 'have_capstone="no"');
```
henderkes, line 132: there's a decent chance that I'll try to get capstone working.

I need to cross-confirm, but this was one of the settings that seemed to impact performance a little bit, even though I have no explanation for why it would

I also disabled a few others in the same run though, so might have been them.

crazywhalecc: I suggest that we link or mention this as a task in #838 .

henderkes: I think it was unrelated and instead the re2c option.

src/SPC/util/GlobalEnvManager.php
```diff
@@ -117,6 +118,15 @@ public static function afterInit(): void
...
             ToolchainManager::afterInitToolchain();
         }
+        // test bison
+        if (PHP_OS_FAMILY === 'Darwin') {
+            if ($bison = SystemUtil::findCommand('bison', ['/opt/homebrew/opt/bison/bin', '/usr/local/opt/bison/bin'])) {
+                self::putenv("BISON={$bison}");
+            }
+            if ($yacc = SystemUtil::findCommand('yacc', ['/opt/homebrew/opt/bison/bin', '/usr/local/opt/bison/bin'])) {
+                self::putenv("YACC={$yacc}");
+            }
```
henderkes, line 128: is it to be expected that php 8.5 will require users to have bison, yacc and re2c installed - or is it just temporary because there are no releases with fully compiled autotools?

README-zh.md
```diff
@@ -82,16 +82,17 @@ static-php-cli（简称 `spc`）有许多特性：
...
+|-------------|--------------------|---------------------------------------------------------|
+| 7.2         | :x:                |                                                         |
+| 7.3         | :x:                | phpmicro 和许多扩展不支持 7.3、7.4 版本                            |
+| 7.4         | :x:                | phpmicro 和许多扩展不支持 7.3、7.4 版本                            |
+| 8.0         | :warning:          | PHP 官方已停止 8.0 的维护，我们不再处理 8.0 相关的 backport 支持            |
+| 8.1         | :heavy_check_mark: | PHP 官方仅对 8.1 提供安全更新，在 8.5 发布后我们不再处理 8.1 相关的 backport 支持 |
+| 8.2         | :heavy_check_mark: |                                                         |
+| 8.3         | :heavy_check_mark: |                                                         |
+| 8.4         | :heavy_check_mark: |                                                         |
+| 8.5         | :heavy_check_mark: |                                                         |
```
henderkes: warning here that it's not a stable release

README.md
```diff
@@ -95,10 +95,12 @@ Currently supported PHP versions for compilation:
...
 | 7.4         | :x:                | phpmicro and some extensions not supported on 7.x                                                  |
 | 8.0         | :warning:          | PHP official has stopped maintenance of 8.0, we no longer provide backport support for version 8.0 |
-| 8.1         | :heavy_check_mark: | PHP official has security fixes only                                                               |
+| 8.1         | :heavy_check_mark: | PHP official has security fixes only,  we no longer provide backport support when 8.5 released     |
 | 8.2         | :heavy_check_mark: |                                                                                                    |
 | 8.3         | :heavy_check_mark: |                                                                                                    |
 | 8.4         | :heavy_check_mark: |                                                                                                    |
+| 8.5 (alpha) | :heavy_check_mark: |                                                                                                    |
```
henderkes: same here, we don't have fully compatibility with 8.5 yet because many extensions still require patches

.github/workflows/ext-matrix-tests.yml
```diff
@@ -82,9 +82,9 @@ jobs:
           - zlib
           - zstd
         php-version:
-          - "8.4"
+          - "git"
         operating-system:
-          - "ubuntu-latest"
+          - "ubuntu-24.04"
```
henderkes: intended?

src/SPC/util/GlobalEnvManager.php
crazywhalecc: The PHP always require these tools but the default developer tools version for macOS cannot be used for PHP.

It shouldn't be a problem for PHP 8.5, but accidentally I removed my Xcode and installed pure command line tools. That means I cannot use bison, yacc and re2c with PHP. And GHA has Xcode installed, installing homebrew php will also install these things so we didn't hit this before.

As for re2c, I found that our gnu-docker (which uses CentOS 7) had 0.14.3 and PHP requires >= 1.0.3.

https://github.com/php/php-src/blob/659f55a617a066646f808941cb61cc1fc13361c2/configure.ac#L151

I'm not sure why 8.5 will check it, but just in case we check it as well.

crazywhalecc:
Looks good to me, and this is for 2.7.0 release. The rest of 8.5 support will do after upstream fixed and announced.

src/SPC/store/SourcePatcher.php
```diff
@@ -588,13 +588,18 @@ public static function patchGDWin32(): bool
...
         if (!str_contains($file, 'static-php-cli.version')) {
             logger()->debug('Inserting static-php-cli.version to php-src');
             $file = str_replace('PHP_INI_BEGIN()', "PHP_INI_BEGIN()\n\tPHP_INI_ENTRY(\"static-php-cli.version\",\t\"{$version}\",\tPHP_INI_ALL,\tNULL)", $file);
             FileSystem::writeFile(SOURCE_PATH . '/php-src/main/main.c', $file);
         }
+
+        // add PHP_BUILD_PROVIDER
+        if (getenv('PHP_BUILD_PROVIDER') === false) {
+            putenv("PHP_BUILD_PROVIDER=static-php-cli {$version}");
+        }
```
henderkes: we can use GlobalEnvManager::putenv so we can skip the check if it doesn't exist yet.

crazywhalecc: `GlobalEnvManager::putenv` doesn't have checking existence function, but indeed we could add it to `env.ini` to easily modify it for future vendor mode and packaging vendor. But we need expose spc version to env.ini too .

henderkes: I could swear we have a function that checks and only sets if unset. Was that really only in the env.ini handling logic?

crazywhalecc: Yup. https://github.com/crazywhalecc/static-php-cli/blob/34edb6e329fb8f9403e599c794e02d5f7a2db78a/src/SPC/util/GlobalEnvManager.php#L66

crazywhalecc: Previously we have `setOptionIfNotExists` in BuilderBase, but I removed it after env.ini.

config/env.ini
```diff
@@ -43,6 +44,8 @@ SPC_SKIP_PHP_VERSION_CHECK="no"
 SPC_SKIP_DOCTOR_CHECK_ITEMS=""
 ; extra modules that xcaddy will include in the FrankenPHP build
 SPC_CMD_VAR_FRANKENPHP_XCADDY_MODULES="--with github.com/dunglas/frankenphp/caddy --with github.com/dunglas/mercure/caddy --with github.com/dunglas/vulcain/caddy --with github.com/dunglas/caddy-cbrotli"
+; The display message for php version output (PHP >= 8.4 available)
+PHP_BUILD_PROVIDER="static-php-cli ${SPC_VERSION}"
```
henderkes, line 48: there are a couple more environment variables that can be set.

```
PHP_UNAME (probably doesn't make sense overwriting)
PHP_BUILD_SYSTEM
PHP_BUILD_PROVIDER
PHP_BUILD_COMPILER
PHP_BUILD_ARCH
```

And when EXTRA_LDFLAGS contains -release, the ./configure command is not displayed in `phpinfo();`

crazywhalecc: I don't think we should overwrite PHP_BUILD_SYSTEM, PHP_BUILD_ARCH and PHP_UNAME. The default values are enough for everyone.

PHP_BUILD_COMPILER looks like unnecessary for user's change, but could be used internally when using different toolchains like `putenv('PHP_BUILD_COMPILER="Zig $(zig version)')`

crazywhalecc: Added PHP_BUILD_COMPILER.

src/SPC/toolchain/MSVCToolchain.php
```diff
@@ -9,4 +9,9 @@ class MSVCToolchain implements ToolchainInterface
     public function initEnv(): void {}
 
     public function afterInit(): void {}
+
+    public function getCompilerInfo(): ?string
+    {
+        return null;
```
henderkes, line 15: This should be VS16 for php 8.3 and below, VS17 for 8.4+

https://github.com/php/php-src/blob/master/win32/build/confutils.js#L3112

We need this to work so that users can load extension dll's.

See https://github.com/crazywhalecc/static-php-cli/issues/818

crazywhalecc: It's unrelated to `PHP_BUILD_COMPILER`, I think it should be a separated patches because it depends on PHP version. BTW the windows MSVC build does not accept PHP_BUILD_COMPILER at all.

henderkes: I see. We should still patch the VS version like php does it with our windows builds.

crazywhalecc: I've done the patch, but the VS version that shouldn't be used when building PHP <= 8.3 with VS 17.

If I force use `VS16` for 8.2 (locally tested) with VS17 toolset, I cannot load official php_xxx.dll with the same version and TS. I haven't figure it out yet but it should be a true compatibility issue for extension api.

We could only fix the old version to use `VS17` instead of using something like `untested 19.x.x`

crazywhalecc: Oops, seems the spc build of php.exe cannot load shared extensions normally, even if building 8.4.

crazywhalecc: We will need to address this bug to another issue, and it is not related to this PR.

crazywhalecc:
The opcache jit is using wrong macro when building with musl-toolchain: 

https://github.com/php/php-src/blob/55a3e33bb1e38e88b0bd293762346e6c75f9a553/ext/opcache/jit/tls/zend_jit_tls_aarch64.c#L47

https://github.com/php/php-src/blob/55a3e33bb1e38e88b0bd293762346e6c75f9a553/configure.ac#L239

PHP configure is using `ldd` to determine `__MUSL__`, and it's not acceptable for spc's cross toolchain.

henderkes:
That most certainly looks like a terrible idea and should be fixed in php-src. I don't currently know a better way of detecting it, but when we find one we should create a PR. This is essentially problematic for all sorts of cross-compilation.

Although we could add `CPPFLAGS="-D__MUSL__" ` in case of musl target and `-U__MUSL__` in case of gnu target.

crazywhalecc:
A little confusing. When __MUSL__ defined, the openssl sanity check failed with any https url: https://github.com/crazywhalecc/static-php-cli/actions/runs/16692784622/job/47252816667

And `curl` extension is working good with https, only `file_get_contents` is buggy.

henderkes:
I think musl might use different paths for certificate files by default, so it will fail on ubuntu.

crazywhalecc:
```
/app # uname -a
Linux c9636734e919 6.10.14-linuxkit #1 SMP Tue Apr 15 16:00:54 UTC 2025 aarch64 Linux
/app # cat /etc/os-release
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.22.0
PRETTY_NAME="Alpine Linux v3.22"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://gitlab.alpinelinux.org/alpine/aports/-/issues"
```

```
/app # ./php-buggy a.php

Warning: stream_socket_client(): Failed to enable crypto in /app/a.php on line 15

Warning: stream_socket_client(): Unable to connect to ssl://php.net:443 (Unknown error) in /app/a.php on line 15
int(0)
string(0) ""
/app # ./php-common a.php
int(0)
string(0) ""
*.php.net was digitally signed by E5
```

The php built by GHA from previous failed in alpine linux too, and curl works well. So it's not related to certificate path. In all my years of PHP I have never seen an Unknown error.

```php
<?php

$hostname = "php.net";
$ssloptions = array(
    "capture_peer_cert" => true,
    "capture_peer_cert_chain" => true,
    "allow_self_signed"=> false,
    "CN_match" => $hostname,
    "verify_peer" => true,
    "SNI_enabled" => true,
    "SNI_server_name" => $hostname,
);

$ctx = stream_context_create( array("ssl" => $ssloptions) );
$result = stream_socket_client("ssl://$hostname:443", $errno, $errstr, 30, STREAM_CLIENT_CONNECT, $ctx);

var_dump($errno, $errstr);

if ($errno === 0 && $result !== false) {
    $cont = stream_context_get_params($result);
    $x509 = $cont["options"]["ssl"]["peer_certificate"];
    $certparsed = openssl_x509_parse($x509);

    foreach($cont["options"]["ssl"]["peer_certificate_chain"] as $chaincert)
    {
        $chainparsed = openssl_x509_parse($chaincert);
        $chain_public_key = openssl_get_publickey($chaincert);
        $r = openssl_x509_verify($x509, $chain_public_key);
        if ($r==1)
        {
            echo $certparsed['subject']['CN'];
            echo " was digitally signed by ";
            echo $chainparsed['subject']['CN']."\n";
        }
    }
}
```

henderkes:
I find a few related issues like here: https://stackoverflow.com/questions/14078182/openssl-file-get-contents-failed-to-enable-crypto

Question is why that only happens when __MUSL__ is defined

crazywhalecc:
Interesting. When I use `configure.ac` patches instead of passing `CPPFLAGS=-D__MUSL__`, everything works as expected.

So the conclusion for now:

- `configure.ac` musl check not working with cross toolchain.
- `CPPFLAGS=-D__MUSL__` or defining anything else will make stream client with ssl verification failing or build failure: https://github.com/crazywhalecc/static-php-cli/actions/runs/16696190022/job/47260655518

Currently we could know that the ssl bug is not related to `__MUSL__`, but related to `CPPFLAGS` define.

henderkes:
I will upstream a fix for that to php-src. There are better ways to detect musl without breaking cross-compilation.

crazywhalecc:
> I will upstream a fix for that to php-src. There are better ways to detect musl without breaking cross-compilation.

That's great! btw  how do you plan to detect musl from cross toolchain?

And please review all the changes again. I will merge this branch soon and then introduce some logger refactoring based on it.

henderkes:
Musl doesn't define the _USE_GNU macro in features.h, so we can write a feature test

```c
#ifndef _GNU_SOURCE
#define _GNU_SOURCE
#endif
#include <sys/features.h>
#ifdef _USE_GNU
#error Not using musl
#endif
```

and try to compile that. Won't break cross-compilation.

henderkes:
https://github.com/php/php-src/pull/19352

config/lib.json
```diff
@@ -797,6 +797,12 @@
             "depot.h"
         ]
     },
+    "re2c": {
+        "source": "re2c",
+        "bin-unix": [
+            "re2c"
+        ]
+    },
```
henderkes, line 805: do we really need to build re2c ourselves? RHEL at least has it in appstream. Do ubuntu or alpine not?

src/SPC/builder/extension/openssl.php
```diff
@@ -5,11 +5,23 @@
...
+    {
+        // Fix php 8.5 alpha1~4 zts openssl build bug
+        FileSystem::replaceFileStr(
+            SOURCE_PATH . '/php-src/ext/openssl/config.w32',
+            'WARNING("OpenSSL argon2 hashing not supported in ZTS mode for now");',
+            'AC_DEFINE("HAVE_OPENSSL_ARGON2", 1, "Define to 1 to enable OpenSSL argon2 password hashing.");'
+        );
+        return true;
+    }
+
```
henderkes: revert this, it won't be needed anymore

src/SPC/builder/extension/openssl.php
```diff
@@ -27,6 +39,19 @@ public function patchBeforeMake(): bool
     public function getUnixConfigureArg(bool $shared = false): string
     {
         $openssl_dir = $this->builder->getPHPVersionID() >= 80400 ? '' : ' --with-openssl-dir=' . BUILD_ROOT_PATH;
-        return '--with-openssl=' . ($shared ? 'shared,' : '') . BUILD_ROOT_PATH . $openssl_dir;
+        $args = '--with-openssl=' . ($shared ? 'shared,' : '') . BUILD_ROOT_PATH . $openssl_dir;
+        if ($this->builder->getPHPVersionID() >= 80500) {
+            $args .= ' --with-openssl-argon2 OPENSSL_LIBS="-lz"';
```
henderkes, line 32: We need to disable --with-password-argon2 in case this is enabled. Also, this is already valid for php < 8.5 in non ZTS builds.

src/SPC/builder/linux/LinuxBuilder.php
```diff
@@ -335,12 +344,22 @@ protected function buildEmbed(): void
...
+     * @throws \ReflectionException
+     * @throws \Throwable
+     */
     private function getMakeExtraVars(): array
     {
         $config = (new SPCConfigUtil($this, ['libs_only_deps' => true, 'absolute_libs' => true]))->config($this->ext_list, $this->lib_list, $this->getOption('with-suggested-exts'), $this->getOption('with-suggested-libs'));
         $static = SPCTarget::isStatic() ? '-all-static' : '';
         $lib = BUILD_LIB_PATH;
         return [
+            // 'CPPFLAGS' => '-Dsomethinghere',
```
henderkes:
```suggestion
```

src/SPC/builder/macos/MacOSBuilder.php
```diff
@@ -98,9 +98,17 @@ public function buildPHP(int $build_target = BUILD_TARGET_NONE): void
         $this->emitPatchPoint('before-php-configure');
         SourcePatcher::patchBeforeConfigure($this);
 
-        $json_74 = $this->getPHPVersionID() < 80000 ? '--enable-json ' : '';
+        $phpVersionID = $this->getPHPVersionID();
+        $json_74 = $phpVersionID < 80000 ? '--enable-json ' : '';
```
henderkes, line 102:
```suggestion
        $json_74 = $this->getPHPVersionID() < 80000 ? '--enable-json ' : '';
```

src/SPC/builder/macos/MacOSBuilder.php
```diff
@@ -98,9 +98,17 @@ public function buildPHP(int $build_target = BUILD_TARGET_NONE): void
         $this->emitPatchPoint('before-php-configure');
         SourcePatcher::patchBeforeConfigure($this);
 
-        $json_74 = $this->getPHPVersionID() < 80000 ? '--enable-json ' : '';
+        $phpVersionID = $this->getPHPVersionID();
```
henderkes, line 101:
```suggestion
```

src/SPC/builder/windows/WindowsBuilder.php
```diff
@@ -89,6 +89,9 @@ public function buildPHP(int $build_target = BUILD_TARGET_NONE): void
             }
         }
 
+        $opcache_jit = !$this->getOption('disable-opcache-jit', false);
+        $opcache_jit_arg = $opcache_jit ? '--enable-opcache-jit=yes ' : '--enable-opcache-jit=no ';
```
henderkes, line 93: --enable-opcache-jit
--disable-opcache-jit?

Surely windows' autotools understand --disable?

src/SPC/builder/macos/MacOSBuilder.php
henderkes: (aka revert this)

config/lib.json
crazywhalecc: We could go with a pre-built installation, but I've always felt that dealing with different system would be more of an investment than a one-time setup for any environment.

src/SPC/builder/windows/WindowsBuilder.php
crazywhalecc: Both accepted.

config/lib.json
henderkes: I prefer to keep as many things outside our maintenance frame as possible. These kinds of system tools should really not be compiled by us. I would even refactor out pkg-config long term and use the system installation with our PKG_CONFIG_PATH.

Something that is widely available in a recent enough version can be a doctor fix.

src/SPC/builder/extension/openssl.php
```diff
@@ -30,6 +30,8 @@ public function getUnixConfigureArg(bool $shared = false): string
         $args = '--with-openssl=' . ($shared ? 'shared,' : '') . BUILD_ROOT_PATH . $openssl_dir;
         if ($this->builder->getPHPVersionID() >= 80500 || !$this->builder->getOption('enable-zts')) {
             $args .= ' --with-openssl-argon2 OPENSSL_LIBS="-lz"';
+        } else {
+            $args .= ' --without-openssl-argon2';
```
henderkes: this is not needed, I mean that we must disable the fake password-argon2 extension in case openssl-argon2 is used.

henderkes (changes requested):

src/SPC/builder/extension/openssl.php
crazywhalecc: Ah got it. But `--with-password-argon2` is different from `--with-openssl-argon2`. Considering that Herd and some laravel projects use it, and it has not been deprecated yet, we should treat them separately. We cannot remove it until password-argon2 is deprecated in PHP standard extension.

henderkes: with-openssl-argon2 implements the functionality of with-password-argon2. When both are defined, I'm not sure which backend will be used.

Essentially we need to have this check in both extensions and use the openssl one if possible, but fall back to the coding contest implementation.

crazywhalecc: According to config.m4 and function source, `--with-password-argon2` is defined to use the standalone libargon2, while `--with-openssl-argon2` is using openssl. And these are used from different function: `password_hash` and `openssl_password_hash`. They shouldn't conflict.

henderkes: `password_hash` exists either way. When no implementation is chosen, it falls back to libsodium (no threads option). Argon2 and openssl can change the bachend implementation, but I'm not sure which backend will be chosen in case both are enabled. Remi explicitly enables `--with-openssl-argon2` and `--without-password-argon2`.

We want it to use the openssl backend because it's the most well-maintained implementation.

crazywhalecc: Looks like if we enabled libargon2, it will use it first. Otherwise it will search from zend bucket.

Anyway, I don't think we have to adjust pwhash function and dependency here. But if we could confirm and validate that openssl is able to cover it, we could make `password-argon2` to depends on `openssl` instead of libargon2 later.

henderkes: Well, only from php 8.4 for NTS and 8.5 for ZTS sadly. My point is that we should let openssl be the backend for the password_hash function if possible. I'll look into it tomorrow.

henderkes: https://github.com/search?q=repo%3Aphp%2Fphp-src%20php_password_algo_register&type=code

henderkes:
https://github.com/php/php-src/pull/19360

Could simplify it a little.

henderkes:
closes  #845

henderkes (approved):

PR #824, opened 2025-07-18
add Lerc, JBIG and JXL libraries for use in imagick

### LINKED ISSUE #755: imagick jxl format
we're missing this one currently

henderkes:
## What does this PR do?
add -lLerc and -ljbig for use in imagick

closes #755 

## Checklist before merging

  - [x] `composer cs-fix`
  - [x] `composer analyse`
  - [x] `composer test`


henderkes:
seems like macOS 15 is failing... and pulling in prebuilt packages, despite prebuilt option being set to false? Can you take a look @crazywhalecc ?

crazywhalecc:
> seems like macOS 15 is failing... and pulling in prebuilt packages, despite prebuilt option being set to false? Can you take a look @crazywhalecc ?

I'll take a look this weekend.

crazywhalecc:
The only possibility is the actions downloaded pre-built contents before we removing `cleanLaFiles` function. The latest pre-built content has `liblzma.la` file.

But I have no better solution for resolving download cache ttl yet. The easist way is either we add a `invalidate_cache` option for test-extensions.php or ignore all caches.

src/SPC/builder/unix/library/jbig.php
```diff
@@ -0,0 +1,31 @@
...
+        // Build the library
+        shell()->cd($this->source_dir)->initializeEnv($this)
+            ->exec("make -j{$this->builder->concurrency} {$this->builder->getEnvString()} lib")
+            ->exec('cp libjbig/libjbig.a ' . BUILD_LIB_PATH)
+            ->exec('cp libjbig/libjbig85.a ' . BUILD_LIB_PATH)
+            ->exec('cp libjbig/jbig.h ' . BUILD_INCLUDE_PATH)
+            ->exec('cp libjbig/jbig85.h ' . BUILD_INCLUDE_PATH)
+            ->exec('cp libjbig/jbig_ar.h ' . BUILD_INCLUDE_PATH);
+    }
+}
```
crazywhalecc:
```suggestion
<?php

declare(strict_types=1);

namespace SPC\builder\unix\library;

use SPC\exception\FileSystemException;
use SPC\exception\RuntimeException;
use SPC\store\FileSystem;

trait jbig
{
    /**
     * @throws FileSystemException
     */
    public function patchBeforeBuild(): bool
    {
        // Patch Makefile to add -fPIC flag for position-independent code
        FileSystem::replaceFileStr($this->source_dir . '/Makefile', 'CFLAGS = -O2 -W -Wno-unused-result', 'CFLAGS = -O2 -W -Wno-unused-result -fPIC');
        return true;
    }

    /**
     * @throws RuntimeException
     */
    protected function build(): void
    {
        // Build the library
        shell()->cd($this->source_dir)->initializeEnv($this)
            ->exec("make -j{$this->builder->concurrency} {$this->builder->getEnvString()} lib")
            ->exec('cp libjbig/libjbig.a ' . BUILD_LIB_PATH)
            ->exec('cp libjbig/libjbig85.a ' . BUILD_LIB_PATH)
            ->exec('cp libjbig/jbig.h ' . BUILD_INCLUDE_PATH)
            ->exec('cp libjbig/jbig85.h ' . BUILD_INCLUDE_PATH)
            ->exec('cp libjbig/jbig_ar.h ' . BUILD_INCLUDE_PATH);
    }
}
```

crazywhalecc: Using `patchBeforeBuild` prevents redundant patch.

henderkes:
The problem is that it should not even download prebuilt libraries. $prefer_pre_built is set to false.

Also it seems that the problem is that the .la file doesn't exist, even though it should.

src/SPC/builder/unix/library/libtiff.php
```diff
@@ -24,17 +29,21 @@ protected function build(): void
...
+                '--disable-tools',
+                '--disable-contrib',
                 '--disable-cxx',
+                '--without-x',
             )
+            ->optionalLib('lerc', '--enable-lerc', '--disable-lerc')
+            ->optionalLib('zstd', '--enable-zstd', '--disable-zstd')
+            ->optionalLib('webp', '--enable-webp', '--disable-webp')
+            ->optionalLib('xz', '--enable-lzma', '--disable-lzma')
+            ->optionalLib('jbig', '--enable-jbig', '--disable-jbig')
```
crazywhalecc: If libtiff contain optional libs, we cannot provide pre-built libs for libtiff anymore. That is, we need to remove `libtiff.provide-pre-built` line in `source.json`.

config/source.json
```diff
@@ -388,6 +388,15 @@
             "path": "LICENSE"
         }
     },
+    "jbig": {
+        "type": "url",
+        "url": "https://www.cl.cam.ac.uk/~mgk25/jbigkit/download/jbigkit-2.1.tar.gz",
```
crazywhalecc: How about (using https://github.com/ImageMagick/jbig or self-hosted here) and make this as alt? I cannot pull source from cambridge with stable speed, and it has not been updated in 11 years.

crazywhalecc:
> The problem is that it should not even download prebuilt libraries. $prefer_pre_built is set to false.
> 
> Also it seems that the problem is that the .la file doesn't exist, even though it should.

It indeed does not download, when set to false. What I mean by this is that we're using a deprecated cache, and we're now allowing both pre-built content and source content to exist at the same time, but the build will prior use pre-built content. I just deleted all the repository caches and the build was successful.

config/source.json
henderkes: imagemagick jbig is a windows version, not the jbig-kit.

henderkes: we can fork it in static-php though?

crazywhalecc: I've uploaded to spaces: https://dl.static-php.dev/static-php-cli/deps/jbig/jbigkit-2.1.tar.gz

src/SPC/builder/unix/library/libjxl.php
```diff
@@ -0,0 +1,42 @@
...
+        if (str_contains($depsContent, '# return 0')) {
+            return false;
+        }
+        FileSystem::replaceFileStr(
+            $this->source_dir . '/deps.sh',
+            'return 0',
+            '# return 0',
+        );
+        shell()->cd($this->source_dir)
+            ->exec('./deps.sh');
```
henderkes: Not the biggest fan of this, but I couldn't think of a good solution to do it at the download stage. Maybe a source extract hook?

./deps.sh downloads required other libraries.

config/lib.json
```diff
@@ -400,6 +420,21 @@
             "zlib"
         ]
     },
+    "libjxl": {
+        "source": "libjxl",
+        "static-libs-unix": [
+            "libjxl.a",
+            "libjxl_dec.a",
+            "libjxl_threads.a",
+            "libhwy.a"
```
henderkes, line 429: this is not part of libjxl, but libjxl depends on libhwy, we add it here because it's built with libjxl and because we need to specify it for php make EXTRA_LIBS

henderkes (commented):
comments

henderkes:
of course, ancient centos 7 fails again D:

.php-cs-fixer.php
```diff
@@ -4,6 +4,7 @@
 
 return (new PhpCsFixer\Config())
     ->setRiskyAllowed(true)
+    ->setUnsupportedPhpVersionAllowed(true)
```
crazywhalecc, line 7: Nice move

config/lib.json
```diff
@@ -478,16 +513,14 @@
         "static-libs-windows": [
             "libssh2.lib"
         ],
+        "provide-pre-built": true,
```
crazywhalecc: Should be in `source.json`

config/source.json
```diff
@@ -534,6 +557,14 @@
...
         }
     },
+    "libjxl": {
+        "type": "ghtar",
+        "repo": "libjxl/libjxl",
+        "license": {
+            "type": "file",
+            "path": "LICENSE"
+        }
+    },
```
crazywhalecc: Does it support pre-built content?

src/SPC/builder/unix/library/libjxl.php
crazywhalecc: `patchBeforeBuild` will put `.spc.patched` file in source dir, it doesn't need judge patch manually.

src/SPC/builder/unix/library/libjxl.php
```diff
@@ -0,0 +1,43 @@
...
+    protected function build(): void
+    {
+        UnixCMakeExecutor::create($this)
+            ->addConfigureArgs('-DJPEGXL_ENABLE_TOOLS=OFF')
+            ->addConfigureArgs('-DJPEGXL_ENABLE_EXAMPLES=OFF')
+            ->addConfigureArgs('-DJPEGXL_ENABLE_MANPAGES=OFF')
+            ->addConfigureArgs('-DJPEGXL_ENABLE_BENCHMARK=OFF')
+            ->addConfigureArgs('-DJPEGXL_ENABLE_PLUGINS=OFF')
+            ->addConfigureArgs('-DJPEGXL_ENABLE_JNI=OFF')
+            ->addConfigureArgs('-DJPEGXL_STATIC=' . (SPCTarget::isStatic() ? 'ON' : 'OFF'))
```
crazywhalecc, line 23: If I understand correctly, this `JPEGXL_STATIC` should be the difference between static and dynamic construction of jxl itself, right? Regardless of building shared or static targets, we always require dependencies to be static.

src/SPC/builder/unix/library/libtiff.php
```diff
@@ -16,24 +18,31 @@ trait libtiff
      */
     protected function build(): void
     {
+        $libcpp = SPCTarget::getTargetOS() === 'Linux' ? '-lstdc++' : '-lc++';
+        FileSystem::replaceFileStr($this->source_dir . '/configure', '-lwebp', '-lwebp -lsharpyuv');
+        FileSystem::replaceFileStr($this->source_dir . '/configure', '-l"$lerc_lib_name"', '-l"$lerc_lib_name" ' . $libcpp);
```
crazywhalecc, line 23: Is this modification necessary? I tested it on my mac and it seemed to work the same with or without it.

config/source.json
henderkes: no, it has lib-suggests (many, actually)

src/SPC/builder/unix/library/libjxl.php
henderkes: oh right, thanks

src/SPC/builder/unix/library/libjxl.php
henderkes: JPEGXL_STATIC adds -static to ldflags, which causes the build to fail on any libc implementation that does not support static linking

It does not control whether shared or static libraries are built

henderkes: (well, it does mean that shared libraries won't be built, even if you otherwise enable them, but we don't in base cmake args)

src/SPC/builder/unix/library/libtiff.php
henderkes: yes, it fails linking otherwise. lwebp requires lsharpyuv, but the config.m4 doesn't properly call pkgconfig and only adds -lwebp.

lerc also requires c++, but the makefile uses CC to build instead of CXX, leading to undefined references to c++ symbols

crazywhalecc (approved):
I would rather using `git` type for libjxl, that will automatically pull submodules.

As for redundant submodule, I decide to add `submodules` option to `git` source.json that we can specify the required submodule only. I'll do it in another branch later and then merge to this.

But rest of content looks good to me and ready for merging.

PR #826, opened 2025-07-20
doesn't even work yet, there has to be a better way than to add 29139…

henderkes:
…12309 libraries

## Checklist before merging

  - [ ] `composer analyse`
  - [ ] `composer test`

### DIFF
--- a/config/lib.json
+++ b/config/lib.json
@@ -186,12 +186,12 @@
     "grpc": {
         "source": "grpc",
         "static-libs-unix": [
-            "libgrpc.a",
-            "libcares.a"
+            "libgrpc.a"
         ],
         "lib-depends": [
             "zlib",
-            "openssl"
+            "openssl",
+            "libcares"
         ],
         "frameworks": [
             "CoreFoundation"
--- a/src/SPC/builder/extension/grpc.php
+++ b/src/SPC/builder/extension/grpc.php
@@ -36,9 +36,20 @@ class grpc extends Extension
         }
         return false;
     }
+    public function patchBeforeConfigure(): bool
+    {
+        $libs = ' -l' . join(' -l', $this->getLibraries());
+        FileSystem::replaceFileStr(SOURCE_PATH . '/php-src/configure', '-lgrpc', $libs);
+        return true;
+    }
 
     public function patchBeforeMake(): bool
     {
+        $extra_libs = trim(getenv('SPC_EXTRA_LIBS'));
+        $alibs = join('.a ', $this->getLibraries()) . '.a';
+        $libs = BUILD_LIB_PATH . '/lib' . str_replace(' ', ' ' . BUILD_LIB_PATH . '/lib', $alibs);
+        $extra_libs = str_replace(BUILD_LIB_PATH . '/libgrpc.a', $libs, $extra_libs);
+        f_putenv('SPC_EXTRA_LIBS=' . $extra_libs);
         // add -Wno-strict-prototypes
         GlobalEnvManager::putenv('SPC_CMD_VAR_PHP_MAKE_EXTRA_CFLAGS=' . getenv('SPC_CMD_VAR_PHP_MAKE_EXTRA_CFLAGS') . ' -Wno-strict-prototypes');
         return true;
@@ -47,5 +58,13 @@ class grpc extends Extension
     public function getUnixConfigureArg(bool $shared = false): string
     {
         return '--enable-grpc=' . BUILD_ROOT_PATH . '/grpc GRPC_LIB_SUBDIR=' . BUILD_LIB_PATH;
+    }
+
+    private function getLibraries(): array
+    {
+        [, $out] = shell()->execWithResult('$PKG_CONFIG --libs-only-l grpc');
+        $libs = join(' ', $out) . ' -lupb -lupb_message_lib -lupb_json_lib -lupb_textformat_lib -lupb_mini_descriptor_lib -lupb_wire_lib -lupb_mem_lib -lupb_base_lib -lutf8_range';
+        $libs = str_replace('-l', '', $libs);
+        return explode(' ', $libs);
     }
 }
--- a/src/SPC/builder/unix/library/grpc.php
+++ b/src/SPC/builder/unix/library/grpc.php
@@ -4,22 +4,33 @@ declare(strict_types=1);
 
 namespace SPC\builder\unix\library;
 
-use SPC\store\FileSystem;
+use SPC\util\executor\UnixCMakeExecutor;
 
 trait grpc
 {
     protected function build(): void
     {
-        shell()->cd($this->source_dir)
-            ->exec('EXTRA_DEFINES=GRPC_POSIX_FORK_ALLOW_PTHREAD_ATFORK EMBED_OPENSSL=false CXXFLAGS="-L' . BUILD_LIB_PATH . ' -I' . BUILD_INCLUDE_PATH . '" make static -j' . $this->builder->concurrency);
-        copy($this->source_dir . '/libs/opt/libgrpc.a', BUILD_LIB_PATH . '/libgrpc.a');
-        copy($this->source_dir . '/libs/opt/libboringssl.a', BUILD_LIB_PATH . '/libboringssl.a');
-        if (!file_exists(BUILD_LIB_PATH . '/libcares.a')) {
-            copy($this->source_dir . '/libs/opt/libcares.a', BUILD_LIB_PATH . '/libcares.a');
-        }
-        FileSystem::copyDir($this->source_dir . '/include/grpc', BUILD_INCLUDE_PATH . '/grpc');
-        FileSystem::copyDir($this->source_dir . '/include/grpc++', BUILD_INCLUDE_PATH . '/grpc++');
-        FileSystem::copyDir($this->source_dir . '/include/grpcpp', BUILD_INCLUDE_PATH . '/grpcpp');
-        FileSystem::copyDir($this->source_dir . '/src/php/ext/grpc', BUILD_ROOT_PATH . '/grpc_php_ext_src');
+        UnixCMakeExecutor::create($this)
+            ->addConfigureArgs(
+                '-DgRPC_SSL_PROVIDER=openssl',
+                '-DgRPC_INSTALL_BINDIR=' . BUILD_BIN_PATH,
+                '-DgRPC_INSTALL_LIBDIR=' . BUILD_LIB_PATH,
+                '-DgRPC_INSTALL_SHAREDIR=' . BUILD_ROOT_PATH . '/share/grpc',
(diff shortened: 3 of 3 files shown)


henderkes:
the grpc build is so fucked, it's randomly failing at different percentages with 
```log
[ 65%] Building CXX object CMakeFiles/grpc.dir/src/core/lib/security/credentials/credentials.cc.o
In file included from /home/m/static-php-cli/source/grpc/src/core/lib/security/security_connector/ssl_utils.h:38,
                 from /home/m/static-php-cli/source/grpc/src/core/lib/security/credentials/tls/grpc_tls_certificate_distributor.h:33,
                 from /home/m/static-php-cli/source/grpc/src/core/lib/security/credentials/tls/grpc_tls_certificate_provider.h:33,
                 from /home/m/static-php-cli/source/grpc/src/core/lib/security/credentials/channel_creds_registry_init.cc:37:
/home/m/static-php-cli/source/grpc/src/core/tsi/ssl/key_logging/ssl_key_logging.h:22:10: fatal error: openssl/ssl.h: No such file or directory
   22 | #include <openssl/ssl.h>
      |          ^~~~~~~~~~~~~~~
compilation terminated.
```

Despite the configure phase going through and -I in cppflags and openssl_include_dir set.

If I just restart the build a few times it suddenly succeeds. ???

henderkes:
linux native build is working, musl toolchain build seems to link against conflicting -lstdc++.so.6?

No idea what's going on with macOS.

henderkes:
How are the downloads already downloaded when running spc download for the first time? Are the runners not running from a clean slate anymore?

https://github.com/crazywhalecc/static-php-cli/actions/runs/16413885495/job/46375194631?pr=826#step:12:124

It detects a mismatch and deletes the build folder, leading to failure. Locally, it never does that.

crazywhalecc:
> How are the downloads already downloaded when running spc download for the first time? Are the runners not running from a clean slate anymore?

GHA of course starts from a clean environment, but to prevent repeated download requests from exceeding the rate limit, the workflow will pull the cache of the downloads directory.

https://github.com/crazywhalecc/static-php-cli/blob/f3400b875692652ace66c066ef117be8f6a16dd1/.github/workflows/tests.yml#L168-L172

henderkes:
> > How are the downloads already downloaded when running spc download for the first time? Are the runners not running from a clean slate anymore?
> 
> GHA of course starts from a clean environment, but to prevent repeated download requests from exceeding the rate limit, the workflow will pull the cache of the downloads directory.
> 
> https://github.com/crazywhalecc/static-php-cli/blob/f3400b875692652ace66c066ef117be8f6a16dd1/.github/workflows/tests.yml#L168-L172

Hmm, sadly it seems to cause issues with the hash not matching and spc destroying the source folder (and with that the contents of the required directory, since the ext code lives in there).

henderkes:
I'll leave how to solve that up to you, same as macOS support. I might buy a mac later this year, but I don't really have a use case for one outside of spc.

crazywhalecc:
> https://github.com/crazywhalecc/static-php-cli/actions/runs/16413885495/job/46375194631?pr=826#step:12:124

This is not because of source locks and cache issues, this is because there is a `BUILD` file in the grpc directory, and the default build directory for our cmake executor is `build`. In macOS, because the default file system is case-insensitive (this is stupid), it prompts a `Not a directory` error. 🥲

I got the same error when building highway and it was very misleading.

henderkes:
That explains the macOS failure, but not the hash mismatch in the embed build invocation.

crazywhalecc:
Seems micro and grpc first-time extraction is a little weird: https://github.com/crazywhalecc/static-php-cli/actions/runs/16413885495/job/46375194631?pr=826#step:11:135

henderkes:
Ah, maybe because it preserves git repository information? Then it will complain if ownership does not match root:root in the docker container.

crazywhalecc:
It's possible, I can't find where the lock hashing is working incorrectly, it's just possible that it's really broken. I just checked the cache and found that GitHub prompted that the cache over 10GB will be deleted. I just cleared all the caches and re-triggered the build, and it seems to work well.

henderkes:
Okay good. We could probably work around the BUILD file issue by using ->setBuildDir("fix_macos_build").

henderkes:
okay it builds, the last issue will be easy to fix. Just need to check for existence of the libraries to filter out pthread or resolv.

henderkes:
```fail
[05:26:19] [I] Entering dir: /Users/runner/work/static-php-cli/static-php-cli/source/embed-test
[05:26:19] [I] [EXEC] ./embed
[05:26:19] [D] Executed at: /Users/runner/work/static-php-cli/static-php-cli/src/SPC/util/UnixShell.php:95
dyld[98170]: Library not loaded: @rpath/libz.1.dylib
  Referenced from: <49B5CCE1-2821-39CE-9A44-348D6EB74DD3> /Users/runner/work/static-php-cli/static-php-cli/source/embed-test/embed
  Reason: no LC_RPATH's found
sh: line 1: 98170 Abort trap: 6           ./embed
   SPC\exception\RuntimeException 
  embed failed sanity check: run failed. Error message:
  ```
  
  Yeah I don't think I can debug this from the github actions only :P

File: src/SPC/util/SPCConfigUtil.php [hunk 1]
```diff
@@ -106,10 +106,33 @@ private function getLibsString(array $libraries, bool $withDependencies = false)
...
+
+                if (!empty($llibs)) {
+                    $filtered = str_replace('-pthread', '', $llibs);
+                    $filtered = preg_replace('/-L\S+/', '', $filtered);
+                    $filtered = preg_replace('/(?:\S*\/)?lib([a-zA-Z0-9_+-]+)\.a\b/', '-l$1', $filtered);
+                    $filtered = preg_replace('/\s+/', ' ', $filtered);
+                    foreach (explode(' ', $filtered) as $item) {
+                        $short_name[] = $item;
+                    }
+                } elseif ($withDependencies) {
```
crazywhalecc: What is this used to fix?

henderkes: oh, the pthread one should be changed. 

This is used because grpc needs a literal fuckton of static libraries that change depending on the configuration, so we cannot realistically hardcode them in lib.json.

We should generally use pkgconfig here instead of our own custom logic, to make sure that we can later link against libphp/packages without requiring spc installed.

henderkes:
```
[07:58:16] [D] testing ext: grpc
[07:58:16] [I] [EXEC] /app/buildroot/bin/php -n --ri "grpc"
[07:58:16] [D] Executed at: /app/src/SPC/util/UnixShell.php:95
[07:58:16] [I] [EXEC] /app/buildroot/bin/php -n -r "assert(class_exists('Grpc\ChannelCredentials'));"
[07:58:16] [D] Executed at: /app/src/SPC/util/UnixShell.php:95
[07:58:16] [I]
[07:58:16] [I]    Build complete, used 320.548 s !
[07:58:16] [I]
[07:58:16] [I] Static php binary path (host system): /home/m/static-php-cli/buildroot/bin/php
[07:58:16] [D] Making new directory recursive: /app/buildroot/license
[07:58:16] [I] License path (host system): /home/m/static-php-cli/buildroot/license/
[m@M-TH static-php-cli]$ bin/spc-alpine-docker build grpc --enable-zts --build-cli --debug
```

Everything seems to work, but I'll re-check the grpc failure on ubuntu-latest with tmate.

henderkes:
Found the problem, will simplify the patching logic at the same time.

config/ext.json
```diff
@@ -34,7 +34,7 @@
     },
     "bz2": {
         "type": "builtin",
-        "arg-type-unix": "with-prefix",
+        "arg-type-unix": "with-path",
```
henderkes, line 37: I think this is clearer, because the path is not really a prefix.

config/ext.json
```diff
@@ -233,7 +233,7 @@
         },
         "type": "external",
         "source": "grpc",
-        "arg-type-unix": "custom",
+        "arg-type-unix": "enable-path",
```
henderkes, line 236: grpc is currently the only one using this, but it's a good reminder that --enable also sets $PHP_EXT to the path value. It behaves identical to --with-path except that it seems the latter is convention for paths and enable is convention for yes/no.

But obviously all extensions just pick and choose whatever they feel like...

src/SPC/builder/Extension.php
```diff
@@ -83,13 +83,15 @@ public function getConfigureArg(bool $shared = false): string
      */
     public function getEnableArg(bool $shared = false): string
     {
+        $escapedPath = str_replace("'", '', escapeshellarg(BUILD_ROOT_PATH)) !== BUILD_ROOT_PATH || str_contains(BUILD_ROOT_PATH, ' ') ? escapeshellarg(BUILD_ROOT_PATH) : BUILD_ROOT_PATH;
```
henderkes, line 86: we shouldn't escape it unnecessarily, but if it's necessary we should use escapeshellarg, because windows may need additional escaping other than just ""

henderkes:
@crazywhalecc the macos embed test failure doesn't make much sense to me, as it's also an empty error message. can you check what's going on?

crazywhalecc:
https://github.com/crazywhalecc/static-php-cli/actions/runs/16467368108/job/46547962216#step:12:1111

Looks like the grpc CFLAGS is too long and have some strange parts.

henderkes:
Yeah I don't know where they come from. From the cflags somehow, I will just filter them out.

henderkes:
Works, feel free to review and merge when you can. I would like to get Zig support merged to main within this week, if possible.

src/SPC/util/SPCConfigUtil.php
```diff
@@ -143,6 +143,10 @@ private function getIncludesString(array $libraries): string
             }
             $pc_cflags = implode(' ', $pc);
             if ($pc_cflags !== '' && ($pc_cflags = PkgConfigUtil::getCflags($pc_cflags)) !== '') {
+                $arr = explode(' ', $pc_cflags);
+                $arr = array_unique($arr);
```
crazywhalecc, line 147: I'm not quite sure if there will have a CFLAGS that contains spaces. Simply unique to cflags seems not safe enough.

crazywhalecc (approved):

src/SPC/util/SPCConfigUtil.php
henderkes: Can a cflag be a precursor to another? Something like -target x86_64-linux? I know direct compiler commands can, but I'm not sure of any that would land in cflags.

Also can't think of a better way to do the deduplication. We could live with the long cflags, of course.

crazywhalecc: I didn't find existing CFLAGS with spaces either, the only possibility is that the include paths contain spaces, but it will break autoconf first.

PR #828, opened 2025-07-22
Add real pkg-config integration

crazywhalecc:
## What does this PR do?

Try to make #826 easier.

## Checklist before merging

  - [X] `composer analyse`
  - [X] `composer test`
  - [x] Add your test combination to `src/globals/test-extensions.php`.
  - [x] If adding new or fixing bugs, add commit message containing `extension test` or `test extensions` to trigger full test suite.

### DIFF
--- a/src/SPC/builder/unix/library/libxslt.php
+++ b/src/SPC/builder/unix/library/libxslt.php
@@ -43,7 +43,7 @@ trait libxslt
         }
         $ac->configure()->make();
 
-        $this->patchPkgconfPrefix(['libxslt.pc']);
+        $this->patchPkgconfPrefix(['libexslt.pc', 'libxslt.pc']);
         $this->patchLaDependencyPrefix();
         shell()->cd(BUILD_LIB_PATH)
             ->exec("ar -t libxslt.a | grep '\\.a$' | xargs -n1 ar d libxslt.a")


src/SPC/util/SPCConfigUtil.php
```diff
@@ -81,70 +84,87 @@ public function config(array $extensions = [], array $libraries = [], bool $incl
...
 
-    private function getLibsString(array $libraries, bool $withDependencies = false): string
+    private function getLibsString(array $libraries): string
     {
         $short_name = [];
-        foreach (array_reverse($libraries) as $library) {
+        $frameworks = [];
+
+        foreach ($libraries as $library) {
+            // convert all static-libs to short names
```
henderkes, line 164: thanks, I wanted to do that before

henderkes (approved):
It looks good, but I think we should split the getLibsString method into one that returns short lib names and one that returns absolute names, so we can use it instead of SPC_EXTRA_LIBS for the php make.

crazywhalecc:
@henderkes Added absolute path output for SPCConfigUtil. Now SPC_EXTRA_LIBS is for real extra libs.

henderkes:
Curl runs into a legitimate failure here, the others are just upx download failing.

Edit: extensions are failing too.

config/lib.json
```diff
@@ -36,9 +36,9 @@
     "brotli": {
         "source": "brotli",
         "static-libs-unix": [
+            "libbrotlicommon.a",
             "libbrotlidec.a",
-            "libbrotlienc.a",
-            "libbrotlicommon.a"
+            "libbrotlienc.a"
```
henderkes: This is backwards now. Common must come last. SPCConfigUtil must use the order listed here before.

crazywhalecc: I would use this to let pkg-config automatically order:

```json
{
    "brotli": {
        "source": "brotli",
        "pkg-configs": [
            "libbrotlicommon",
            "libbrotlidec",
            "libbrotlienc"
        ],
        "static-libs-windows": [
            "brotlicommon.lib",
            "brotlienc.lib",
            "brotlidec.lib"
        ],
        "headers": [
            "brotli"
        ]
    }
}
```

henderkes: Yes that would work, but the common library should still come last.

crazywhalecc:
Using `pkg-configs` will make it order-irrelevant. The `gd` failure is caused by `libwebp` incorrect order, after using `pkg-configs`, it's working well.

And we could just replace all static libs that containing `pc` for new libs.

config/lib.json
henderkes: Oh and last but not least, to improve linker performance, when we have a string like `-lssl -lcrypto -lz -lbrotlienc -lbrotlicommon -lbrotlienc -lbrotlicommon -lz` it should be deduplicated into -lssl -lycrypto -lbrotlienc -lbrotlidec -lbrotlicommon -lz` (only keep the last occurence of each library). It's not necessary, but can improve performance a lot.

crazywhalecc: Of course we have: https://github.com/crazywhalecc/static-php-cli/blob/a095d489f35e6d2551ae45aed537708a9d5b36bc/src/SPC/util/SPCConfigUtil.php#L182-L183

But indeed the order may not right, it should `unique` before `reverse`

crazywhalecc:
imagick build failed on my local musl-toolchain with debian. And it shows the libstdc++.la error we have never seen before...

At present, the impact of this PR is greater than expected. Maybe there are many wrong flags that we used.

config/lib.json
```diff
@@ -421,11 +421,11 @@
...
-        "static-libs-unix": [
-            "libjxl.a",
-            "libjxl_cms.a",
-            "libjxl_threads.a",
-            "libhwy.a"
+        "pkg-configs": [
+            "libjxl",
+            "libjxl_cms",
+            "libjxl_threads",
+            "libhwy"
```
henderkes, line 429: I think we only need to use pkgconfig of libjxl and the others should be in the output if needed.

Same goes for other libraries that don't have strictly decoupled components.

crazywhalecc: I thought it before. But it will not work with `$this->patchPkgconfPrefix([])`.

crazywhalecc: Another thing is the `.la` file may contain like `dependency_libs=' -L/app/buildroot/lib /app/buildroot/lib/libgmp.la -lresolv'`. 

And it will cause `libtool: link: cannot find the library '/app/buildroot/lib/libgmp.la' or unhandled argument /app/buildroot/lib/libgmp.la`.

Haven't figure it out yet.

crazywhalecc: Oh, I used pre-built gmp content, and it has `libgmp.la` with `-L/app/buildroot/lib`, and build with musl-toolchain will fail obviously.

crazywhalecc:
Good. We have green CI.

src/SPC/util/PkgConfigUtil.php
```diff
@@ -39,18 +39,23 @@ public static function getLibsArray(string $pkg_config_str, bool $force_short_na
...
         // convert libxxx.a to -L{path} -lxxx
         $exp = explode(' ', trim($result));
         foreach ($exp as $item) {
+            if (str_starts_with($item, '-L')) {
+                $libs[] = $item;
+                continue;
+            }
             // if item ends with .a, convert it to -lxxx
-            if (str_ends_with($item, '.a') && str_starts_with($item, 'lib') && $force_short_name) {
+            if (str_ends_with($item, '.a') && (str_starts_with($item, 'lib') || str_starts_with($item, BUILD_LIB_PATH))) {
```
henderkes, line 51: necessary for -lz in libcrypto.pc, as it contains `BUILD_LIB_PATH/pkgconfig/../../lib/libz.a`

src/SPC/util/PkgConfigUtil.php
```diff
@@ -39,18 +39,23 @@ public static function getLibsArray(string $pkg_config_str, bool $force_short_na
         $libs = explode(' ', trim($result));
 
         // get other things
-        $result = self::execWithResult("pkg-config --static --libs --libs-only-other {$pkg_config_str}");
+        $result = self::execWithResult("pkg-config --static --libs-only-other {$pkg_config_str}");
```
henderkes, line 42: we don't want to add libs again, this adds -lpq again, which will reorder to have pgcommon before pq, which will fail

src/SPC/util/PkgConfigUtil.php
```diff
@@ -39,18 +39,23 @@ public static function getLibsArray(string $pkg_config_str, bool $force_short_na
...
         // convert libxxx.a to -L{path} -lxxx
         $exp = explode(' ', trim($result));
         foreach ($exp as $item) {
+            if (str_starts_with($item, '-L')) {
```
henderkes, line 46: we don't need extra -L paths in the middle of our lib strings

henderkes (commented):
I believe pkgconfig should always return short names, so we can reliably deduplicate. We can use the passed boolean in SPCConfigUtil to switch all -lexample to BUILD_LIB_PATH/libexample.a if it exists, though.

henderkes:
Sorry for the commit, but I think that would have been too many comments.

henderkes:
Ah, test failure because I added -L/buildroot/lib to the $config['libs'] at the start. Sadly the php makefile is stupid and doesn't properly use the ldflags path.

src/SPC/util/PkgConfigUtil.php
crazywhalecc: Oh nice catch. I forgot removing it before.

crazywhalecc:
It seems that relative paths can cause some problems. What do you think about patching the paths of the pc and la files only when packaging the pre-built content? 

But if we do that we need to do this after each library build (by detecting the newly generated pc and la)

henderkes:
>What do you think about patching the paths of the pc and la files only when packaging the pre-built content?

Yes, I think that's sensible.

>But if we do that we need to do this after each library build (by detecting the newly generated pc and la)

If we ignore performance for a second, we could just loop through all .la files in the folder after library build. I can't think of a good way to piece together all .la files a library produces and want to avoid another key in lib.json.

crazywhalecc:
> If we ignore performance for a second, we could just loop through all .la files in the folder after library build. I can't think of a good way to piece together all .la files a library produces and want to avoid another key in lib.json.

I think that shouldn't be a problem because the patching path will only be used in pack-libs command and CI. The default we don't need to patch it at all, we just need to add `patchBeforePack` and `extractPreBuiltHook` two functions.

henderkes:
Oh, the array_reverse was to handle old static-libs cases with dependencies. I'll have to revert that.

henderkes:
> > If we ignore performance for a second, we could just loop through all .la files in the folder after library build. I can't think of a good way to piece together all .la files a library produces and want to avoid another key in lib.json.
> 
> I think that shouldn't be a problem because the patching path will only be used in pack-libs command and CI. The default we don't need to patch it at all, we just need to add `patchBeforePack` and `extractPreBuiltHook` two functions.

Are you sure we don't need to patch the la files before building php? I remember something about that causing fails.

crazywhalecc:
> Are you sure we don't need to patch the la files before building php? I remember something about that causing fails.

The patch I'm talking about is different from the existing `patchLaDependencyPrefix`. I would replace the packaged files with placeholders, and then replace them with the actual locations when unpacking. `patchLaDependencyPrefix` fixes the wrong `/lib/` path pointing, and the two have different purposes.

henderkes:
Aaah that makes sense. Yep.

henderkes:
@crazywhalecc take a look if anything seems off to you and merge this when everything looks good.

crazywhalecc:
Swoole build failed with ubuntu-latest (musl-toolchain).

henderkes:
Yeah to a rather obscure problem at that, turned out that curl didn't list a lib-suggests on ldap, but had a feature check enabled for ldap. So it referenced ldap symbols, but because we didn't have the lib-suggests and curl didn't use pkg-config, it never bothered to pull in -lldap after curl.

crazywhalecc (commented):
Looks like we could merge now.

PR #834, opened 2025-07-25
Fix macOS linking issue for some edge cases  & install-pkg downloading alt bug

crazywhalecc:
## What does this PR do?

- Fix macOS linking issue for some edge cases
- Fix install-pkg downloading alt bug

## Checklist before merging

  - [X] `composer analyse`
  - [X] `composer test`

### DIFF
--- a/config/env.ini
+++ b/config/env.ini
@@ -34,7 +34,6 @@
 ;   SPC_LINUX_DEFAULT_CXX:                  the default c++ compiler for linux. (For alpine linux: `g++`, default: `$GNU_ARCH-linux-musl-g++`)
 ;   SPC_LINUX_DEFAULT_AR:                   the default archiver for linux. (For alpine linux: `ar`, default: `$GNU_ARCH-linux-musl-ar`)
 
-
 [global]
 ; Build concurrency for make -jN, default is CPU_COUNT, this value are used in every libs.
 SPC_CONCURRENCY=${CPU_COUNT}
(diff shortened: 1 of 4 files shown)


config/env.ini
```diff
@@ -155,6 +155,8 @@ SPC_CMD_VAR_PHP_CONFIGURE_CPPFLAGS="-I${BUILD_INCLUDE_PATH}"
 SPC_CMD_VAR_PHP_CONFIGURE_LDFLAGS="-L${BUILD_LIB_PATH}"
 ; EXTRA_CFLAGS for `make` php
 SPC_CMD_VAR_PHP_MAKE_EXTRA_CFLAGS="-g -fstack-protector-strong -fpic -fpie ${SPC_DEFAULT_C_FLAGS}"
+; EXTRA_LDFLAGS_PROGRAM for `make` php
+SPC_CMD_VAR_PHP_MAKE_EXTRA_LDFLAGS_PROGRAM="-L${BUILD_LIB_PATH}"
```
henderkes: again, we should remove these from env.ini, the user doesn't gain anything by setting them

henderkes (changes requested):

config/env.ini
henderkes: also, configure LDFLAGS and make LDFLAGS should always be the same.

henderkes: (with the single exception of removing libraries from ./configure that would add std lib polyfills, like libpgport. 

It's actually beyond me how they thought that was a good idea in the first place.

crazywhalecc:
I agree with hardcoding variables that would cause the build to fail if modified, but I think it's still necessary to expose these additional variables instead of prohibiting external modification of it.

There are many ideas that I think are not suitable for this small PR, I will review them carefully in zig.

henderkes (approved):

henderkes:
> I agree with hardcoding variables that would cause the build to fail if modified, but I think it's still necessary to expose these additional variables instead of prohibiting external modification of it.
> 
> There are many ideas that I think are not suitable for this small PR, I will review them carefully in zig.

I agree, it should be its own PR after zig is merged. Zig currently only affects Linux so I have only removed them there in the zig branch, we can do macOS later.

Issue #842, opened 2025-07-31
PHP 8.5 support state


crazywhalecc:
## PHP 8.5 roadmap support status

- [X] Alpha 1 Announce Files
- [X] Alpha 2 Announce Files
- Alpha 3 (skipped)
- [x] Alpha 4
- [x] Feature freeze
- [x] Beta 1
- [x] Beta 2
- [x] Beta 3
- [ ] RC1
- [ ] RC2
- [ ] RC3
- [ ] RC4
- [ ] GA

## Extension support status (waiting for upstream)

> These are known issues for spc's supported extensions.

- [ ] amqp (ubuntu native build) https://github.com/php-amqp/php-amqp/pull/595
- [ ] swoole https://github.com/swoole/swoole-src/pull/5817
- [x] redis https://github.com/phpredis/phpredis/pull/2676
- [ ] igbinary, memcache, rar, yac, imagick,  https://github.com/remicollet/remirepo/issues/307

PR #853, opened 2025-08-11
Enhance exception handling by binding builder and extra info to exception handler

### LINKED ISSUE #852: TypeError in SPCException
This piece of code tries to get the `options` property from a class within the stack trace.

https://github.com/crazywhalecc/static-php-cli/blob/1e9434221be58c9d0f5f524e6ba229d596ee588a/src/SPC/exception/SPCException.php#L158

This code will always produce a TypeError with an error message like this:

```
Fatal error: Uncaught TypeError: ReflectionProperty::getValue(): Argument #1 ($object) must be provided for instance properties in ...
```

While the first parameter of `ReflectionProperty::getValue()` is technically optional, it can actually only be omitted, if the property is static. But `\SPC\builder\BuilderBase::$options` is in fact not static.

https://github.com/crazywhalecc/static-php-cli/blob/1e9434221be58c9d0f5f524e6ba229d596ee588a/src/SPC/builder/BuilderBase.php#L38

As far as I know, the actual object of that class is not accessible from a stack trace. So, I suggest to make the property static. As I understand it, there will not be multiple instances of `\SPC\builder\BuilderBase` anyway, since the instance is determined once per execution.

crazywhalecc:
## What does this PR do?

Fix #852 

## Checklist before merging

  - [X] `composer cs-fix`
  - [X] `composer analyse`
  - [X] `composer test`

### DIFF
--- a/src/SPC/builder/BuilderProvider.php
+++ b/src/SPC/builder/BuilderProvider.php
@@ -32,7 +32,7 @@ class BuilderProvider
         };
 
         // bind the builder to ExceptionHandler
-        ExceptionHandler::setBindBuilder(self::$builder);
+        ExceptionHandler::bindBuilder(self::$builder);
 
         return self::$builder;
     }
--- a/src/SPC/builder/BuilderProvider.php
+++ b/src/SPC/builder/BuilderProvider.php
@@ -32,7 +32,7 @@ class BuilderProvider
         };
 
         // bind the builder to ExceptionHandler
-        ExceptionHandler::$bind_builder = self::$builder;
+        ExceptionHandler::setBindBuilder(self::$builder);
 
         return self::$builder;
     }
--- a/src/SPC/builder/BuilderProvider.php
+++ b/src/SPC/builder/BuilderProvider.php
@@ -8,6 +8,7 @@ use SPC\builder\freebsd\BSDBuilder;
 use SPC\builder\linux\LinuxBuilder;
 use SPC\builder\macos\MacOSBuilder;
 use SPC\builder\windows\WindowsBuilder;
+use SPC\exception\ExceptionHandler;
 use SPC\exception\WrongUsageException;
 use Symfony\Component\Console\Input\InputInterface;
 
@@ -29,6 +30,10 @@ class BuilderProvider
             'BSD' => new BSDBuilder($input->getOptions()),
             default => throw new WrongUsageException('Current OS "' . PHP_OS_FAMILY . '" is not supported yet'),
         };
+
+        // bind the builder to ExceptionHandler
+        ExceptionHandler::$bind_builder = self::$builder;
+
         return self::$builder;
     }
 
--- a/src/SPC/command/BuildPHPCommand.php
+++ b/src/SPC/command/BuildPHPCommand.php
@@ -167,7 +167,7 @@ class BuildPHPCommand extends BuildCommand
         $this->printFormatInfo($indent_texts);
 
         // bind extra info to exception handler
-        ExceptionHandler::$bind_build_php_extra_info = $indent_texts;
+        ExceptionHandler::setBindBuildPhpExtraInfo($indent_texts);
 
         logger()->notice('Build will start after 2s ...');
         sleep(2);
--- a/src/SPC/command/BuildPHPCommand.php
+++ b/src/SPC/command/BuildPHPCommand.php
@@ -5,7 +5,7 @@ declare(strict_types=1);
 namespace SPC\command;
 
 use SPC\builder\BuilderProvider;
-use SPC\exception\SPCException;
+use SPC\exception\ExceptionHandler;
 use SPC\store\Config;
 use SPC\store\FileSystem;
 use SPC\store\SourcePatcher;
@@ -165,8 +165,9 @@ class BuildPHPCommand extends BuildCommand
         }
         $this->printFormatInfo($this->getDefinedEnvs(), true);
         $this->printFormatInfo($indent_texts);
-        // bind extra info to SPCException
-        SPCException::bindBuildPHPExtraInfo($indent_texts);
+
+        // bind extra info to exception handler
+        ExceptionHandler::$bind_build_php_extra_info = $indent_texts;
 
         logger()->notice('Build will start after 2s ...');
         sleep(2);
--- a/src/SPC/command/BuildPHPCommand.php
+++ b/src/SPC/command/BuildPHPCommand.php
@@ -167,7 +167,7 @@ class BuildPHPCommand extends BuildCommand
         $this->printFormatInfo($indent_texts);
 
         // bind extra info to exception handler
-        ExceptionHandler::setBindBuildPhpExtraInfo($indent_texts);
+        ExceptionHandler::bindBuildPhpExtraInfo($indent_texts);
 
         logger()->notice('Build will start after 2s ...');
         sleep(2);
--- a/src/SPC/exception/ExceptionHandler.php
+++ b/src/SPC/exception/ExceptionHandler.php
@@ -169,7 +169,7 @@ class ExceptionHandler
         self::$bind_builder = $bind_builder;
     }
 
-    public static function setBindBuildPhpExtraInfo(array $bind_build_php_extra_info): void
+    public static function bindBuildPhpExtraInfo(array $build_php_extra_info): void
     {
         self::$bind_build_php_extra_info = $bind_build_php_extra_info;
     }
--- a/src/SPC/exception/ExceptionHandler.php
+++ b/src/SPC/exception/ExceptionHandler.php
@@ -133,10 +133,6 @@ class ExceptionHandler
             $info = $e->getBuildPHPInfo();
             self::logError('', output_log: defined('DEBUG_MODE'));
             self::logError('Builder function: ' . ConsoleColor::yellow($info['builder_function']), output_log: defined('DEBUG_MODE'));
-            if (self::$builder) {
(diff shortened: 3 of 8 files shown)


src/SPC/builder/BuilderProvider.php
```diff
@@ -29,6 +30,10 @@ public static function makeBuilderByInput(InputInterface $input): BuilderBase
             'BSD' => new BSDBuilder($input->getOptions()),
             default => throw new WrongUsageException('Current OS "' . PHP_OS_FAMILY . '" is not supported yet'),
         };
+
+        // bind the builder to ExceptionHandler
+        ExceptionHandler::$bind_builder = self::$builder;
```
henderkes: I don't think we should be using static state for this. A method with a private member `$builders` would be better.

src/SPC/command/BuildPHPCommand.php
```diff
@@ -165,8 +165,9 @@ public function handle(): int
         }
         $this->printFormatInfo($this->getDefinedEnvs(), true);
         $this->printFormatInfo($indent_texts);
-        // bind extra info to SPCException
-        SPCException::bindBuildPHPExtraInfo($indent_texts);
+
+        // bind extra info to exception handler
+        ExceptionHandler::$bind_build_php_extra_info = $indent_texts;
```
henderkes: likewise

src/SPC/exception/ExceptionHandler.php
```diff
@@ -26,21 +27,27 @@ class ExceptionHandler
...
             InterruptException::class => "⚠ Build interrupted by user: {$e->getMessage()}",
-            PatchException::class => "Patch apply failed: {$e->getMessage()}",
-            SPCInternalException::class => "SPC internal error: {$e->getMessage()}",
-            ValidationException::class => "Validation failed: {$e->getMessage()}",
+            PatchException::class => "✗ Patch apply failed: {$e->getMessage()}",
+            SPCInternalException::class => "✗ SPC internal error: {$e->getMessage()}",
+            ValidationException::class => "⚠ Validation failed: {$e->getMessage()}",
             WrongUsageException::class => $e->getMessage(),
-            default => "Unknown SPC exception {$class}: {$e->getMessage()}",
+            default => "✗Unknown SPC exception {$class}: {$e->getMessage()}",
```
henderkes: missing a space

henderkes (changes requested):

src/SPC/exception/ExceptionHandler.php
henderkes:
```suggestion
            default => "✗ Unknown SPC exception {$class}: {$e->getMessage()}",
```

src/SPC/builder/BuilderProvider.php
```diff
@@ -32,7 +32,7 @@ public static function makeBuilderByInput(InputInterface $input): BuilderBase
         };
 
         // bind the builder to ExceptionHandler
-        ExceptionHandler::$bind_builder = self::$builder;
+        ExceptionHandler::setBindBuilder(self::$builder);
```
henderkes:
```suggestion
        ExceptionHandler::bindBuilder(self::$builder);
```

src/SPC/command/BuildPHPCommand.php
```diff
@@ -167,7 +167,7 @@ public function handle(): int
         $this->printFormatInfo($indent_texts);
 
         // bind extra info to exception handler
-        ExceptionHandler::$bind_build_php_extra_info = $indent_texts;
+        ExceptionHandler::setBindBuildPhpExtraInfo($indent_texts);
```
henderkes:
```suggestion
        ExceptionHandler::bindBuildPhpExtraInfo($indent_texts);
```

src/SPC/exception/ExceptionHandler.php
```diff
@@ -151,6 +164,16 @@ public static function handleDefaultException(\Throwable $e): void
         self::logError('⚠ Please report this exception to: https://github.com/crazywhalecc/static-php-cli/issues');
     }
 
+    public static function setBindBuilder(?BuilderBase $bind_builder): void
```
henderkes:
```suggestion
    public static function bindBuilder(?BuilderBase $bind_builder): void
```

src/SPC/exception/ExceptionHandler.php
```diff
@@ -151,6 +164,16 @@ public static function handleDefaultException(\Throwable $e): void
...
+    {
+        self::$bind_builder = $bind_builder;
+    }
+
+    public static function setBindBuildPhpExtraInfo(array $bind_build_php_extra_info): void
```
henderkes:
```suggestion
    public static function bindBuildPhpExtraInfo(array $build_php_extra_info): void
```

src/SPC/exception/ExceptionHandler.php
```diff
@@ -151,6 +164,16 @@ public static function handleDefaultException(\Throwable $e): void
...
+    {
+        self::$bind_build_php_extra_info = $bind_build_php_extra_info;
```
henderkes:
```suggestion
        self::$build_php_extra_info = $build_php_extra_info;
```

src/SPC/exception/ExceptionHandler.php
```diff
@@ -28,10 +32,10 @@ class ExceptionHandler
     ];
 
     /** @var null|BuilderBase Builder binding */
-    public static ?BuilderBase $bind_builder = null;
+    private static ?BuilderBase $bind_builder = null;
```
henderkes:
```suggestion
    private static ?BuilderBase $builder = null;
```

src/SPC/exception/ExceptionHandler.php
```diff
@@ -28,10 +32,10 @@ class ExceptionHandler
...
 
     /** @var array<string, mixed> Build PHP extra info binding */
-    public static array $bind_build_php_extra_info = [];
+    private static array $bind_build_php_extra_info = [];
```
henderkes:
```suggestion
    private static array $build_php_extra_info = [];
```

henderkes (changes requested):
`bind` is a verb like `set` is.

henderkes:
Could you trigger a ./configure command failure instead please?

I also think that a lot of the currently shown options are useless.
```yaml
Builder options:
    with-libs:                 ""
    build-shared:              ""
 --- useless unless crashing during specific sapi build
    build-micro:               true
    build-cli:                 true
    build-fpm:                 true
    build-embed:               false
    build-frankenphp:          false
    build-all:                 false
--- end useless
    no-strip:                  false # useless because stripping is safe
    disable-opcache-jit:       false
    with-config-file-path:     /usr/local/etc/php # useless
    with-config-file-scan-dir: /usr/local/etc/php/conf.d # useless
    with-hardcoded-ini:        [] # useless
    with-micro-fake-cli:       false
    with-suggested-libs:       false
    with-suggested-exts:       false
    with-added-patch:          []
    without-micro-ext-test:    false
    with-upx-pack:             false
    with-micro-logo:           null
    enable-micro-win32:        false
    debug:                     true # useless
    no-motd:                   false # useless
    preserve-log:              false
    with-clean:                false # difference to rebuild?
    rebuild:                   false
    enable-zts:                true
    help:                      false # useless
    silent:                    false # useless
    quiet:                     false # useless
    verbose:                   false # useless
    version:                   false # we already see the php version earlier in the log, also, why is this a thing in the builder? it only concerns the download, no?
    ansi:                      null # what is this?
    no-interaction:            false # useless
 ```

**Comment by henderkes** [comment]
The options shown right above contain everything we need I believe:
```yaml
Build PHP extra info:
    Build OS:               Linux (x86_64)
    Build Target:           false
    Build Toolchain:        SPC\toolchain\MuslToolchain
    Build SAPI:             cli, micro, fpm
    Static Extensions (30): bcmath,bz2,calendar,ctype,curl,dom,exif,fileinfo,filter,ftp,iconv,xml,mbstring,mbregex,mysqlnd,zlib,openssl,pdo,pdo_mysql,sqlite3,pdo_sqlite,phar,session,simplexml,soap,sockets,tokenizer,xmlwriter,xmlreader,zip
    Shared Extensions (0):  ""
    Libraries (13):         pkg-config,bzip2,zlib,openssl,libiconv,libxml2,curl,onig,sqlite,libzip
    Strip Binaries:         yes
    Enable ZTS:             yes
    Config File Path:       /usr/local/etc/php
    PHP Version:            unknown

Builder function: proveExts
```

Just need to add if opcache-jit was disabled or not.

crazywhalecc:
> I also think that a lot of the currently shown options are useless.

This just prints all symfony command arguments directly. I don't want to show all of them either but it's hard to pick and remove the default values only. Currently it will output to console only if we are using `--debug` option, and without it it will not be shown and only be logged in file. Honestly I almost never use `--debug` after updating the exception enhancement, because it is more convenient to view the log directly and console is clean with key error msg only.

Alternatively we could replace them to `implode(' ', $_SERVER['argv'])`.

crazywhalecc:
I noticed the `spc.output.log` contains `argv` debug log at the beginning. So removing it.

henderkes:
Looks like the git master doesn't contain some of the header files. They are present in the php 8.X branches.

henderkes:
https://downloads.php.net/~edorian/ beta 1 was released there. It also doesn't contain the header files.

henderkes:
https://github.com/php/php-src/issues/19483

crazywhalecc:
> https://github.com/php/php-src/issues/19483

This is not related to this PR. I'll track them later in #842 . This could be merge now if nothing wrong.

henderkes (approved):

PR #856, opened 2025-08-16
Fix WindowsCmd execution, use popen instead of proc_open

### LINKED ISSUE #854: Windows weekly build unexpectedly timed out
https://github.com/static-php/static-php-cli-hosted/actions/runs/16926150672/job/47983009246

crazywhalecc:
## What does this PR do?

Fix #854 .

The `proc_open` does not work with windows cmd very well. The `popen` could avoid deadlock and infinite loop. For other OS, keeping `proc_open` is better.

## Checklist before merging

  - [X] `composer cs-fix`
  - [X] `composer analyse`
  - [X] `composer test`

### DIFF
--- a/src/SPC/util/shell/WindowsCmd.php
+++ b/src/SPC/util/shell/WindowsCmd.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace SPC\util\shell;
 
+use SPC\exception\ExecutionException;
 use SPC\exception\SPCInternalException;
 use ZM\Logger\ConsoleColor;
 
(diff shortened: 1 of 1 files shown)


henderkes:
Have you tried setting 
```php
proc_open(..., options: ['bypass_shell' => true]);
```
?

crazywhalecc:
`proc_open` with `bypass_shell=true` does not accept `cd /d "\dir" && $cmd` like this. And if we wrap `cmd /c` it performs exactly the same.

henderkes:
Oh, indeed. I will look at this tomorrow. In Udon for the weekend.

crazywhalecc:
@henderkes Well, I should noticed that I can use search engine to solve the problem.

https://github.com/php/php-src/pull/5777

PR #843, opened 2025-08-01
Add msgpack option to Redis extension builder

pecherskiy-v:
## Checklist before merging

> If your PR involves the changes mentioned below and completed the action, please tick the corresponding option.
> If a modification is not involved, please skip it directly.

- If you modified `*.php` or `*.json`, run them locally to ensure your changes are valid:
  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`
  - [ ] `bin/spc dev:sort-config`
- If it's an extension or dependency update, please ensure the following:
  - [ ] Add your test combination to `src/globals/test-extensions.php`.
  - [ ] If adding new or fixing bugs, add commit message containing `extension test` or `test extensions` to trigger full test suite.

PR #857, opened 2025-08-19
move -lstdc++ to the end of the libraries

henderkes:
## What does this PR do?
https://github.com/static-php/spc-packages/actions/runs/17057743582/job/48358419331#step:10:54

libjxl requests -lstdc++ in pkgconfig, so it's not added at the end again. icu comes later and also needs -lstdc++ though, so that leads to undefined references during linking

closes https://github.com/crazywhalecc/static-php-cli/pull/843


henderkes:
closes https://github.com/crazywhalecc/static-php-cli/pull/843

crazywhalecc (approved):

PR #864, opened 2025-08-25
swoole hooks can be compiled if pgsql/sqlite are not compiled in statically

henderkes:
## What does this PR do?
enable swoole features as possible

## Checklist before merging

- Features working
  - [x] pgsql
  - [x] mysql
  - [x] sqlite
  - [x] threads
  - [x] liburing
  - [x] odbc (https://github.com/swoole/swoole-src/pull/5847 to get rid of patch)

### DIFF
--- a/config/ext.json
+++ b/config/ext.json
@@ -843,12 +843,10 @@
             "zlib"
         ],
         "lib-suggests": [
-            "zstd",
-            "unixodbc"
+            "zstd"
         ],
         "lib-suggests-linux": [
             "zstd",
-            "unixodbc",
             "liburing"
         ],
         "ext-depends": [
@@ -859,7 +857,8 @@
             "sockets",
             "swoole-hook-pgsql",
             "swoole-hook-mysql",
-            "swoole-hook-sqlite"
+            "swoole-hook-sqlite",
+            "swoole-hook-odbc"
         ]
     },
     "swoole-hook-mysql": {
@@ -907,6 +906,22 @@
             "sqlite3",
             "pdo",
             "swoole"
+        ]
+    },
+    "swoole-hook-odbc": {
+        "support": {
+            "Windows": "no",
+            "BSD": "wip"
+        },
+        "notes": true,
+        "type": "addon",
+        "arg-type": "none",
+        "ext-depends": [
+            "pdo",
+            "swoole"
+        ],
+        "lib-depends": [
+            "unixodbc"
         ]
     },
     "swow": {
--- a/config/ext.json
+++ b/config/ext.json
@@ -863,7 +863,8 @@
         "ext-depends": [
             "mysqlnd",
             "pdo",
-            "pdo_mysql"
+            "pdo_mysql",
+            "swoole"
         ],
         "ext-suggests": [
             "mysqli"
@@ -880,7 +881,8 @@
         "arg-type": "custom",
         "ext-depends": [
             "pgsql",
-            "pdo"
+            "pdo",
+            "swoole"
         ]
     },
     "swoole-hook-sqlite": {
@@ -893,7 +895,8 @@
         "arg-type": "custom",
         "ext-depends": [
             "sqlite3",
-            "pdo"
+            "pdo",
+            "swoole"
         ]
     },
     "swow": {
--- a/config/ext.json
+++ b/config/ext.json
@@ -842,11 +842,15 @@
             "nghttp2",
             "zlib"
         ],
+        "lib-suggests": [
+            "zstd"
+        ],
         "ext-depends": [
             "openssl",
             "curl"
         ],
         "ext-suggests": [
+            "sockets",
             "swoole-hook-pgsql",
             "swoole-hook-mysql",
             "swoole-hook-sqlite"
--- a/config/ext.json
+++ b/config/ext.json
@@ -843,7 +843,8 @@
             "zlib"
         ],
         "lib-suggests": [
-            "zstd"
+            "zstd",
+            "unixodbc"
         ],
         "ext-depends": [
             "openssl",
@@ -863,7 +864,7 @@
         },
         "notes": true,
         "type": "addon",
-        "arg-type": "custom",
+        "arg-type": "none",
         "ext-depends": [
             "mysqlnd",
             "pdo",
@@ -882,7 +883,7 @@
         },
         "notes": true,
         "type": "addon",
-        "arg-type": "custom",
+        "arg-type": "none",
         "ext-depends": [
             "pgsql",
             "pdo",
@@ -896,7 +897,7 @@
         },
         "notes": true,
         "type": "addon",
-        "arg-type": "custom",
+        "arg-type": "none",
         "ext-depends": [
             "sqlite3",
             "pdo",
--- a/config/ext.json
+++ b/config/ext.json
@@ -846,6 +846,11 @@
             "zstd",
             "unixodbc"
         ],
+        "lib-suggests-linux": [
+            "zstd",
+            "unixodbc",
+            "liburing"
+        ],
         "ext-depends": [
             "openssl",
             "curl"
--- a/src/SPC/builder/extension/swoole.php
+++ b/src/SPC/builder/extension/swoole.php
@@ -61,7 +61,7 @@ class swoole extends Extension
         }
         $arg .= $this->builder->getLib('nghttp2') ? (' --with-nghttp2-dir=' . BUILD_ROOT_PATH) : '';
         $arg .= $this->builder->getLib('zstd') ? ' --enable-zstd' : '';
-        $arg .= $this->builder->getLib('iouring') ? ' --enable-iouring' : '';
+        $arg .= $this->builder->getLib('liburing') ? ' --enable-iouring' : '';
         $arg .= $this->builder->getExt('sockets') ? ' --enable-sockets' : '';
 
         // additional feature: swoole-pgsql, it should depend on lib [postgresql], but it will lack of CFLAGS etc.
--- a/src/SPC/builder/extension/swoole.php
+++ b/src/SPC/builder/extension/swoole.php
@@ -61,11 +61,11 @@ class swoole extends Extension
 
         // additional feature: swoole-pgsql, it should depend on lib [postgresql], but it will lack of CFLAGS etc.
         // so this is a tricky way (enable ext [pgsql,pdo] to add postgresql hook and pdo_pgsql support)
-        $arg .= $this->builder->getExt('swoole-hook-pgsql') ? '' : ' --disable-swoole-pgsql';
+        $arg .= $this->builder->getExt('swoole-hook-pgsql') ? ' --enable-swoole-pgsql' : ' --disable-swoole-pgsql';
 
         // enable this feature , need remove pdo_sqlite
         // more info : https://wenda.swoole.com/detail/109023
-        $arg .= $this->builder->getExt('swoole-hook-sqlite') ? '' : ' --disable-swoole-sqlite';
+        $arg .= $this->builder->getExt('swoole-hook-sqlite') ? ' --enable-swoole-sqlite' : ' --disable-swoole-sqlite';
 
         // enable this feature , need stop pdo_*
         // $arg .= $this->builder->getLib('unixodbc') ? ' --with-swoole-odbc=unixODBC,'  : ' ';
--- a/src/SPC/builder/extension/swoole.php
+++ b/src/SPC/builder/extension/swoole.php
@@ -49,8 +49,7 @@ class swoole extends Extension
         // commonly-used feature: coroutine-time
         $arg .= ' --enable-swoole-coro-time --with-pic';
 
-        $arg .= $this->builder->getOption('enable-zts') && SPCTarget::getTargetOS() !== 'Darwin' ? ' --enable-swoole-thread' : ' --disable-swoole-thread';
-        $arg .= $this->builder->getOption('enable-zts') || SPCTarget::getTargetOS() === 'Darwin' ? ' --disable-thread-context' : ' --enable-thread-context';
+        $arg .= $this->builder->getOption('enable-zts') ? ' --enable-swoole-thread --disable-thread-context' : ' --disable-swoole-thread --enable-thread-context';
 
         // required feature: curl, openssl (but curl hook is buggy for php 8.0)
         $arg .= $this->builder->getPHPVersionID() >= 80100 ? ' --enable-swoole-curl' : ' --disable-swoole-curl';
--- a/src/SPC/builder/extension/swoole.php
+++ b/src/SPC/builder/extension/swoole.php
@@ -69,8 +69,10 @@ class swoole extends Extension
         $arg .= $this->builder->getExt('swoole-hook-mysql') ? ' --enable-mysqlnd' : ' --disable-mysqlnd';
         $arg .= $this->builder->getExt('swoole-hook-sqlite') ? ' --enable-swoole-sqlite' : ' --disable-swoole-sqlite';
 
-        $config = (new SPCConfigUtil($this->builder, ['libs_only_deps' => true]))->config([], ['unixodbc']);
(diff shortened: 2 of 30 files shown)


jingjingxyk:
new feature:
 - [`--enable-swoole-stdext'](https://mp.weixin.qq.com/s/RmdKC5TLREwjcJFc-upZLQ)
- [stdext enhance ](https://mp.weixin.qq.com/s/yFuxYdNMqxI1VuZg0J07Wg)

henderkes:
> new feature:
> 
> * [`--enable-swoole-stdext'](https://mp.weixin.qq.com/s/RmdKC5TLREwjcJFc-upZLQ)
> * [stdext enhance ](https://mp.weixin.qq.com/s/yFuxYdNMqxI1VuZg0J07Wg)

What is swoole-stdext?

Edit: Oh damn, I wished for that for a long time. Too bad I can't personally use swoole, we're making use of too many caddy features.

henderkes:
edit: it's v6.1+ only, the stable release is still v6.0.2, so that will need a feature request that we can do later

henderkes:
Looks like the swoole fails are not introduced by this pr (they still happen when I disable the new --enable options), but by Swoole src not checking for extended posix support before using it? Hard to track down without a mac.

crazywhalecc:
I'll check it tomorrow evening.

src/SPC/builder/extension/swoole.php
```diff
@@ -43,32 +43,40 @@ public function getExtVersion(): ?string
...
-        $arg .= ' --enable-swoole-coro-time --disable-thread-context';
+        // commonly-used feature: coroutine-time
+        $arg .= ' --enable-swoole-coro-time --with-pic';
+
+        $arg .= $this->builder->getOption('enable-zts') ? ' --enable-swoole-thread --disable-thread-context' : ' --disable-swoole-thread --enable-thread-context';
 
         // required feature: curl, openssl (but curl hook is buggy for php 8.0)
         $arg .= $this->builder->getPHPVersionID() >= 80100 ? ' --enable-swoole-curl' : ' --disable-swoole-curl';
-        $arg .= ' --enable-openssl';
+        $arg .= ' --enable-openssl'; // we depend on openssl so it's always enabled
```
henderkes:
```suggestion
        $arg .= ' --enable-openssl'; 
```

didn't see the comment above, so this one is superfluous

src/SPC/command/BuildPHPCommand.php
```diff
@@ -252,7 +252,7 @@ public function handle(): int
                 $path = FileSystem::convertPath("{$build_root_path}/modules/{$ext}.so");
                 if (file_exists(BUILD_MODULES_PATH . "/{$ext}.so")) {
                     logger()->info("Shared extension [{$ext}] path{$fixed}: {$path}");
-                } else {
+                } elseif (!str_contains($ext, 'swoole-hook')) {
```
henderkes:
```suggestion
                } else {
```

henderkes:
The odbc change needs fixing in swoole-src, I've created a PR there - if they don't merge it, we'll have to apply a patch on swoole src extraction.

config/ext.json
```diff
@@ -890,10 +901,27 @@
...
+            "pdo",
+            "swoole"
+        ]
+    },
+    "swoole-hook-odbc": {
+        "support": {
+            "Windows": "no",
+            "BSD": "wip"
+        },
+        "notes": true,
```
crazywhalecc, line 916: Looks like this pr does not contain swoole-hook-odbc notes.

crazywhalecc:
Looks good to me now, I tested locally and building successfully. Will review later.

henderkes:
Yes, it builds on all compilers and OS. I am trying to upstream the patches to swoole, but added them as patches for now.

src/SPC/builder/extension/swoole.php
```diff
@@ -43,32 +45,40 @@ public function getExtVersion(): ?string
...
-        $arg = '--enable-swoole';
+        $arg = '--enable-swoole' . ($shared ? '=shared' : '');
 
-        // commonly-used feature: coroutine-time, disable-thread-context
-        $arg .= ' --enable-swoole-coro-time --disable-thread-context';
+        // commonly used feature: coroutine-time
+        $arg .= ' --enable-swoole-coro-time --with-pic';
 
-        // required feature: curl, openssl (but curl hook is buggy for php 8.0)
+        $arg .= $this->builder->getOption('enable-zts') ? ' --enable-swoole-thread --disable-thread-context' : ' --disable-swoole-thread --enable-thread-context';
```
crazywhalecc, line 53: Disabling thread need enabling thread context?

src/SPC/builder/extension/swoole.php
```diff
@@ -43,32 +45,40 @@ public function getExtVersion(): ?string
...
-        // more info : https://wenda.swoole.com/detail/109023
-        $arg .= $this->builder->getExt('swoole-hook-sqlite') ? '' : ' --disable-swoole-sqlite';
+        if ($this->builder->getExt('swoole-hook-odbc')) {
+            $config = (new SPCConfigUtil($this->builder, ['libs_only_deps' => true]))->config([], ['unixodbc']);
+            $arg .= ' --with-swoole-odbc=unixODBC,' . BUILD_ROOT_PATH . ' SWOOLE_ODBC_LIBS="' . $config['libs'] . '"';
+        }
+
+        if (SPCTarget::getTargetOS() === 'Darwin') {
+            $arg .= ' ac_cv_lib_pthread_pthread_barrier_init=no';
+        }
```
crazywhalecc, line 80: I haven't figure it out but I suggest adding comments and links here.

src/SPC/store/SourcePatcher.php
```diff
@@ -269,7 +269,10 @@ public static function patchSwoole(): bool
         if ($version === '5.1.3') {
             self::patchFile('spc_fix_swoole_50513.patch', SOURCE_PATH . '/php-src/ext/swoole');
         }
+        // remove when https://github.com/swoole/swoole-src/pull/5848 is merged
         self::patchFile('swoole_fix_date_time.patch', SOURCE_PATH . '/php-src/ext/swoole');
+        // remove when https://github.com/swoole/swoole-src/pull/5847 is merged
+        self::patchFile('swoole_fix_odbclibs.patch', SOURCE_PATH . '/php-src/ext/swoole');
```
crazywhalecc: Is this patch available for latest swoole version only?

src/SPC/command/BuildPHPCommand.php
```diff
@@ -252,7 +252,7 @@ public function handle(): int
                 $path = FileSystem::convertPath("{$build_root_path}/modules/{$ext}.so");
                 if (file_exists(BUILD_MODULES_PATH . "/{$ext}.so")) {
                     logger()->info("Shared extension [{$ext}] path{$fixed}: {$path}");
-                } else {
+                } elseif (Config::getExt($ext, 'type', 'ext') !== 'addon') {
```
crazywhalecc: `'ext'` seems redundant here

src/SPC/builder/linux/library/liburing.php
```diff
@@ -0,0 +1,59 @@
...
+                '--enable-pic',
+            )
+            ->addConfigureArgs(
+                $use_libc ? '--use-libc' : '',
+            )
+            ->configure()
+            ->make('library', with_clean: false)
+            ->exec("rm -rf {$this->getLibDir()}/liburing*.so*");
+
+        $this->patchPkgconfPrefix(['liburing.pc', 'liburing-ffi.pc']);
```
crazywhalecc: We don't need to explictly define when we defined in config.

src/SPC/util/SPCConfigUtil.php
```diff
@@ -52,6 +52,15 @@ public function __construct(?BuilderBase $builder = null, array $options = [])
...
     {
+        $extra_exts = [];
+        foreach ($extensions as $ext) {
+            $extra_exts = array_merge($extra_exts, Config::getExt($ext, 'ext-suggests', []));
+        }
+        foreach ($extra_exts as $ext) {
+            if ($this->builder?->getExt($ext) && !in_array($ext, $extensions)) {
+                $extensions[] = $ext;
+            }
+        }
```
crazywhalecc, line 63: `DependencyUtil::getExtsAndLibs` will automatically sort the suggested extensions. And is there any reason we need to add suggested extensions here?

src/SPC/builder/extension/swoole.php
henderkes: Yes. ZTS builds can --enable-swoole-thread, but not --enable-thread-context while it's vice versa for NTS. I haven't looked if that actually makes sense, but it's how Swooles config.m4 is.

src/SPC/builder/unix/UnixBuilderBase.php
```diff
@@ -280,10 +281,19 @@ protected function buildFrankenphp(): void
...
 
         $config = (new SPCConfigUtil($this))->config($this->ext_list, $this->lib_list);
+        $cflags = "{$this->arch_c_flags} {$config['cflags']} " . getenv('SPC_CMD_VAR_PHP_MAKE_EXTRA_CFLAGS');
+        $libs = $config['libs'];
+        $libs .= PHP_OS_FAMILY === 'Linux' ? ' -lrt' : '';
+        if ((str_contains((string) getenv('SPC_DEFAULT_C_FLAGS'), '-fprofile') ||
+                str_contains((string) getenv('SPC_CMD_VAR_PHP_MAKE_EXTRA_CFLAGS'), '-fprofile')) &&
+            ToolchainManager::getToolchainClass() === GccNativeToolchain::class) {
+            $cflags .= ' -Wno-error=missing-profile';
+            $libs .= ' -lgcov';
```
crazywhalecc, line 292: Same as swoole's patches: comments.

config/ext.json
henderkes: I just copied it from the other hooks. I'll have to see where they are added.

src/SPC/builder/extension/swoole.php
henderkes: If you could figure out why this unexpectedly succeeds on macos, we should submit another PR to swoole. It might be that the linker check succeeds to stubs even if the headers don't contain the symbol.

crazywhalecc: Okay I'll look into it later.

src/SPC/util/SPCConfigUtil.php
henderkes: Yes! Example: swoole suggests ext swoole-hook-pgsql. Swoole-hook-pgsql requires pdo and pgsql libs. If we get the config from swoole, it doesn't directly depend on pgsql, but when pgsql hook is enabled, it does.

It's the same way where libraries become dependencies of libraries that suggest them, when they are enabled.

src/SPC/builder/extension/swoole.php
crazywhalecc: Got it: https://github.com/swoole/swoole-src/blob/8057ba9b4e1a972e892149552467a27898e6e22b/ext-src/php_swoole_private.h#L135-L137

src/SPC/builder/unix/UnixBuilderBase.php
henderkes: Will add, you're right. This is just a workaround for Go's terrible gcc driver.

src/SPC/util/SPCConfigUtil.php
crazywhalecc: But `DependencyUtil::getExtsAndLibs` will automatically add suggested extension's dependencies if defined. The DependencyUtil's `platExtToLibs` also includes ext-suggests.

I also tested the difference with and without this code, and for `bin/spc spc-config swoole,swoole-hook-pgsql`, the output is exactly the same.

src/SPC/store/SourcePatcher.php
henderkes: I have not looked if it succeeds with swoole 5.x, but it works on 6.x.

src/SPC/util/SPCConfigUtil.php
henderkes: >bin/spc spc-config swoole,swoole-hook-pgsql

Extension::buildShared calls SpcConfigUtil::config([$ext->name]).

So it only passes `swoole` but `swoole-hook-pgsql` is enabled in the builder. This code turns ['swoole'] into ['swoole', 'swoole-hook-pgsql'] before DependencyUtil.

crazywhalecc: You mean: `bin/spc build swoole-hook-pgsql --build-shared=swoole` like this?

henderkes: Yes.

I made `bin/spc build bcmath --build-shared=swoole,swoole-hook-pgsql` work too, by the way, because it's nicer to have them next to each other.

crazywhalecc: Ah, that makes sense. My initial thought was to provide the full list of extensions in buildShared, but that seems more complicated than this.

henderkes: That would not work because it would pull in definitions for symbols that should not be available (looking at you, postgresql, defining libc methods...)

src/SPC/store/SourcePatcher.php
henderkes: I've added >6.0 < 6.1 version constraints. Let's hope they merge the fixes in.

henderkes: Btw 5.x does not even work with php 8.4.

henderkes:
@crazywhalecc good for merge?

crazywhalecc:
Haven't tested on linux yet, after my day job is finished I'll look into it. I've been a bit busy at work lately.

crazywhalecc:
The macOS bug reason: 

- macOS doesn't implement `pthread_barrier_*` function in pthread.
- The pg_pthread has `pthread_barrier` port.
- When configuring `pthread_barrier_init`, this includes pg libs for conftest, and it will pass the check.
- Making does not contain the pg libs, that is, the posix thread barrier obviously not implemented and got failed build.

The root cause is that the configure built in spc will bring other libraries into a separate `AC_CHECK_LIB` additional library, which leads to this problem.

crazywhalecc:
And for linux liburing, I dont' see anything wrong, but I recommend triggering full extension tests and all versions of bulk build tests with swoole's hooks at the end.

henderkes:
Ah, it's pgsql again...

henderkes:
> And for linux liburing, I dont' see anything wrong, but I recommend triggering full extension tests and all versions of bulk build tests with swoole's hooks at the end.

https://github.com/static-php/spc-packages/actions/runs/17260452046

Already tested!

henderkes:
I think we need to disable opcache jit with zts builds for php < 8.4 on alpine. It was fixed, but apparently not backported.

crazywhalecc (approved):
I added missing notes and updated the cli generator, good to me now. But I don't think we should add pkg-config things in this PR. 

Actually I also prefer hosting pkg-config musl-static binaries in dl.static-php.dev without building in doctor. The pkg-config build actions would be in static-php-cli-hosted.

henderkes:
Yes, you can change that in the other branch. I don't have access to the dl.static-php so I can't upload a prebuilt version of it.

henderkes:
Oh, as a side note, what is supposed to happen if swoole-pgsql and pdo_pgsql extension are both loaded shared? Does one need to initialise first to get it working correctly?

crazywhalecc:
> Oh, as a side note, what is supposed to happen if swoole-pgsql and pdo_pgsql extension are both loaded shared? Does one need to initialise first to get it working correctly?

I haven't tested shared build for swoole extension yet, but theoretically it should work together.

henderkes:
I've tested it and it works fine, but only the general pdo_pgsql. I have never worked with swoole.

PR #869, opened 2025-08-27
turn pkg-config into a package

henderkes:
## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`

### DIFF
--- a/src/SPC/toolchain/ToolchainManager.php
+++ b/src/SPC/toolchain/ToolchainManager.php
@@ -7,6 +7,7 @@ namespace SPC\toolchain;
 use SPC\builder\linux\SystemUtil;
 use SPC\exception\WrongUsageException;
 use SPC\util\GlobalEnvManager;
+use SPC\util\PkgConfigUtil;
 use SPC\util\SPCTarget;
 
 class ToolchainManager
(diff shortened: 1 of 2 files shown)


crazywhalecc:
Building in pkg is not recommended if not necessary. Generally, the build things should be library type.

Because we already support building from source and releasing weekly, there's no need to modify existing code. Simply modify the dependencies and install the package in Doctor with one click.

crazywhalecc (approved):

henderkes:
We're keeping it in lib.json and in unix/library/pkgconfig too? What's the point in erroring out if it's not found then?

crazywhalecc:
> We're keeping it in lib.json and in unix/library/pkgconfig too?

Yeah, since we still need a location to build own pkg-config, I think the best option is to keep the current state. But for backward compatibility and to achieve this (avoid building pkg-config every time), downloading our existing pre-built content is decent.

> What's the point in erroring out if it's not found then?

You mean auto-fix in afterInit?

src/SPC/toolchain/ToolchainManager.php
```diff
@@ -56,6 +57,15 @@ public static function afterInitToolchain(): void
         if (SPCTarget::getLibc() === 'glibc' && SystemUtil::isMuslDist()) {
             throw new WrongUsageException('You are linking against glibc dynamically, which is only supported on glibc distros.');
         }
+
+        // init pkg-config for unix
+        if (is_unix()) {
+            if (($found = PkgConfigUtil::findPkgConfig()) === null) {
+                throw new WrongUsageException('Cannot find pkg-config executable. Please run `doctor` to fix this.');
+            }
+            GlobalEnvManager::putenv("PKG_CONFIG={$found}");
```
henderkes, line 66: Yes here, when it's still a library it could be automatically built if someone forgot to build it with the doctor fix.

crazywhalecc: I still don't recommend building libraries in pkg; they should simply involve downloading and installing whenever possible.

As for the "auto-repair if pkg-config not found" in afterInit in GlobalEnv, I don't think it should be downloaded here, at least for now. **People should run `bin/spc doctor` in new environments anyway**.

But I think we could package the `pkg-config` prebuilt binary into a single `spc` file, automatically extracting it to pkgroot/bin if it's not found there. And then do this in the `setup-runtime` method, since it's always required to be installed.

PR #871, opened 2025-08-29
build frankenphp and embed after shared extensions

henderkes:
make sure the linker doesn't drop libphp.a symbols that extensions need

## What does this PR do?

previously, if we built libphp.a, the linker cleans up dead symbols when creating the executable. This is undesirable, because when you try to load shared extensions with frankenphp, it will fail to load them with undefined references to `zend_ini_display_cb...` and other functions.

Now, instead of marking all symbols as dynamic, which *massively* bloats the executable size, we loop through all shared extensions, figure out what symbols they need, match that against what libphp.a provides and then mark those as export-dynamic.

Keep in mind that this *does* still increase the executable size a little! But it only takes effect when people build shared extensions.

### DIFF
--- a/src/SPC/builder/traits/UnixSystemUtilTrait.php
+++ b/src/SPC/builder/traits/UnixSystemUtilTrait.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace SPC\builder\traits;
 
 use SPC\exception\ExecutionException;
+use SPC\exception\SPCInternalException;
 use SPC\exception\WrongUsageException;
 use SPC\toolchain\ToolchainManager;
 use SPC\toolchain\ZigToolchain;
@@ -66,7 +67,10 @@ trait UnixSystemUtilTrait
     {
         $symbol_file = "{$lib_file}.dynsym";
         if (!is_file($symbol_file)) {
-            return null;
+            self::exportDynamicSymbols($lib_file);
+        }
+        if (!is_file($symbol_file)) {
+            throw new SPCInternalException("The symbol file {$symbol_file} does not exist, please check if nm command is available.");
         }
         // https://github.com/ziglang/zig/issues/24662
         if (ToolchainManager::getToolchainClass() === ZigToolchain::class) {
--- a/src/SPC/builder/traits/UnixSystemUtilTrait.php
+++ b/src/SPC/builder/traits/UnixSystemUtilTrait.php
@@ -4,15 +4,88 @@ declare(strict_types=1);
 
 namespace SPC\builder\traits;
 
-/**
- * Unix 系统的工具函数 Trait，适用于 Linux、macOS
- */
+use SPC\exception\ExecutionException;
+use SPC\exception\WrongUsageException;
+use SPC\toolchain\ToolchainManager;
+use SPC\toolchain\ZigToolchain;
+use SPC\util\SPCTarget;
+
 trait UnixSystemUtilTrait
 {
     /**
-     * @param  string      $name  命令名称
-     * @param  array       $paths 寻找的目标路径（如果不传入，则使用环境变量 PATH）
-     * @return null|string 找到了返回命令路径，找不到返回 null
+     * Export static library dynamic symbols to a .dynsym file.
+     * It will export to "/path/to/libxxx.a.dynsym".
+     *
+     * @param string $lib_file Static library file path (e.g. /path/to/libxxx.a)
+     */
+    public static function exportDynamicSymbols(string $lib_file): void
+    {
+        // check
+        if (!is_file($lib_file)) {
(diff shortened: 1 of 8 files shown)


henderkes:
I don't know how to get macos to work :(

crazywhalecc:
Do we have any other scenario of exporting dynamic symbols?

henderkes:
I don't think so. Only for things that use libphp.a. The php code itself (libphp.so, cli, fpm, likely micro) uses all necessary symbols so they aren't stripped.

crazywhalecc:
I can't think of any other uses for this, but I think it might be clearer to separate it into "exporting the archive's symbol list to a file" and "getting dynamic symbol parameters from the symbol file." Also, I don't see the point of caching this arguments, as we only need to read the file at most twice.

henderkes:
I had considered splitting it up before, but found that I can't think of a situation where it makes sense to use for anything but libphp.a. We don't want to export symbols from anything but core php code - extensions will pull in the library symbols they need themselves.

crazywhalecc:
How about:

```php
// after building embed
SystemUtil::exportDynamicSymbols(BUILD_LIB_PATH . '/libphp.a');

// use symbols in sanity check, frankenphp builds, etc.
$dyn_symbols = SystemUtil::getDynamicExportedSymbols(BUILD_LIB_PATH . '/libphp.a');

$exportDynamicSymbols = function (string $lib_file) {
    // check
    if (!is_file($lib_file)) {
        throw new WrongUsageException("The lib archive file {$lib_file} does not exist, please build it first.");
    }
    // shell out
    $cmd = 'nm -g --defined-only -P ' . escapeshellarg($lib_file) . ' | grep -vE ".o]?:$" | grep -vE "^@?$" | awk -F "[\t| ]" "{print $1}"';
    $result = shell()->execWithResult($cmd);
    if ($result[0] !== 0) {
        throw new ExecutionException($cmd, 'Failed to get defined symbols from ' . $lib_file);
    }
    $defined = array_unique($result[1]);
    sort($defined);
    // export
    if (SPCTarget::getTargetOS() === 'Linux') {
        file_put_contents("{$lib_file}.dynsym", "{\n" . implode(";\n", array_map(fn ($x) => "  {$x}", $defined)) . ";\n};\n");
    } else {
        file_put_contents("{$lib_file}.dynsym", implode("\n", $defined) . "\n");
    }
};

$getDynamicExportedSymbols = function (string $lib_file): ?string {
    $symbol_file = "{$lib_file}.dynsym";
    if (!is_file($symbol_file)) {
        return null;
    }
    $argument = "-Wl,--dynamic-list={$symbol_file}";
    if (ToolchainManager::getToolchainClass() === ZigToolchain::class) {
        $argument = '-Wl,--export-dynamic'; // https://github.com/ziglang/zig/issues/24662
    }
    if (SPCTarget::getTargetOS() !== 'Linux') {
        $argument = "-Wl,-exported_symbols_list,{$symbol_file}";
    }
    return $argument;
};
```

henderkes:
I think the idea is good, but 

```php
    $cmd = 'nm -g --defined-only -P ' . escapeshellarg($lib_file) . ' | grep -vE ".o]?:$" | grep -vE "^@?$" | awk -F "[\t| ]" "{print $1}"';
 ```
 
is overkill. Keep some good old php code in there to keep it debuggable.

**Review by crazywhalecc — APPROVED** [review]

File: src/SPC/builder/traits/UnixSystemUtilTrait.php [hunk 1]
```diff
@@ -4,15 +4,88 @@
...
+        // https://github.com/ziglang/zig/issues/24662
+        if (ToolchainManager::getToolchainClass() === ZigToolchain::class) {
+            return '-Wl,--export-dynamic';
+        }
+        // macOS
+        if (SPCTarget::getTargetOS() !== 'Linux') {
+            return "-Wl,-exported_symbols_list,{$symbol_file}";
+        }
+        return "-Wl,--dynamic-list={$symbol_file}";
+    }
```
henderkes: I don't like that this requires the file to exist to get the argument, when the logic to create that file must be called earlier. If we forget it, it will be stupid to track this down. Either it should be one method like I did before, or one method should be private and called by the public one we use in code.

crazywhalecc: I don't see any problem with creating the file here, since we have to create a file to handle. Separating them makes for more reasonable behavior, but yes, it can still call it internally in the getter if the file doesn't exist.

However, rather than reverting to the previous version, I prefer to report an error if the symbol isn't retrieved when it should be, rather than returning null. This way, we'll know immediately what went wrong, eliminating the debugging issues you mentioned.

PR #878, opened 2025-09-01
Fix unixodbc config searching path for macOS and general linux

crazywhalecc:
## What does this PR do?

Previously we use `--prefix=` to build unixodbc, which will search `odbc.ini` in `$prefix/etc/odbc.ini`, that is, `/etc/odbc.ini`. 

For different linux distros and different ways of installing the msodbcsql driver, we may need to set different values ​​or variables, but here I am mainly trying to fix the default installation and use of macOS.

## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -9,6 +9,7 @@ parameters:
     - '#Attribute class JetBrains\\PhpStorm\\ArrayShape does not exist#'
     - '#Function Swoole\\Coroutine\\run not found.#'
     - '#Static call to instance method ZM\\Logger\\ConsoleColor#'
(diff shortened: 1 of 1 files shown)


henderkes:
Why do x86_64 and aarch64 have different paths?

crazywhalecc:
The Apple Silicon mac's homebrew is under `/opt/homebrew`, and intel mac is under `/usr/local`. And the mssql driver is using homebrew tap to install it.

PR #882, opened 2025-09-04
Fix docker doctor fix pkgroot volume overwrite bug

### LINKED ISSUE #881: actions to linux-x86_64-glibc
Image

crazywhalecc:
## What does this PR do?

Fix #881 

## Checklist before merging

  - [X] `composer cs-fix`
  - [X] `composer analyse`
  - [X] `composer test`

### DIFF
--- a/bin/spc-alpine-docker
+++ b/bin/spc-alpine-docker
@@ -3,7 +3,7 @@
 set -e
 
 # This file is using docker to run commands
-SPC_DOCKER_VERSION=v5
+SPC_DOCKER_VERSION=v6
 
 # Detect docker can run
 if ! which docker >/dev/null; then
@@ -122,6 +122,9 @@ ADD ./src /app/src
 COPY ./composer.* /app/
 ADD ./bin /app/bin
 RUN composer install --no-dev
+ADD ./config /app/config
+RUN PKG_ROOT_PATH=/app/pkgroot-private bin/spc doctor --auto-fix
+RUN PKG_ROOT_PATH=/app/pkgroot-private bin/spc install-pkg upx
 EOF
 fi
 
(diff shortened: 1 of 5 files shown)


henderkes:
Why does it not work with a mount? Why mount the directory at all then?

crazywhalecc:
> Why does it not work with a mount?

If someone running `bin/spc-alpine-docker install-pkg XXX`, it will install to `/path/to/spc/pkgroot/` (for docker inside, it's `/app/pkgroot/`). This because of the mount `-v $(pwd)/pkgroot:/app/pkgroot`.

But the image build procedure `RUN bin/spc install-pkg` or `RUN bin/spc doctor` don't have mount point and it will install in `/app/pkgroot` only. After mounting from host, it will disappear.

This PR will fix the required packages to be installed inside image to avoid mounting overwrite. Although we can solve it by not mounting pkgroot, I think this is inconsistent with the buildroot and other directories.

> Why mount the directory at all then?

pkgroot is not part of docker, it is the same as buildroot and should be visible and modifiable in the host. pkg-config and upx are exceptions.

henderkes:
I understand the problem now, but I think the right way to combat it is to have the mount available for bin/spc doctor. The current solution is very confusing.

crazywhalecc:
Indeed. I just noticed that the go-xcaddy will also have this issue. Mounting when building seems to be appropriate.

crazywhalecc:
Seems `docker build` command does not accept mounting volume yet. How about installing to `pkgroot-private` during `docker build` and copying to `pkgroot` when `docker run`?

crazywhalecc:
I think splitting different OS and arch for `pkgroot` dir is necessary when using docker on different arch and OS.

henderkes:
That's a good idea because it would also mean that packages themselves would no longer need arch and os in their folder.

crazywhalecc:
Looks good to me now.

PR #885, opened 2025-09-04
switch getting frankenphp version to go mod (bypass rate limits)

henderkes:
## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`

### DIFF
--- a/src/SPC/builder/unix/UnixBuilderBase.php
+++ b/src/SPC/builder/unix/UnixBuilderBase.php
@@ -143,13 +143,11 @@ abstract class UnixBuilderBase extends BuilderBase
             if (getenv('SPC_CMD_VAR_PHP_EMBED_TYPE') === 'shared') {
                 if (PHP_OS_FAMILY === 'Darwin') {
                     $ext_path = 'DYLD_LIBRARY_PATH=' . BUILD_LIB_PATH . ':$DYLD_LIBRARY_PATH ';
-                }
-                else {
+                } else {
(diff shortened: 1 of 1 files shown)


henderkes:
I'm not sure why, but this step is failing nearly every time on at least one runner.

henderkes:
No idea what happened in the CI here, php cs fixer hallucinated?

crazywhalecc:
Sigh. This is why I like to keep downloading and building completely separate. It's also one of the things I don't like about Go: it always requires a network connection to GitHub.

I think it's time to change this approach to getting releases. Issues with accessing the GitHub API are the most common download issues we encounter.

crazywhalecc:
> No idea what happened in the CI here, php cs fixer hallucinated?

Nope. The cs fixer just has another minor update that introduced new rules. I've updated in another branch.

henderkes:
>Sigh. This is why I like to keep downloading and building completely separate. It's also one of the things I don't like about Go: it always requires a network connection to GitHub.

Technically we can pull everything in the download phase and use `go build` manually. It's just rather annoying to do - xcaddy doesn't fail with the download too often (not once in CI yet).

We can add frankenphp as a downloader source and store the version, though. What do you think?

crazywhalecc:
> We can add frankenphp as a downloader source and store the version, though. What do you think?

Initially I want this way, but it will be more redundant downloads for building non-frankenphp sapis. And I remember we discussed it before, the `xcaddy` seems no way to download without building.

Only for this PR related thing: I think we could fetch frankenphp version after `go-xcaddy` package installed.

PR #896, opened 2025-09-08
Fix zig -isystem overriding

henderkes:
## What does this PR do?

and rename folders from pkgroot/$arch-$os/zig-$arch-$os to pkgroot/$arch-$os/zig

### DIFF
--- a/src/SPC/store/scripts/zig-cc.sh
+++ b/src/SPC/store/scripts/zig-cc.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 SCRIPT_DIR="$(dirname "${BASH_SOURCE[0]}")"
-BUILDROOT_ABS="$(realpath "$SCRIPT_DIR/../../buildroot/include" 2>/dev/null || true)"
+BUILDROOT_ABS="$(realpath "$SCRIPT_DIR/../../../buildroot/include" 2>/dev/null || true)"
 PARSED_ARGS=()
 
 while [[ $# -gt 0 ]]; do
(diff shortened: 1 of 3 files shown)


src/SPC/store/scripts/zig-cc.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 SCRIPT_DIR="$(dirname "${BASH_SOURCE[0]}")"
-BUILDROOT_ABS="$(realpath "$SCRIPT_DIR/../../buildroot/include" 2>/dev/null || true)"
+BUILDROOT_ABS="$(realpath "$SCRIPT_DIR/../../../buildroot/include" 2>/dev/null || true)"
```
crazywhalecc, line 4: How about using `BUILD_INCLUDE_PATH` variable? We call `zig-cc.sh` inside SPC can use it.

crazywhalecc (approved):
I noticed the other packages would have the same issue here, and these are working with using `extract-files` to solve this. So far so good for me, but for unification it looks better to refactor them all in 3.0.

src/SPC/store/scripts/zig-cc.sh
henderkes: That would make it a problem with docker. or when the env variable was set at oen time but then unset at another We need to use relative paths unfortunately.

crazywhalecc:
```php
// ConsoleApplication.php
require_once ROOT_DIR . '/src/globals/internal-env.php';

// internal-env.php
putenv('BUILD_INCLUDE_PATH=' . BUILD_INCLUDE_PATH);
```

It shouldn't have issue with docker, it's pre-defined in SPC startup, and it said in `env.ini`: these variables cannot be changed anywhere.

henderkes: Oh, you mean like passing the environment variable into the script. I thought you meant hardcoding it at the time of running install-pkg. That makes sense.

PR #897, opened 2025-09-09
fix CustomPackage handling

henderkes:
## What does this PR do?

- splits getPath() into its own method
- make use of the ::isInstalled() method instead of manually checking paths

### DIFF
--- a/src/SPC/store/pkg/CustomPackage.php
+++ b/src/SPC/store/pkg/CustomPackage.php
@@ -30,9 +30,13 @@ abstract class CustomPackage
 
     /**
      * Get the environment variables this package needs to be usable.
-     * PATH needs to be appended, rather than replaced.
      */
     abstract public static function getEnvironment(): array;
+
+    /**
+     * Get the PATH required to use this package.
+     */
+    abstract public static function getPath(): ?string;
 
     abstract public static function isInstalled(): bool;
 
--- a/src/SPC/builder/BuilderBase.php
+++ b/src/SPC/builder/BuilderBase.php
@@ -10,6 +10,7 @@ use SPC\exception\WrongUsageException;
 use SPC\store\Config;
 use SPC\store\FileSystem;
 use SPC\store\LockFile;
+use SPC\store\pkg\GoXcaddy;
 use SPC\store\SourceManager;
 use SPC\store\SourcePatcher;
 use SPC\util\AttributeMapper;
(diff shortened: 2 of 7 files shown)


src/SPC/store/pkg/CustomPackage.php
```diff
@@ -30,10 +30,14 @@ abstract public function fetch(string $name, bool $force = false, ?array $config
...
     /**
      * Get the environment variables this package needs to be usable.
-     * PATH needs to be appended, rather than replaced.
      */
     abstract public static function getEnvironment(): array;
 
+    /**
+     * Get the PATH required to use this package.
+     */
+    abstract public static function getPath(): ?string;
```
crazywhalecc, line 39: I think here can return a default value `PKG_ROOT_PATH . '/bin'`, rather than making an abstract function.

henderkes: PKG_ROOT_PATH . '/bin' doesn't exist. PKG_ROOT_PATH . '/$name/bin' is not guaranteed to exist. Both go and zig use non-standard locations for the binaries.

crazywhalecc: If we only look at these two special packages, we don't seem to need to add this abstract method - these two getPath calls are both certain classes.

henderkes: Yes, but I would expect more packages to need to add a path to the PATH. We already have the abstract getEnvironment method because of the same assumption, so I believe it should be made here too. In case we don't need one, the method can return null, but it's a hint to us to think about whether the package needs a path, so I make it abstract.

henderkes: @crazywhalecc if this is a blocker for you, I'm okay to change it to a function that returns null by default. We should look into merging this PR, though, because at the moment spc craft doesn't work until this is merged.

crazywhalecc: Ah I just on business trip and have almost no time to deal with other things. Will check it soon.

crazywhalecc: I'm open to changes in this PR, but I might also remove or modify it in 3.0. I don't think it's a big deal since we haven't finalized vendor mode yet.

crazywhalecc (approved):

PR #905, opened 2025-09-18
fix building of shared extensions (grpc, simdjson, soap)

henderkes:
## What does this PR do?
fixes grpc shared compilation - it just ignores CXXFLAGS and instead abuses CPPFLAGS for what should in CXXFLAGS...
closes https://github.com/crazywhalecc/static-php-cli/issues/908

## Checklist before merging

  - [x] `composer cs-fix`
  - [x] `composer analyse`
  - [ ] `composer test`

### DIFF
--- a/config/env.ini
+++ b/config/env.ini
@@ -34,6 +34,7 @@
 ;   SPC_LINUX_DEFAULT_CC:                   the default compiler for linux. (For alpine linux: `gcc`, default: `$GNU_ARCH-linux-musl-gcc`)
 ;   SPC_LINUX_DEFAULT_CXX:                  the default c++ compiler for linux. (For alpine linux: `g++`, default: `$GNU_ARCH-linux-musl-g++`)
 ;   SPC_LINUX_DEFAULT_AR:                   the default archiver for linux. (For alpine linux: `ar`, default: `$GNU_ARCH-linux-musl-ar`)
+;   SPC_EXTRA_PHP_VARS:                     the extra vars for building php, used in `configure` and `make` command.
 
 [global]
 ; Build concurrency for make -jN, default is CPU_COUNT, this value are used in every libs.
--- a/src/SPC/builder/Extension.php
+++ b/src/SPC/builder/Extension.php
@@ -427,13 +427,15 @@ class Extension
             logger()->info("Extension [{$this->getName()}] patched before shared configure");
         }
 
+        $phpvars = getenv('SPC_EXTRA_PHP_VARS') ?: '';
+
         shell()->cd($this->source_dir)
             ->setEnv($env)
             ->appendEnv($this->getExtraEnv())
             ->exec(
                 './configure ' . $this->getUnixConfigureArg(true) .
                 ' --with-php-config=' . BUILD_BIN_PATH . '/php-config ' .
-                '--enable-shared --disable-static ' . getenv('SPC_EXTRA_PHP_VARS')
+                "--enable-shared --disable-static {$phpvars}"
             );
 
         if ($this->patchBeforeSharedMake()) {
--- a/src/SPC/builder/linux/LinuxBuilder.php
+++ b/src/SPC/builder/linux/LinuxBuilder.php
@@ -95,6 +95,8 @@ class LinuxBuilder extends UnixBuilderBase
             // 'LIBS' => SPCTarget::getRuntimeLibs(), // do not pass static libraries here yet, they may contain polyfills for libc functions!
         ]);
 
+        $phpvars = getenv('SPC_EXTRA_PHP_VARS') ?: '';
+
         $embed_type = getenv('SPC_CMD_VAR_PHP_EMBED_TYPE') ?: 'static';
         if ($embed_type !== 'static' && SPCTarget::isStatic()) {
             throw new WrongUsageException(
@@ -116,7 +118,7 @@ class LinuxBuilder extends UnixBuilderBase
                 $json_74 .
                 $zts .
                 $maxExecutionTimers .
-                getenv('SPC_EXTRA_PHP_VARS') . ' ' .
+                "{$phpvars} " .
                 $this->makeStaticExtensionArgs() . ' '
         ));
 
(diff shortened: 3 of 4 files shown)


henderkes:
meh, it looks like that's not enough. may have to remove -Wl,--as-needed instead (just for ImageMagick)

henderkes:
closes #908

src/globals/functions.php
```diff
@@ -245,6 +245,16 @@ function clean_spaces(string $string): string
     return trim(preg_replace('/\s+/', ' ', $string));
 }
 
+function deduplicate_flags(string $flags): string
```
crazywhalecc, line 255: We need PHPDoc or docs for new functions.

src/SPC/builder/unix/library/imagemagick.php
```diff
@@ -12,6 +12,11 @@ trait imagemagick
 {
     protected function build(): void
     {
+        $original_ldflags = $this->builder->arch_ld_flags;
+        if (str_contains($this->builder->arch_ld_flags, '-Wl,--as-needed')) {
+            $this->builder->arch_ld_flags = str_replace('-Wl,--as-needed', '', $original_ldflags);
+        }
+
```
crazywhalecc, line 19: Is it possible to change only for this scope rather than modifying builder property? I don't think modifying builder's value is the best idea.

src/SPC/util/SPCConfigUtil.php
```diff
@@ -87,7 +87,7 @@ public function config(array $extensions = [], array $libraries = [], bool $incl
         if (SPCTarget::getTargetOS() === 'Darwin') {
             $libs .= " {$this->getFrameworksString($extensions)}";
         }
-        if ($this->builder->hasCpp()) {
+        if ($this->builder->hasCpp() || collect($extensions)->contains(static fn (string $ext) => Config::getExt($ext)['cpp-extension'] ?? false)) {
```
crazywhalecc: Since the only usage for `builder->hasCpp()` is here, I think it could be moved to SPCConfigUtil.

src/SPC/builder/extension/simdjson.php
```diff
@@ -31,4 +33,18 @@ public function patchBeforeBuildconf(): bool
...
+    public function getSharedExtensionEnv(): array
+    {
+        $env = parent::getSharedExtensionEnv();
+        if (ToolchainManager::getToolchainClass() === ZigToolchain::class) {
+            $extra = getenv('SPC_COMPILER_EXTRA');
+            if (!str_contains((string) $extra, '-lstdc++')) {
+                f_putenv('SPC_COMPILER_EXTRA=' . clean_spaces($extra . ' -lstdc++'));
+            }
+            $env['CFLAGS'] .= ' -Xclang -target-feature -Xclang +evex512';
+            $env['CXXFLAGS'] .= ' -Xclang -target-feature -Xclang +evex512';
```
crazywhalecc, line 46: For generic use, it's not happy with old CPU though.

henderkes: Unfortunately, simdjson fails to build with zig on older cpus. We would need to patch every usage of it in the source code.

src/globals/functions.php
henderkes: Okay, I will add it if it makes sense to keep the function. Might inline it into where it was used.

henderkes:
I'll work on it next week. On a business trip until Sunday.

src/SPC/builder/unix/library/imagemagick.php
henderkes: I can't think of a way, unfortunately, other than patching imagemagick's ./configure, but that would be much worse than temporarily changing the ldflags.

src/SPC/builder/Extension.php
```diff
@@ -455,13 +429,17 @@ public function buildUnixShared(): void
...
+            ToolchainManager::getToolchainClass() !== ZigToolchain::class;
+
         shell()->cd($this->source_dir)
             ->setEnv($env)
             ->appendEnv($this->getExtraEnv())
             ->exec(
                 './configure ' . $this->getUnixConfigureArg(true) .
                 ' --with-php-config=' . BUILD_BIN_PATH . '/php-config ' .
-                '--enable-shared --disable-static'
+                '--enable-shared --disable-static' . (!$has_avx512 ? ' php_cv_have_avx512=no php_cv_have_avx512vbmi=no' : '')
```
crazywhalecc: This is a php.m4 level variable and I would prefer declaring it with `SPC_CMD_PREFIX_SHARED_EXT_CONFIGURE` as well. Using too many conditions on it makes it confusing and cross-variable dependencies are not straightforward.

As for toolchain class specific option, it looks better to modify it in toolchain init.

src/SPC/builder/linux/LinuxBuilder.php
```diff
@@ -116,6 +122,7 @@ public function buildPHP(int $build_target = BUILD_TARGET_NONE): void
                 $json_74 .
                 $zts .
                 $maxExecutionTimers .
+                (!$has_avx512 ? 'php_cv_have_avx512=no php_cv_have_avx512vbmi=no ' : '') .
```
crazywhalecc: Same above

config/env.ini
```diff
@@ -34,6 +34,7 @@
 ;   SPC_LINUX_DEFAULT_CC:                   the default compiler for linux. (For alpine linux: `gcc`, default: `$GNU_ARCH-linux-musl-gcc`)
 ;   SPC_LINUX_DEFAULT_CXX:                  the default c++ compiler for linux. (For alpine linux: `g++`, default: `$GNU_ARCH-linux-musl-g++`)
 ;   SPC_LINUX_DEFAULT_AR:                   the default archiver for linux. (For alpine linux: `ar`, default: `$GNU_ARCH-linux-musl-ar`)
+;   SPC_EXTRA_PHP_VARS:                     the extra vars for building php, used in `configure` and `make` command.
```
henderkes, line 37: I don't think this needs to be exposed to users. They can use the configure command variable.

crazywhalecc: This is not expose, just for env notes. Moreover, env.ini itself should not be modified at will by people who use it normally.

Issue #911, opened 2025-09-29
Use latest pecl version for built-in extensions


henderkes:
I realized that the 8.3/8.4 releases don't always ship the latest compatible version for e.g. php-zip, which is found on pecl. We should get the source from pecl and overwrite the one built into php-src.

henderkes:
Heh, I had hoped this would be easy, but e.g. pecl's zip extension doesn't work for static compilation (it attempts to use $PHP_CONFIG).

henderkes:
Unfortunately, it seems like even when patching config.m4 to use the correct php version and add -lz, it's still failing to build statically:

```log
/bin/sh /home/m/static-php-cli/source/php-src/libtool --silent --preserve-dup-deps --tag=CC --mode=compile gcc -Isapi/cli/ -I/home/m/static-php-cli/source/php-src/sapi/cli/ -I/home/m/static-php-cli/source/php-src/main -I/home/m/static-php-cli/source/php-src -I/home/m/static-php-cli/source/php-src/ext/date/lib -I/home/m/static-php-cli/source/php-src/ext/zip/php81 -I/home/m/static-php-cli/source/php-src/TSRM -I/home/m/static-php-cli/source/php-src/Zend  -I/home/m/static-php-cli/buildroot/include -D_GNU_SOURCE  -fno-common -Wstrict-prototypes -Wformat-truncation -Wlogical-op -Wduplicated-cond -Wno-clobbered -Wall -Wextra -Wno-unused-parameter -Wno-sign-compare -fPIC -Os -fPIE -ffp-contract=off -fvisibility=hidden -Wimplicit-fallthrough=1 -DZEND_SIGNALS  -g -fstack-protector-strong -fno-ident -fPIE -fPIC -Os -DZEND_ENABLE_STATIC_TSRMLS_CACHE=1 -c /home/m/static-php-cli/source/php-src/sapi/cli/php_cli_process_title.c -o sapi/cli/php_cli_process_title.lo  -MMD -MF sapi/cli/php_cli_process_title.dep -MT sapi/cli/php_cli_process_title.lo
main/internal_functions_cli.c:45:9: error: ‘phpext_zip_ptr’ undeclared here (not in a function); did you mean ‘phpext_spl_ptr’?
   45 |         phpext_zip_ptr,
      |         ^~~~~~~~~~~~~~
```

Makes me believe it hard assumes to be built as a shared extension.

henderkes:
[...]

PR #913, opened 2025-10-02
add -shared suffix extension to allow building newest e.g. zip.so

henderkes:
## What does this PR do?

WIP

## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`

### DIFF
--- a/config/ext.json
+++ b/config/ext.json
@@ -1171,6 +1171,27 @@
             "xz"
         ]
     },
+    "zip-shared": {
+        "support": {
+            "BSD": "wip"
+        },
+        "type": "external",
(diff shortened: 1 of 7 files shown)


henderkes:
see #911

henderkes:
Maybe simpler idea: extract to php-src/zip/spc-shared and if the directory exists, build from there in Extension.php. Not sure if this would also override $this->source_dir, though.

henderkes:
The only frequently changing version of these is `zip`. Maybe instead of doing something special here, we can just override the buildshared method of zip.php.

Issue #757, opened 2025-06-10
update to postgresql 18


henderkes:
we're still on version 16. 17 changes a few things that I didn't want to add to the already massive PR, but we should tackle it afterwards

crazywhalecc:
Seems we cannot upgrade version tag and update the patch for specific line as we do for 16. Also, I found vcpkg does not support libpq 17 either.

The mail list result shows that the pgsql team still don't want anyone to statically linking against libpq.a.

henderkes:
To my big surprise, I upgraded to pgsql 18 and it's actually working without changing our source at all.

henderkes:
We can simplify the source a lot though. Version 17.0+ is always thread-safe and we don't need to check for >= 16.1 anymore,

henderkes:
It needed some source changes for ancient glibc (HWCAP macros not defined) and macOS (it ignores LIBS="...", had to str_replace -lldap).

PR #916, opened 2025-10-06
update pgsql version to 18.0, php 8.5 to RC2

henderkes:
## What does this PR do?
closes https://github.com/crazywhalecc/static-php-cli/issues/757

### DIFF
--- a/src/SPC/builder/linux/LinuxBuilder.php
+++ b/src/SPC/builder/linux/LinuxBuilder.php
@@ -157,7 +157,11 @@ class LinuxBuilder extends UnixBuilderBase
         $shared_extensions = array_map('trim', array_filter(explode(',', $this->getOption('build-shared'))));
         if (!empty($shared_extensions)) {
             if (SPCTarget::isStatic()) {
-                throw new WrongUsageException("You're building against musl libc statically, but you're trying to build shared extensions. Static musl libc does not implement `dlopen`, so your php binary is not able to load shared extensions.");
+                throw new WrongUsageException(
+                    "You're building against musl libc statically (the default on Linux), but you're trying to build shared extensions.\n" .
+                    'Static musl libc does not implement `dlopen`, so your php binary is not able to load shared extensions.' . "\n" .
+                    'Either use SPC_LIBC=glibc to link against glibc on a glibc OS, or use SPC_TARGET="native-native-musl -dynamic" to link against musl libc dynamically using `zig cc`.'
+                );
             }
             logger()->info('Building shared extensions...');
             $this->buildSharedExts();


src/SPC/builder/unix/library/postgresql.php
```diff
@@ -65,32 +83,38 @@ protected function build(): void
...
-            shell()->cd($this->source_dir . '/build')
-                ->exec('sed -i.backup "s/invokes exit\'; exit 1;/invokes exit\';/"  ../src/interfaces/libpq/Makefile')
-                ->exec('sed -i.backup "278 s/^/# /"  ../src/Makefile.shlib')
-                ->exec('sed -i.backup "402 s/^/# /"  ../src/Makefile.shlib');
-        } else {
-            throw new BuildFailureException('Unsupported version for postgresql: ' . $version . ' !');
+        # 有静态链接配置  参考文件： src/interfaces/libpq/Makefile
+        shell()->cd($this->source_dir . '/build')
+            ->exec('sed -i.backup "s/invokes exit\'; exit 1;/invokes exit\';/"  ../src/interfaces/libpq/Makefile')
+            ->exec('sed -i.backup "278 s/^/# /"  ../src/Makefile.shlib')
```
crazywhalecc: I remember that The number of rows in different major versions is different, it may crash the file if version is not specified. 

But I haven't verified it yet, I will verify it locally later too.

henderkes: Not sure about this, I didn't change it and it continued working. Not sure what that was for or if it was even required.

henderkes:
> Looks good to me, but it needs connection and runtime test. Will do later.

Works for me.

```php
<?php
$host = "localhost";
$db   = "appdb";
$user = "appuser";
$pass = "strongpassword";

$dsn = "pgsql:host=$host;dbname=$db";

try {
    $pdo = new PDO($dsn, $user, $pass, [
        PDO::ATTR_ERRMODE => PDO::ERRMODE_EXCEPTION
    ]);
} catch (PDOException $e) {
    die("Connection failed: " . $e->getMessage());
}

// Create table if not exists
$pdo->exec("
    CREATE TABLE IF NOT EXISTS test_data (
        id SERIAL PRIMARY KEY,
        value TEXT NOT NULL
    );
");

// Insert random data
$randomValue = bin2hex(random_bytes(8));
$stmt = $pdo->prepare("INSERT INTO test_data (value) VALUES (:val)");
$stmt->execute([':val' => $randomValue]);

// Select and print all rows
$stmt = $pdo->query("SELECT id, value FROM test_data ORDER BY id ASC");
$rows = $stmt->fetchAll(PDO::FETCH_ASSOC);

header('Content-Type: text/plain');
foreach ($rows as $row) {
    echo $row['id'] . " | " . $row['value'] . "\n";
}
?>
```

```
[m@M static-php-cli]$ buildroot/bin/php ../pgsql.php
1 | bbd58a0ed06e6857
2 | d13bef167723c0d3
```

src/SPC/builder/unix/library/postgresql.php
crazywhalecc: It's commenting `$(INSTALL_SHLIB) $< '$(DESTDIR)$(pkglibdir)/$(shlib)'` like this: to disable shared libs building and linking. For 17 and 18, it's from different lines.

crazywhalecc:
I refactored a few things for pgsql. Now it's compatible with internal SPCConfigUtil with no pain.

For specific line patching, I found it could be disabled by just removing them, make it version-independent now.

PR #917, opened 2025-10-06
use latest zip version for building shared

henderkes:
## What does this PR do?
see title


henderkes:
@crazywhalecc this should be an easy merge. Pecl is reliable and doesn't rate limit and it's only used for shared compilation, static compilation still uses the built in version.

config/source.json
```diff
@@ -263,6 +263,16 @@
             "path": "LICENSE"
         }
     },
+    "ext-zip": {
+        "type": "url",
+        "url": "https://pecl.php.net/get/zip",
+        "path": "php-src/ext/zip/latest",
```
crazywhalecc: This may be confused. It could be in `source` dir by default.

henderkes:
```suggestion
```

crazywhalecc: Just remove this to let it extract to `source/{source name}`.

PR #924, opened 2025-10-13
Fix PostgreSQL patch compatibility

crazywhalecc:
## What does this PR do?

Just in case someone still want to build with older version like 16. The old version may not have these files need to be patched.

## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`

### DIFF
--- a/src/SPC/builder/unix/library/postgresql.php
+++ b/src/SPC/builder/unix/library/postgresql.php
@@ -50,7 +50,7 @@ trait postgresql
         $config = $spc->config(libraries: $libs, include_suggest_lib: $this->builder->getOption('with-suggested-libs'));
 
         $env_vars = [
-            'CFLAGS' => "{$config['cflags']} -fno-ident",
+            'CFLAGS' => $config['cflags'],
             'CPPFLAGS' => '-DPIC',
             'LDFLAGS' => $config['ldflags'],
             'LIBS' => $config['libs'],


src/SPC/builder/unix/library/postgresql.php
```diff
@@ -47,7 +49,7 @@ protected function build(): void
         $spc = new SPCConfigUtil($this->getBuilder(), ['no_php' => true, 'libs_only_deps' => true]);
         $config = $spc->config(libraries: $libs, include_suggest_lib: $this->builder->getOption('with-suggested-libs'));
 
-        $macos_15_bug_cflags = PHP_OS_FAMILY === 'Darwin' ? ' -Wno-unguarded-availability-new' : '';
+        $macos_15_bug_cflags = PHP_OS_FAMILY === 'Darwin' ? '' : '';
```
henderkes: huh?

src/SPC/builder/unix/library/postgresql.php
```diff
@@ -13,17 +14,18 @@ trait postgresql
...
-            );
+            try {
+                FileSystem::replaceFileStr("{$this->source_dir}/src/port/pg_popcount_aarch64.c", 'HWCAP_SVE', '0');
+                FileSystem::replaceFileStr(
+                    "{$this->source_dir}/src/port/pg_crc32c_armv8_choose.c",
+                    '#if defined(__linux__) && !defined(__aarch64__) && !defined(HWCAP2_CRC32)',
+                    '#if defined(__linux__) && !defined(HWCAP_CRC32)'
+                );
+            } catch (FileSystemException) {
+                // allow file not-existence to make it compatible with old and new version
```
henderkes, line 27: I kind of disahree with this. If someone would like to build with older versions, they should use an older version of spc. We have REL_18 listed as a version constraint in the source.

src/SPC/builder/unix/library/postgresql.php
crazywhalecc: To verify that newer version does fix the issue that here bypassed. I'll remove later if my local vm works well.

src/SPC/builder/unix/library/postgresql.php
crazywhalecc: That's another issue and I don't want to introduce more breaking changes in patch version. We could drop support if there are really hard or unnecessary to maintain, but here it's just simply fixing unexpected patch exception for older missing files.

henderkes: The nature of the project doesn't allow us to stay backwards compatible with all sources, so I think it would be totally fine to violate it. But if this is enough to fix it for pgsql 16 (17 never succeeded, by the way) then so be it.

crazywhalecc:
~~That's odd. The intel runner always queued~~

src/SPC/builder/unix/library/postgresql.php
crazywhalecc: BTW this will also fix the 17. I build successfully with 16, 17 and 18 on my local arm64 linux.

src/SPC/builder/unix/library/postgresql.php
```diff
@@ -49,10 +49,8 @@ protected function build(): void
         $spc = new SPCConfigUtil($this->getBuilder(), ['no_php' => true, 'libs_only_deps' => true]);
         $config = $spc->config(libraries: $libs, include_suggest_lib: $this->builder->getOption('with-suggested-libs'));
 
-        $macos_15_bug_cflags = PHP_OS_FAMILY === 'Darwin' ? '' : '';
-
         $env_vars = [
-            'CFLAGS' => "{$config['cflags']} -fno-ident{$macos_15_bug_cflags}",
+            'CFLAGS' => "{$config['cflags']} -fno-ident",
```
henderkes:
```suggestion
            'CFLAGS' => $config['cflags'],
```

PR #927, opened 2025-10-15
Add Windows CGI SAPI build support

### LINKED ISSUE #879: [Feature Request] Windows CGI API
As I saw in many other projects (and even participating in a few discutions in here) the uses of fpm apps in a desktop enviroment has been incrementing in the last few months. Unfortunately, the FPM api does not work in windows, but the CGI binary works in most (if not all) the cases. Having a compiled CGI binary would make those multiplatform apps not exclude windows

crazywhalecc:
## What does this PR do?

- Fix #595 
- Fix #879 

## Checklist before merging

  - [X] `composer cs-fix`
  - [X] `composer analyse`
  - [X] `composer test`

### DIFF
--- a/src/SPC/builder/windows/WindowsBuilder.php
+++ b/src/SPC/builder/windows/WindowsBuilder.php
@@ -103,14 +103,13 @@ class WindowsBuilder extends BuilderBase
             ->exec(
                 "{$this->sdk_prefix} configure.bat --task-args \"" .
                 '--disable-all ' .
-                '--disable-cgi ' .
                 '--with-php-build=' . BUILD_ROOT_PATH . ' ' .
                 '--with-extra-includes=' . BUILD_INCLUDE_PATH . ' ' .
                 '--with-extra-libs=' . BUILD_LIB_PATH . ' ' .
-                ($enableCli ? '--enable-cli=yes ' : '--enable-cli=no ') .
-                ($enableMicro ? ('--enable-micro=yes ' . $micro_logo . $micro_w32) : '--enable-micro=no ') .
-                ($enableEmbed ? '--enable-embed=yes ' : '--enable-embed=no ') .
-                ($enableCgi ? '--enable-cgi=yes ' : '--enable-cgi=no ') .
+                ($enableCli ? '--enable-cli ' : '--disable-cli ') .
+                ($enableMicro ? ('--enable-micro ' . $micro_logo . $micro_w32) : '--disable-micro ') .
+                ($enableEmbed ? '--enable-embed ' : '--disable-embed ') .
+                ($enableCgi ? '--enable-cgi ' : '--disable-cgi ') .
                 $config_file_scan_dir .
                 $opcache_jit_arg .
                 "{$this->makeStaticExtensionArgs()} " .


src/SPC/builder/windows/WindowsBuilder.php
```diff
@@ -159,6 +165,24 @@ public function buildCli(): void
...
+        $extra_libs = getenv('SPC_EXTRA_LIBS') ?: '';
+
+        // add nmake wrapper
+        FileSystem::writeFile(SOURCE_PATH . '\php-src\nmake_cgi_wrapper.bat', "nmake /nologo LIBS_CGI=\"ws2_32.lib shell32.lib {$extra_libs}\" EXTRA_LD_FLAGS_PROGRAM= %*");
+
+        cmd()->cd(SOURCE_PATH . '\php-src')->exec("{$this->sdk_prefix} nmake_cgi_wrapper.bat --task-args php-cgi.exe");
+
+        // deploy cgi binary
+        logger()->info('Deploying cgi file');
+        FileSystem::createDir(BUILD_ROOT_PATH . '\bin');
```
henderkes: BUILD_BIN_PATH

src/SPC/builder/windows/WindowsBuilder.php
```diff
@@ -298,6 +322,17 @@ public function sanityCheck(mixed $build_target): void
                 }
             }
         }
+
+        // sanity check for php-cgi
+        if (($build_target & BUILD_TARGET_CGI) === BUILD_TARGET_CGI) {
+            logger()->info('running cgi sanity check');
+            FileSystem::writeFile(SOURCE_PATH . '\php-cgi-test.php', '<?php echo "<h1>Hello, World!</h1>"; ?>');
```
henderkes, line 324: place this with our other tests I'd say

src/SPC/store/SourcePatcher.php
```diff
@@ -549,6 +549,31 @@ public static function patchWindowsCLITarget(): void
...
+    /**
+     * Patch cgi SAPI Makefile for Windows.
+     */
+    public static function patchWindowsCGITarget(): void
+    {
+        // search Makefile code line contains "$(BUILD_DIR)\php.exe:"
+        $content = FileSystem::readFile(SOURCE_PATH . '/php-src/Makefile');
+        $lines = explode("\r\n", $content);
+        $line_num = 0;
+        $found = false;
```
henderkes, line 561: can't we use a patch file for this?

henderkes (changes requested):

src/SPC/builder/windows/WindowsBuilder.php
crazywhalecc: This is the right place. The other tests for other OS is in `UnixBuilderBase` and win is not unix.

src/SPC/store/SourcePatcher.php
crazywhalecc: It copied from cli target patches from the same file. Of course we could use a patch, but I just want to keep unification for current situation. WindowsBuilder has other things more need to refactor for now.

crazywhalecc:
The low version problem is a bit tricky. I build the exact same content locally and remotely, but both errors are reported, but the errors are different.

crazywhalecc:
The implementation seems to be tricky in PHP <= 8.3, which requires debugging this crash.

crazywhalecc:
I've fixed the issue with older CGI builds. It's related to the Win32 API. Since the patches are more relevant to phpmicro, I've updated them to the static-php's phpmicro fork project.

For PHP 8.5, it's related to some unsupported extensions, such as sqlsrv. [Remi's roadmap](https://github.com/remicollet/remirepo/issues/307) seems to be up to date and good for reference. This isn't relevant to this pull request. We'll need to track it later in another issue or pull request.

henderkes:
Remi's tracker is based on PR's existing or php 8.5 compatibility being merged into master. Most extensions haven't had a stable release with the fixes, so the roadmap is inapplicable unless we overhaul our source.json into an unstable state, which I do not want to do.

We'll just have to wait - until then, some extensions just won't be 8.5 compatible.

src/SPC/builder/windows/WindowsBuilder.php
```diff
@@ -109,6 +110,7 @@ public function buildPHP(int $build_target = BUILD_TARGET_NONE): void
                 ($enableCli ? '--enable-cli=yes ' : '--enable-cli=no ') .
                 ($enableMicro ? ('--enable-micro=yes ' . $micro_logo . $micro_w32) : '--enable-micro=no ') .
                 ($enableEmbed ? '--enable-embed=yes ' : '--enable-embed=no ') .
+                ($enableCgi ? '--enable-cgi=yes ' : '--enable-cgi=no ') .
```
henderkes: we should change these to be in line with the unix version

$enable ? '--enable-sapi' : '--disable-sapi'.

henderkes (approved):

PR #930, opened 2025-10-17
Add log dir expose, unify SPC_FIX_DEPLOY_ROOT parsing

### LINKED ISSUE #929: spc-alpine-docker doesn't expose logs
When using the bin/spc-alpine-docker command to build the binary i ran into an issue.

The command doesn't expose the log folder so we cannot easily read the log output if there's a problem.

crazywhalecc:
## What does this PR do?

Fix #929 

## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,13 +1,9 @@
 name: Tests
 
 on:
-  push:
-    branches:
-      - main
-    paths:
-      - 'src/globals/test-extensions.php'
   pull_request:
     branches: [ "main" ]
+    types: [ opened, synchronize, reopened ]
     paths:
       - 'src/**'
       - 'config/**'


henderkes:
While we're at it, could we try to use the same user inside a docker container as the host uses?

`bin/spc-alpine-docker build bcmath --build-cli`
`rm -rf source buildroot` -> error, needs sudo, because we're not the owner of the files

crazywhalecc:
Like `--user "$(id -u):$(id -g)"`?

henderkes:
Yep

crazywhalecc:
Unfortunately there are no simple solution to use `$(whoami)` current user as mount dir for all docker engines. It depends on docker engine platform and version. Using `--user` will make docker inside full of permission denied error. 

The only perfect option for linux is to restart docker CE as rootless mode. But luckily macOS docker always create mount dir as current user permission.

henderkes:
Restarting docker-ce in a different mode is not a solution, as it may be running on a host and linux could be a VM without docker installed.

If Docker can't handle it natively, there's a chown at the end, but that would prompt a password and only work if the user has sudo...

henderkes (approved):

PR #934, opened 2025-10-20
PIE download support & Downloader and ConfigValidator enhance

crazywhalecc:
## What does this PR do?

- Add pie download type support
- Use strip when extracting with `unzip` and `7z.exe` (This may impact all windows related things, will fix them later)
- Strict ConfigValidator, fix potential wrong configs
- Update composer dependencies

The pie download is not available for many pecl extensions yet, it is not appropriate to replace all extensions downloaded from pecl.php.net with pie now.

## Checklist before merging

  - [X] `composer cs-fix`
  - [X] `composer analyse`
  - [X] `composer test`

### DIFF
--- a/src/SPC/store/FileSystem.php
+++ b/src/SPC/store/FileSystem.php
@@ -635,7 +635,7 @@ class FileSystem
 
     private static function extractWithType(string $source_type, string $filename, string $extract_path): void
     {
-        logger()->debug('Extracting source [' . $source_type . ']: ' . $filename);
+        logger()->debug("Extracting source [{$source_type}]: {$filename}");
         /* @phpstan-ignore-next-line */
         match ($source_type) {
             SPC_SOURCE_ARCHIVE => self::extractArchive($filename, $extract_path),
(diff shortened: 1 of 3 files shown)


henderkes:
Looks good so far, but I'll need to check out the lib changes on a pc later.

crazywhalecc:
Windows pkgs using zip all failed as expected. The dir stripping is not good for everything, some zip needs wrapping in previous dir.

src/SPC/store/FileSystem.php
```diff
@@ -644,4 +644,59 @@ private static function extractWithType(string $source_type, string $filename, s
             SPC_SOURCE_LOCAL => symlink(self::convertPath($filename), $extract_path),
         };
     }
+
+    /**
+     * Unzip file with stripping top-level directory
+     */
+    private static function unzipWithStrip(string $zip_file, string $extract_path): void
```
henderkes, line 651: where does this method strip anything? what is stripped?

crazywhalecc: Unlike `--strip-components 1` behavior in `tar` command, `unzip` does not have this function so far. Some archives compress to `libXXX-B1A1E04F2B99/*` like this and we cannot predict what's its name. 

Strip component can extract the only dir to our own name, that's `tar` extraction we are using now.

crazywhalecc: But current version of `unzipWithStrip` does not work with `strawberry-perl` and other packages for Windows, it does need put into a dedicated dir.

PR #945, opened 2025-10-24
Fix libevent build, use newer cmake options

crazywhalecc:
## What does this PR do?

Fix https://github.com/static-php/static-php-cli-hosted/actions/runs/18767020353/job/53543930166

## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`


src/SPC/util/executor/UnixCMakeExecutor.php
```diff
@@ -205,7 +205,7 @@ private function makeCmakeToolchainFile(): string
 SET(CMAKE_INSTALL_LIBDIR "lib")
 
 set(PKG_CONFIG_EXECUTABLE "{$pkgConfigExecutable}")
-list(APPEND PKG_CONFIG_EXECUTABLE "--static")
+set(PKG_CONFIG_ARGN "--static" CACHE STRING "Extra arguments for pkg-config" FORCE)
```
crazywhalecc, line 208: `PKG_CONFIG_ARGN` introduced in CMake 3.22

henderkes (approved):
weird, libevent compiles perfectly fine for me

where does it fail to compile with the old option?

crazywhalecc:
> weird, libevent compiles perfectly fine for me
> 
> where does it fail to compile with the old option?

No idea. This is not related to libevent itself, maybe using CMake 3.x works well, but I have upgraded to cmake 4 already. The safest way at the moment is to use this new version's var and add the `--static` option to override.

henderkes:
I've been on CMake 4 since the week it came out. I remember creating a rather large PR back then that removed deprecated options everywhere, before we had the UnixCmakeConfigurator class.

PR #952, opened 2025-10-29
fix postgresql libraries when --with-suggested-libs is false

henderkes:
## What does this PR do?
fix https://github.com/php/frankenphp/issues/1937

## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`

### DIFF
--- a/src/SPC/builder/unix/library/postgresql.php
+++ b/src/SPC/builder/unix/library/postgresql.php
@@ -45,7 +45,7 @@ trait postgresql
 
     protected function build(): void
     {
-        $libs = array_map(fn ($x) => $x->getName(), $this->getDependencies());
+        $libs = array_map(fn ($x) => $x->getName(), $this->getDependencies(true));
         $spc = new SPCConfigUtil($this->builder, ['no_php' => true, 'libs_only_deps' => true]);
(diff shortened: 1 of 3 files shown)


config/lib.json
```diff
@@ -607,8 +607,8 @@
     },
     "libxml2": {
         "source": "libxml2",
-        "static-libs-unix": [
-            "libxml2.a"
+        "pkg-configs": [
+            "libxml-2.0"
```
crazywhalecc, line 611: I'm not sure if this will affect other dependencies, because it seems to have caused compilation failures before. I'll have to try a full test.

henderkes: I believe if it caused compilation failures before, we should fix the root cause of them, not that we're not using pkgconfig for libxml.

henderkes:
> We don't need to add these functions to achieve this, just change `postgresql.php` file, add `true` to `getDependencies()` function.
> 
> ```
> $libs = array_map(fn ($x) => $x->getName(), $this->getDependencies(true));
> ```

Good catch. Is there even a point in keeping a non-recursive version of it?

crazywhalecc:
> Good catch. Is there even a point in keeping a non-recursive version of it?

For displaying current-level dependencies to check optional libraries status. It's usually used with my local dependency debugging for some cases. There's no need to remove this.

PR #953, opened 2025-10-31
Export php binary external debug symbol files

### LINKED ISSUE #949: Add option  to extract debug infos into <file>.debug file and link it
like
```php
if (!$this->getOption('no-strip', false) && file_exists(BUILD_BIN_PATH . '/frankenphp')) {
    shell()->cd(BUILD_BIN_PATH)->exec('objcopy --only-keep-debug frankenphp frankenphp.debug');
    if (PHP_OS_FAMILY === 'Linux') {
        shell()->cd(BUILD_BIN_PATH)->exec('strip --strip-unneeded frankenphp');
    } else {
        shell()->cd(BUILD_BIN_PATH)->exec('strip -S frankenphp');
    }
    shell()->cd(BUILD_BIN_PATH)->exec('objcopy --add-gnu-debuglink=frankenphp.debug frankenphp');
}
```

crazywhalecc:
## What does this PR do?

Closes #949 

## Checklist before merging

  - [X] `composer cs-fix`
  - [X] `composer analyse`
  - [X] `composer test`

### DIFF
--- a/src/SPC/store/source/PhpSource.php
+++ b/src/SPC/store/source/PhpSource.php
@@ -16,7 +16,7 @@ class PhpSource extends CustomSourceBase
     {
         $major = defined('SPC_BUILD_PHP_VERSION') ? SPC_BUILD_PHP_VERSION : '8.4';
         if ($major === '8.5') {
-            Downloader::downloadSource('php-src', ['type' => 'url', 'url' => 'https://downloads.php.net/~edorian/php-8.5.0RC4.tar.xz'], $force);
+            Downloader::downloadSource('php-src', ['type' => 'url', 'url' => 'https://downloads.php.net/~daniels/php-8.5.0RC5.tar.xz'], $force);
         } elseif ($major === 'git') {
             Downloader::downloadSource('php-src', ['type' => 'git', 'url' => 'https://github.com/php/php-src.git', 'rev' => 'master'], $force);
         } else {


src/SPC/builder/unix/UnixBuilderBase.php
```diff
@@ -210,21 +211,71 @@ protected function sanityCheck(int $build_target): void
...
+    protected function deployBinary(int $type): void
     {
+        FileSystem::createDir(BUILD_BIN_PATH);
+        $copy_files = [];
         $src = match ($type) {
             BUILD_TARGET_CLI => SOURCE_PATH . '/php-src/sapi/cli/php',
             BUILD_TARGET_MICRO => SOURCE_PATH . '/php-src/sapi/micro/micro.sfx',
             BUILD_TARGET_FPM => SOURCE_PATH . '/php-src/sapi/fpm/php-fpm',
             BUILD_TARGET_CGI => SOURCE_PATH . '/php-src/sapi/cgi/php-cgi',
+            BUILD_TARGET_FRANKENPHP => BUILD_BIN_PATH . '/frankenphp',
```
henderkes, line 308: we also need target embed for libphp(-postfix).so

src/SPC/builder/unix/UnixBuilderBase.php
```diff
@@ -210,21 +211,71 @@ protected function sanityCheck(int $build_target): void
...
+        if (!$no_strip_option && PHP_OS_FAMILY === 'Darwin') {
+            shell()
+                ->exec("dsymutil -f {$src}") // generate .dwarf file
+                ->exec("strip -S {$src}"); // strip unneeded symbols
+            $copy_files[] = "{$src}.dwarf";
+        } elseif (!$no_strip_option && PHP_OS_FAMILY === 'Linux') {
+            shell()
+                ->exec("objcopy --only-keep-debug {$src} {$src}.debug") // extract debug symbols
+                ->exec("objcopy --strip-debug --add-gnu-debuglink={$src}.debug {$src}") // link debug symbols
+                ->exec("strip --strip-unneeded {$src}"); // strip unneeded symbols
```
henderkes: --strip-debug is unnecessary here, --strip-unneeded already catches it right after

src/SPC/builder/unix/UnixBuilderBase.php
henderkes: and for all shared extensions!

crazywhalecc: `embed` is not a executable type and different with other sapis. I would treat `libphp.{a|so|dylib}` and `extension.so` as another type of binaries. I believe we need a different function to handle shared libs.

I've tried adding libphp.so here before, but it's not really suitable for this, especially since UPX compression is involved, and it's clearly not appropriate for the library.

BTW I think the postfix patching could be in a separated patch point file. It is not strongly related to the core build process. I will move it to external patch point, and add more patch point events soon (not in this PR).

henderkes: >It is not strongly related to the core build process.

It is to the php build process. Php appends the suffix based on the extra flags.

crazywhalecc: I'm not talking about the `-release` of LDFLAGS, but the patch for soname. It not quite belongs to `build` though. 

I understand the concerns about separating it out, as this is libtool's approach, but it's better suited to be implemented as a standalone patches.

As for the current version for my perspective, I wouldn't exclude it from the list of default internal patches, but it certainly not appropriate to be here.

henderkes:
I think we should refactor the debug copy out to its own function and put it in buildroot/debug. And of course also do it for shared libs.

crazywhalecc:
> I think we should refactor the debug copy out to its own function and put it in buildroot/debug. And of course also do it for shared libs.

I'm not quite sure if it's necessary to generate both of them simultaneously. But I think we could just keep the unstripped version in `source`. That is, strip after copy. WDYT?

henderkes:
No, it should behave the same way for all binaries.
My idea is to create a function:

```php
function extractDebugInfo($path) {
     $file = filename($path);
     $dir = basename($path);
     shell()->cd($dir)
          ->exec("objcopy ---only-debug-info $file $file.debug");
     ...
     mv($file . '.debug', BUILD_DEBUG_PATH);
}
```

And then call that for every binary in BIN/LIB/MODULES paths.

crazywhalecc:
> And then call that for every binary in BIN/LIB/MODULES paths.

It looks impossible for current situation, we have no defines for the executables that should be stripped. But we could separate functions `stripBinary()`, `extractDebugInfo()` and `deployBinary()` and call them when needed.

src/SPC/builder/Extension.php
```diff
@@ -448,6 +449,15 @@ public function buildUnixShared(): void
...
             ->exec('make install');
+
+        // process *.so file
+        $soFile = BUILD_MODULES_PATH . '/' . $this->getName() . '.so';
+        if (!file_exists($soFile)) {
+            throw new ValidationException("extension {$this->getName()} build failed: {$soFile} not found", validation_module: "Extension {$this->getName()} build");
+        }
+        /** @var UnixBuilderBase $builder */
+        $builder = $this->builder;
+        $builder->deployBinary($soFile, $soFile, false);
```
crazywhalecc, line 460: @henderkes Here and `LinuxBuilder.php:284` shows it will call extractDebugInfo() already. See: `UnixBuilderBase.php:142`.

henderkes: It's missing for shared extensions built by php's make install. All the ones in ext.json that have `"build-with-php": true`

henderkes:
Looks good now, I'll spin it up tonight.

henderkes (approved):

Issue #954, opened 2025-11-03
[RdKafka] OpenSSL not available at build time


lucas-angeli-gimenes:
The Kafka `SASL_SSL` authentication method is not available or working, even though the `openssl` extension is enabled: `Unsupported value \"SASL_SSL\" for configuration property \"security.protocol\": OpenSSL not available at build time`.

When running the command: `php -r "echo (new RdKafka\Conf())->dump()['builtin.features'];"` it's possible to see that the `ssl` option is not available.

I tried to force openssl through the `--with-libs` option, but it still didn't work, for example: bin/spc build --build-cli "openssl,rdkafka,zlib" --with-libs="openssl,librdkafka"

php -m Output:
```
[PHP Modules]
amqp
apcu
ast
bcmath
brotli
calendar
Core
ctype
curl
date
dom
ds
exif
FFI
fileinfo
filter
gd
gettext
hash
iconv
igbinary
imagick
inotify
intl
json
ldap
libxml
mbstring
memcache
memcached
mongodb
msgpack
mysqli
mysqlnd
oneagentphp
openssl
opentelemetry
pcntl
pcre
PDO
pdo_mysql
pgsql
Phar
posix
random
rdkafka
readline
redis
Reflection
session
shmop
SimpleXML
soap
sockets
sodium
SPL
sqlite3
standard
swoole
tokenizer
xlswriter
xml
xmlreader
xmlwriter
xsl
yaml
Zend OPcache
zip
zlib
zstd

[Zend Modules]
Zend OPcache
```

php -i | grep ssl Output:
```
Registered Stream Socket Transports => tcp, udp, unix, udg, ssl, tls, tlsv1.0, tlsv1.1, tlsv1.2, tlsv1.3
openssl
Openssl default config => /etc/ssl/openssl.cnf
openssl.cafile => no value => no value
openssl.capath => no value => no value
openssl => OpenSSL 3.6.0 1 Oct 2025
SSL_CERT_FILE => /etc/ssl/certs/ca-certificates.crt
$_SERVER['SSL_CERT_FILE'] => /etc/ssl/certs/ca-certificates.crt
$_ENV['SSL_CERT_FILE'] => /etc/ssl/certs/ca-certificates.crt
```

henderkes:
Is ssl missing or sasl_ssl? ssl should exist.

henderkes:
Oh never mind, we build with --disable-ssl.

henderkes:
```php
[...]

PR #955, opened 2025-11-03
Use cmake to build librdkafka, add helper function for SPCConfigUtil

henderkes:
## What does this PR do?
closes https://github.com/crazywhalecc/static-php-cli/issues/954

### DIFF
--- a/src/SPC/builder/Extension.php
+++ b/src/SPC/builder/Extension.php
@@ -516,11 +516,6 @@ class Extension
             }
         }
 
-        if (array_key_exists(0, $deps)) {
-            $zero = [0 => $deps[0]];
-            unset($deps[0]);
-            return $zero + $deps;
-        }
         return $deps;
     }
 
@@ -552,7 +547,7 @@ class Extension
             }
             logger()->info("enabling {$this->name} without library {$name}");
         } else {
-            $this->dependencies[] = $depLib;
+            $this->dependencies[$name] = $depLib;
         }
     }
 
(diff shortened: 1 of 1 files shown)


crazywhalecc:
Looks like we don't need to patch or pass redundant variables to librdkafka if we use cmake build. I'll push cmake version later. And btw I added some helper function in SPCConfigUtil to make it more convenient to use.

crazywhalecc:
Feel free to merge if it is good.

src/SPC/builder/unix/library/librdkafka.php
```diff
@@ -26,42 +26,18 @@ public function patchBeforeBuild(): bool
 
     protected function build(): void
     {
-        UnixAutoconfExecutor::create($this)
-            ->appendEnv(['CFLAGS' => '-Wno-int-conversion -Wno-unused-but-set-variable -Wno-unused-variable'])
```
henderkes, line 30: please test with zig and clang on linux, both failed witho -Wno-int-conversion last time

src/SPC/builder/Extension.php
```diff
@@ -486,18 +486,51 @@ public function isBuildStatic(): bool
...
+                    }
+                }
+                if (!array_key_exists($depName, $deps)) {
+                    $deps[$depName] = $dep;
+                }
+            }
+        }
+
+        if (array_key_exists(0, $deps)) {
+            $zero = [0 => $deps[0]];
```
henderkes: what is this for?

crazywhalecc: Oops, it's a previous buggy code workaround

src/SPC/builder/unix/library/librdkafka.php
crazywhalecc: Just tested with `native-native-musl` and `native-native-gnu.2.17` target. It works well.

PR #956, opened 2025-11-04
Add google analytics for docs

crazywhalecc:
## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`


henderkes:
Don't we need to get the users consent before letting google track them? There are quite stringent laws here.

crazywhalecc:
This is a completely static page, so there's no data to collect. The only thing I want to know is the actual number of times the docs has been accessed. BTW I just found it could be done enough by CF proxies for our current situation. Let me close this.

> Don't we need to get the users consent before letting google track them? There are quite stringent laws here.

True. But I have not encountered similar cases for: non-commercial individuals and open-source projects without any ads, and the laws of various countries only apply to companies. I'm not sure, AFAIK this kind of regulation should only apply to the EU, Mainland China and some specific region though.

PR #966, opened 2025-11-11
remove SPC_CMD_VAR_PHP_CONFIGURE_CFLAGS, use eu-strip

henderkes:
## What does this PR do?

* attention *
targets no-strip branch!

- removes SPC_CMD_VAR_PHP_CONFIGURE_CFLAGS and uses SPC_CMD_VAR_PHP_MAKE_CFLAGS instead
- use FRANKENPHP_VERSION and FRANKENKPHP_SOURCE_PATH to not require downloading source (for people developing frankenphp)
- use eu-strip if available (doesn't remove interpreter name)
- sort configs
- remove -w -s from frankenphp build

## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`

### DIFF
--- a/src/SPC/builder/unix/UnixBuilderBase.php
+++ b/src/SPC/builder/unix/UnixBuilderBase.php
@@ -128,9 +128,6 @@ abstract class UnixBuilderBase extends BuilderBase
     {
         logger()->debug('Deploying binary from ' . $src . ' to ' . $dst);
 
-        // UPX for linux
-        $upx_option = (bool) $this->getOption('with-upx-pack', false);
-
         // file must exists
         if (!file_exists($src)) {
             throw new SPCInternalException("Deploy failed. Cannot find file: {$src}");
@@ -152,13 +149,14 @@ abstract class UnixBuilderBase extends BuilderBase
         $this->extractDebugInfo($dst);
 
         // strip
-        if (!$this->getOption('no-strip', false)) {
+        if (!$this->getOption('no-strip')) {
             $this->stripBinary($dst);
         }
 
-        // Compress binary with UPX if needed (only for Linux)
+        // UPX for linux
+        $upx_option = $this->getOption('with-upx-pack');
         if ($upx_option && PHP_OS_FAMILY === 'Linux' && $executable) {
-            if ($this->getOption('no-strip', false)) {
+            if ($this->getOption('no-strip')) {
                 logger()->warning('UPX compression is not recommended when --no-strip is enabled.');
             }
             logger()->info("Compressing {$dst} with UPX");


src/SPC/builder/unix/UnixBuilderBase.php
```diff
@@ -105,9 +105,16 @@ public function extractDebugInfo(string $binary_path): string
...
-                ->exec("objcopy --add-gnu-debuglink={$debug_file} {$binary_path}");
+            $has_eu_strip = shell()->execWithResult('which eu-strip')[0] === 0;
+            if ($has_eu_strip) {
+                shell()
+                    ->exec("eu-strip -f {$debug_file} {$binary_path}")
+                    ->exec("objcopy --add-gnu-debuglink={$debug_file} {$binary_path}");
+            } else {
+                shell()
+                    ->exec("objcopy --only-keep-debug {$binary_path} {$debug_file}")
+                    ->exec("objcopy --add-gnu-debuglink={$debug_file} {$binary_path}");
```
crazywhalecc: Use `SystemUtil::findCommand('eu-strip')` to find command.

And extract the same execution.

src/SPC/builder/unix/UnixBuilderBase.php
```diff
@@ -376,7 +385,11 @@ protected function processFrankenphpApp(): void
 
     protected function getFrankenPHPVersion(): string
     {
-        $goModPath = SOURCE_PATH . '/frankenphp/caddy/go.mod';
+        if ($version = getenv('FRANKENPHP_VERSION')) {
+            return $version;
+        }
+        $frankenphpSourceDir = getenv('FRANKENPHP_SOURCE_PATH') ?: SOURCE_PATH . '/frankenphp';
+        $goModPath = $frankenphpSourceDir . '/caddy/go.mod';
```
crazywhalecc, line 390: Is there any point of these new introduced env vars? In which scenarios would they be used?

henderkes: essentially for frankenphp developers, so they can use the current frankenphp source directory instead of having spc download it

henderkes: I would prefer not to use the FRANKENPHP_VERSION variable, but frankenphp source currently expects it to be defined by CGO_CFLAGS. It's not optimal :/

src/SPC/builder/unix/UnixBuilderBase.php
henderkes: Thanks, updated.

src/SPC/builder/unix/UnixBuilderBase.php
henderkes: see also: https://github.com/php/frankenphp/issues/1958

crazywhalecc: > essentially for frankenphp developers, so they can use the current frankenphp source directory instead of having spc download it

We currently have a source type `local`, but there's no option to specify a custom `local`. I'll think about how to handle that tomorrow.

crazywhalecc: I'm currently refactoring the artifact part of version 3.0, and I don't really want to add more new APIs/options to version 2.x, so this workaround is acceptable for me, since there's currently no way to customize the `local` source.

henderkes: Yeah I don't mind the FRANKENPHP_SOURCE_PATH, that part seems alright to me. It's the FRANKENPHP_VERSION that I'm not a fan of, but I have no idea how to mark builds as development builds rather than stable releases otherwise.

PR #968, opened 2025-11-18
add extra libs for curl (psl, krb5, unistring, idn2)

henderkes:
## Checklist before merging

  - [x] `composer cs-fix`
  - [x] `composer analyse`
  - [x] `composer test`
  - [x] Add your test combination to `src/globals/test-extensions.php`.

### DIFF
--- a/src/SPC/builder/unix/library/krb5.php
+++ b/src/SPC/builder/unix/library/krb5.php
@@ -11,6 +11,7 @@ trait krb5
 {
     protected function build(): void
     {
+        $origin_source_dir = $this->source_dir;
         $this->source_dir .= '/src';
         shell()->cd($this->source_dir)->exec('autoreconf -if');
         $libs = array_map(fn ($x) => $x->getName(), $this->getDependencies(true));
@@ -51,5 +52,6 @@ trait krb5
             'mit-krb5.pc',
             'gssrpc.pc',
         ]);
+        $this->source_dir = $origin_source_dir;
     }
 }


henderkes:
okay, musl doesn't have python?

other one just needs LD_LIBRARY_PATH.

crazywhalecc:
I remember that the reason I didn't add these dependencies to curl in the beginning was because it had a library dependency on Python. Dependent on a language with various system-level dependency issues can be quite a headache.

henderkes:
Yeah I think I'll remove psl. Looks like python is no longer default in Ubuntu either.

config/source.json
```diff
@@ -373,16 +372,6 @@
...
-    "frankenphp": {
-        "type": "ghtar",
-        "repo": "php/frankenphp",
-        "prefer-stable":  true,
-        "provide-pre-build": false,
-        "license": {
-            "type": "file",
-            "path": "LICENSE"
-        }
-    },
```
crazywhalecc, line 385: Revert?

src/SPC/builder/unix/library/krb5.php
```diff
@@ -0,0 +1,55 @@
...
+
+use SPC\util\executor\UnixAutoconfExecutor;
+use SPC\util\SPCConfigUtil;
+
+trait krb5
+{
+    protected function build(): void
+    {
+        $this->source_dir .= '/src';
+        shell()->cd($this->source_dir)->exec('autoreconf -if');
```
crazywhalecc, line 16: If there is no need to modify the root directory, it is best to pass the subpath directly instead of modifying it.

config/source.json
henderkes: It's up 200 lines, removing the duplicate here.

src/SPC/builder/unix/library/krb5.php
henderkes: How do we pass a subpath directly? I tried searching for it, but couldn't find it.

crazywhalecc: I mean `shell()->cd("{$this->source_dir}/src")->exec()` instead of modifying it.

henderkes: Not working for UnixAutoconfExecutor, that's why I changed it.

crazywhalecc: Ah that makes sense. But I think it would be more appropriate to add an arg for executor. I would make changes in 3.0 rather than this PR. I will test later to see if it affects other functions, or we can restore it at the end.

PR #976, opened 2025-11-26
Rename static-php-cli to StaticPHP

crazywhalecc:
## What does this PR do?

Prepare for future brand promotion :/

I have reservations about renaming and moving the repository, as GitHub will automatically redirect to the new repository, so it may not be difficult. But I think it's not too late to do it after version 3.0.

## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`

### DIFF
--- a/README-zh.md
+++ b/README-zh.md
@@ -1,24 +1,24 @@
-# static-php-cli
+# StaticPHP
 
-[![English readme](https://img.shields.io/badge/README-English%20%F0%9F%87%AC%F0%9F%87%A7-moccasin?style=flat-square)](README.md)
 [![Chinese readme](https://img.shields.io/badge/README-%E4%B8%AD%E6%96%87%20%F0%9F%87%A8%F0%9F%87%B3-moccasin?style=flat-square)](README-zh.md)
+[![English readme](https://img.shields.io/badge/README-English%20%F0%9F%87%AC%F0%9F%87%A7-moccasin?style=flat-square)](README.md)
 [![Releases](https://img.shields.io/packagist/v/crazywhalecc/static-php-cli?include_prereleases&label=Release&style=flat-square)](https://github.com/crazywhalecc/static-php-cli/releases)
 [![CI](https://img.shields.io/github/actions/workflow/status/crazywhalecc/static-php-cli/tests.yml?branch=main&label=Build%20Test&style=flat-square)](https://github.com/crazywhalecc/static-php-cli/actions/workflows/tests.yml)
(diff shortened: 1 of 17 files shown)


henderkes:
static-php instead of StaticPHP? Not sure, but StaticPHP looks unexpected.

crazywhalecc:
StaticPHP appears to be more SEO-friendly. Using it as the brand's display name (while the internal names and IDs are still static-php and static-php-cli) is, I guess, a relatively safe solution.

henderkes:
I think SEO friendly is not a thing we need to care about. My concern is that we already have static-php.dev, my packages link to it and people already know the project as static-php(-cli).

crazywhalecc:
Haha, actually I personally prefer calling it `spc` and `static-php`, but `static-php` with a hyphen feels more like a static word for PHP than a brand name. `Static PHP` with a space doesn't feel as good as `static-php`, but it's not elegant enough. I'm intentionally referencing the syntax of FrankenPHP, CakePHP, and NativePHP etc, so I'm thinking of doing it this way (brand name is Camel Case and internal is lowercase).

Of course, given that our project is already well-known, I think keeping `static` and `php` is fine, ultimately. The name people call it will depend on the repository name and the document title.

henderkes (approved):
> FrankenPHP, CakePHP, and NativePHP

That's a fair point. Is staticphp.dev for the grabs?

crazywhalecc:
> That's a fair point. Is staticphp.dev for the grabs?

Now I've owned `static-php.dev` and `staticphp.sh` (not used but no idea how to use and I don't plan to renew my subscription for now), `staticphp.dev` is available for now.  I could have it and redirect to static-php.dev or vice versa.

henderkes:
> > That's a fair point. Is staticphp.dev for the grabs?
> 
> Now I've owned `static-php.dev` and `staticphp.sh` (not used but no idea how to use and I don't plan to renew my subscription for now), `staticphp.dev` is available for now. I could have it and redirect to static-php.dev or vice versa.

When we're switching to `StaticPHP` it should probably be `static-php.dev` -> `staticphp.dev`.

crazywhalecc:
Since we need more time to finish v3 version of SPC, I think we could update it and move the repo when we published the first new beta version, and announce it soon. Mainly because we need to reorganize the documents and README.

henderkes:
Agreed, let's keep the branch open, though.

crazywhalecc:
We need to update all names on v3 branch.

Issue #964, opened 2025-11-06
[RFC] 3.0 Updates - Friendly Console


crazywhalecc:
# [RFC] 3.0 Updates - Friendly Console

The console output and logger is one of the most important components for SPC tool. But currently we have very green output and many redundant terminal shell outputs.

In 3.0, it will be refactored in several ways:

- Deprecate `--debug` command, use symfony's original `-v` and `-vv` for outputs. (**For backward compatibility, it will be equal to `-vvv`** for the first major version of SPC)
- Separate outputs for 4 levels:
  - Normal interactive mode: Which only shows `⚙️ Build configuration:` the necessary info for current status with  some progress bar.
  - `-v` mode: It shows build time and INFO level like current version, but with colored `[TIME <green>INFO</green>]
  - `-vv` mode: It shows all level of PSR logs, but without execution output.
  - `-vvv` mode: It shows all level of PSR logs, and output all console output for executed commands.

When using `-v` mode, the interactive message will be shown as `INFO` level, and the progress indicator will be disabled.

Example of normal mode displays:

```
⚙️ Build Configuration:
    OS: Darwin (arm64)
    ...

▶ Downloading artifacts... (90.2s)
▶ Extracting sources... (3.2s)
▶ Building libraries...
  ✓ libiconv (21.5s)
  ✓ zlib (1.1s) // finished status
  • libxml2 ... (6.6s) // wip status
▶ Configuring PHP... (22.0s)
▶ Compiling PHP... (20.3s)

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
✓ BUILD SUCCESSFUL (80.2s)
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

▶ php-cli binary: /path/to/buildroot/bin/php
▶ License: /path/to/buildroot/license
```

If you have any question or have better suggestions, please leave comment here with quote.

crazywhalecc:
<img width="672" height="185" alt="Image" src="https://github.com/user-attachments/assets/83357b2a-b89d-434b-b6a2-0df54ea3ca12" />

Demo for download command

henderkes:
Very cool! We should also display version.

crazywhalecc:
> We should also display version.

Version getter looks tricky for now. We need to implement remote version getter, source version getter and binary version getter for each package. But git and github sources could be done in downloader functions to reduce workload.

crazywhalecc:
[...]

Issue #963, opened 2025-11-05
[RFC] 3.0 Updates - Parallel Building and Downloading


crazywhalecc:
# [RFC] 3.0 Updates - Parallel Building and Downloading

static-php-cli 2.x has downloader for all sources, but downloading them one by one is quite slow.

It's best to support parallel downloads by default:

1. Calculating download artifacts
2. Queue download sources
3. Use fiber to call curl or git

For building, it's more complicated than downloading, since it depends on the dependency tree structure calculated by builder input, and `make` command always use `-j$(nproc)` to use all cores simuteniously, we may only reduce several minutes when parallel configuring.

If you have any question or have better suggestions, please leave comment here with quote.

henderkes:
For downloading yes, for sure. For building it will not increase build times, only save a bit of time for ./configure commands. I think it's not worth it, because it makes build errors less reproducible.

crazywhalecc:
I'm working on parallel downloading, but it's a little tricky. Introducing an event loop would simplify things considerably, I don't think it's necessary to add dependencies though

henderkes:
We could use swoole or uv for this. People who don't install the extensions can use the prebuilt spc binary, after all.

jingjingxyk:
config 目录下的配置信息，组成行为树，遍历行为树的节点，每个树节点执行命令，并记录执行过程 和 结果到节点


### 让多条命令  并行 或者 并发 运行，可从如下角度进行选取
1. 多进程 
2. 多线程
3. 协程
4. 队列

需要解决： 线程间、进程间通信 

### 控制运行逻辑：
1. 遍历行为树的节点
2. 行为树的节点使用状态机控制命令执行





如果不限制编程语言，可选项可多了，如果限制在 PHP 编程语言，swoole workerman swow 等 实现事件循环的框架

boonkerz:
parallel extension is also an option

crazywhalecc:
Okay. Good news: we can make shell execution like coroutine using fiber pool with very small changes and without any extensions or dependencies !

```php
 // fill pool
while (count($fiber_pool) < $pool_count && ($artifact = array_shift($this->artifacts)) !== null) {
    $current = $count - count($this->artifacts);
    $fiber = new \Fiber(function () use ($artifact, $current, $count) {
        $this->downloadWithType($artifact, $current, $count, true);
    });
    $fiber->start();
    $fiber_pool[] = $fiber;
}
// check pool
[...]

Issue #962, opened 2025-11-05
[RFC] 3.0 Updates - Zig Support


crazywhalecc:
# [RFC] 3.0 Updates - Zig Support

In current 2.x version, we regard `zig` as an alternative build toolchain. Based on our previous ideas, we will use ZigToolchain as the default build toolchain for Linux in version 3.0. This process will be virtually seamless for the normal build workflow, requiring almost no modifications.

However, given the current stability of the toolchain, whether to retain support for the old toolchain in version 3.0 is a big question mark.

Because gcc/clang has stood the test of time and solutions can be found for most situations, while ziglang is still in version 0.x, but SPC currently uses it almost perfectly, so we're unsure whether to officially abandon other build solutions.

This is because, to date, SPC, whether used to build PHP, extensions, or other packages, has aimed to build portable binaries; we won't invest too much effort in building ordinary, dynamically linked, and non-portable binaries.

If you have any question or have better suggestions, please leave comment here with quote.

henderkes:
We can't abandon gcc because of the performance in php < 8.5.

We could deprecate clang, but there's not really a point as everything clang needs is covered with our zig support.

boonkerz:
currently the zig linker can not handle thin archive.
so i have issues to build wayland for my sdl project

i not know if any other extension is affected.

https://github.com/ziglang/zig/issues/25694

crazywhalecc:
@henderkes What are your thoughts on linux default toolchain for v3? This could impact the module refactoring of our hosted pre-built libraries (and drop SPC_LIBC right?)

henderkes:
Optimally, default for 8.5+ is zig, default for <=8.4 should remain gcc.

For dropping SPC_LIBC I think that's okay. If someone wants musl binaries compiled with gcc they'll just have to use alpine.

henderkes:
@crazywhalecc I think we should switch the macOS default to Zig as well. We actually only need to support Zig on macOS.

Advantages:
- no need to install a ton of brew packages, which takes time
- we get an up-to-date clang toolchain (would fix current tailcall clang crashes)
- we don't need to support older versions of clang or macos, because zig runs the same no matter where you run it and ignores system libraries

Disadvantages:
- none?

Issue #961, opened 2025-11-05
[RFC] 3.0 Updates - Vendor Mode Support


crazywhalecc:
# [RFC] 3.0 Updates - Vendor Mode Support

For long time, the SPC can only be used as a standalone tool or complete CLI project for PHP, and there are almost no scalable ways and integration with existing PHP projects is not very good. Every project depends on SPC needs to fork, or clone, or download spc binary and call it using shell command.

In 3.0, we will support vendor mode for most PHP projects. That is, you can use `composer require crazywhalecc/static-php-cli`, and call SPC functions using PHP, or expand the usage of SPC, e.g. adding private extension support, re-use static lib packages for building other open-source apps, or even publishing your own rpm/deb packages, etc.

Regarding this part, there is no way to implement it yet, but in general, it is based on the following parts:

- Skeleton command: allows you to create your own package with ease.
- Custom command registry: expand the `spc` command ability.
- Expose calling helper function: integrate build, install, patch process in your workflows or projects.

If you have any question or have better suggestions, please leave comment here with quote.

crazywhalecc:
@henderkes 

I'm not entirely sure which interfaces should be exposed in vendor mode, but you're familiar with the spc-packages workflow. I haven't looked at the packages code clearly, and I'd like to ask.

Taking building release packages as an example, in the build pipeline, which form of injection do you think is best for pre-operations and post-stages?

- Now we can register new commands in the external registry, running as the `vendor/bin/spc <do-sth>` command.
- Now we have `#[AfterStage('<package-name>', 'build')]` (and a before version), allowing us to declare events externally using annotations.
- Now we can directly call `PackageInstaller->run()` in vendor mode to trigger the package download, installation, and build process, including shared PHP extensions, smoke tests, etc.


If only considering spc-packages, which method of embedding the workflow do you prefer? Annotation, PHP call, or another new command to combine them from outside scripts?

henderkes:
>Now we can register new commands in the external registry, running as the vendor/bin/spc <do-sth> command.

That'll be helpful for the package step. Currently the repo does two steps, build (first) and package (second). build essentially just calls `spc craft`.

>Now we have #[AfterStage('<package-name>', 'build')] (and a before version), allowing us to declare events externally using annotations.

That will be very helpful for the package related patches, e.g. php-config and other embed-packages have to be patched so they can point to system locations.

[...]

PR #980, opened 2025-11-30
StaticPHP v3

crazywhalecc:
## Important notes

STATUS: dev -> **`alpha`** -> beta -> rc -> stable

Contribution notes: 

- Before we released the first `alpha` version, any changes should be merged into `v3-dev`, not `v3`.
- After we released the first `alpha` version, any changes should be merged into `v3` and we will deprecate `v3-dev` branch.
- The structure of StaticPHP v3 may be changed before `beta` version.
- Publishing new RFC is welcome.

## RFCs

- #959 
- #960 
- #961 
- #962 
- #963
- #964 
- #985 

### `bin/` refactoring

- [X] `bin/spc` for v3
- [x] `bin/spc-alpine-docker`
- [ ] `bin/spc-gnu-docker`
- [x] `bin/docker-entrypoint.sh`
- [x] `bin/setup-runtime`

### v3 new modules

- [X] **Registry** module loading for all configs and packages (#980)
- [ ] **Vendor mode** support
  - [X] Separate a dedicated `bootstrap` file for initialization (#980) (src/bootstrap.php)
  - [ ] Add `init-registry` command for vendor user

### Commands refactoring

- [X] **ConsoleApplication** dynamic generating commands for `target` type packages (#980)
- [X] **Custom commands** in external registry (#980)

- [X] DownloadCommand
  - [X] **shallow-clone** by default (#980)
  - [X] **clean** action (#980)
  - [X]  `--from-zip`, `--all`, `--with-openssl11` have been removed (#980)
  - [X] New options: `--without-suggests`, `--for-packages`, `--ignore-cache`, `--parallel`, `--no-shallow-clone` (#980)
  - [X] Porting common options: `--with-php`, `--clean`, `--custom-url`, `--custom-git`, `--no-alt`, `--with-php` (#980)
  - [X] Porting v2 compatible options: `--for-extensions`, `--for-libs`, `--without-suggestions`, `--ignore-cache-sources`, `--prefer-pre-built`  (#980)
- [X] BaseCommand
  - [X] New verbosity (#980)
  - [X] Git version display (#980)
- [x] BuildLibsCommand
  - [X] Base function of building libs (#980)
  - [ ] **clean** action
  - [X] Remove `--all` option for `build:libs` (#980)
- [ ] BuildPHPCommand -> BuildTargetCommand with `php` package
  - [ ] Options
    - [X] `--with-libs`: Keep for compatibility, add `--with-packages` option (#980)
    - [x] `--build-shared`
    - [X] `--build-all`: Special parsing in v3 (#980)
    - [X] `--no-strip`, `--disable-opcache-jit`, `--build-{sapi}`, `--with-config-file-path`, `--with-config-file-scan-dir`, `--with-hardcoded-ini`, `--with-micro-fake-cli`, `--with-added-patch (-P)`, `--without-micro-ext-test`, `--with-upx-pack`, `--with-micro-logo`, `--enable-micro-win32`, `--with-frankenphp-app` has been moved to `php` package definition (#980)
    - [ ] Implementation status
      - [x] `no-strip`
      - [x] `disable-opcache-jit`
      - [x] `with-config-file-path`
      - [x] `with-config-file-scan-dir`
      - [x] `with-hardcoded-ini`
      - [x] `with-micro-fake-cli`
      - [ ] `with-addeed-patch`
      - [ ] `without-micro-ext-test`
      - [x] `with-upx-pack`
      - [x] `with-micro-logo`
      - [x] `enable-micro-win32`
      - [x] `with-frankenphp-app`
      - [x] `build-shared`
      - [x] `with-packages` (`with-libs`)
      - [ ] `build-all`
      - [X] `build-{sapi}` except `build-frankenphp` (#980)
      - [x] `build-frankenphp`
      - [x] `with-suggests` (`with-suggested-libs`, `with-suggested-exts`)
      - [X] `no-download` (#980)
      - [X] Download options in build command (#980)
      - [ ] `with-clean`
  - [X] Build information printing: refactored to attribute `#[StaticPHP\Attribute\Package\Info]` (#980)
  - [x] Bind build extra info to ExceptionHandler
  - [X] Validate packages before building and after extracting (#980)
  - [X] Patch static-php-cli version to PHP has been moved to `php` package `build` stage before (#980)
  - [x] Display build PHP result in terminal
  - [x] Export licenses and build extensions and lib info after building
- [x] CraftCommand
- [ ] DeleteDownloadCommand
- [X] DoctorCommand (#980) (Just command, not Doctor items)
- [x] DumpExtensionsCommand
- [x] DumpLicenseCommand
- [X] ExtractCommand (Just command, not Extractor)
  - [X] Extractor calling
  - [X] Options
    - [X] New `for-extensions` (#980)
    - [X] New `for-libs` (#980)
    - [X] New `for-packages` (#980)
    - [X] New `without-suggests` for `for-*` options (#980)
- [x] InstallPkgCommand ->InstallPackageCommand
  - [X] PackageInstaller calling
  - [ ] Options
    - [ ] `skip-extract`
- [x] MicroCombineCommand
- [X] SPCConfigCommand
  - [X] SPCConfigUtil calling
  - [X] Options
- [X] SwitchPhpVersionCommand
- [x] Dev commands
  - [ ] AllExtCommand
  - [x] EnvCommand
  - [ ] ExtVerCommand
  - [ ] GenerateExtDepDocsCommand
  - [ ] GenerateExtDocCommand
  - [ ] GenerateLibDepDocsCommand
  - [ ] LibVerCommand
  - [x] PackLibCommand
  - [ ] PhpVerCommand
  - [x] SortConfigCommand

#### Doctor module

- [X] **Core framework refactored** (#980)
  - [X] Doctor main class
  - [X] DoctorLoader for loading check items
  - [X] CheckResult for storing check results
  - [X] Attribute-based check system: #[CheckItem], #[OptionalCheck], #[FixItem]
- [x] **Doctor check items**
  - [X] OSCheck (migrated from v2's OSCheckList)
  - [X] Re2cVersionCheck
  - [X] PkgConfigCheck
  - [X] MacOSToolCheck (migrated from v2's MacOSToolCheckList)
  - [X] LinuxMuslCheck
  - [x] LinuxToolCheck (v2 has LinuxToolCheckList)
  - [ ] BSDToolCheck (v2 has BSDToolCheckList)
  - [x] WindowsToolCheck (v2 has WindowsToolCheckList)
  - [x] ZigCheck (v2 has ZigCheck)

### Legacy modules (v2 only - SPC namespace)

These modules exist in v2 but are refactored/merged into new v3 modules:

- [X] **store/** - Deprecated and refactored
  - [X] Downloader → StaticPHP\Artifact\ArtifactDownloader
  - [X] LockFile → StaticPHP\Artifact\ArtifactCache
  - [X] CurlHook → Merged into StaticPHP\Artifact\ArtifactDownloader
  - [X] FileSystem → StaticPHP\Util\FileSystem
  - [X] DirDiff → StaticPHP\Util\DirDiff
  - [X] SourcePatcher → StaticPHP\Util\SourcePatcher (partial)
  - [X] PackageManager, SourceManager → Removed (functionality in new Package/Artifact systems)
  - [X] pkg/ → Migrated to src/Package/
  - [X] source/ → Migrated to Artifact system


src/StaticPHP/Command/InstallPackageCommand.php
```diff
@@ -8,19 +8,19 @@
...
-        $this->addArgument('package', null, 'The package to install (name or path)');
+        $this->addArgument('packages', null, 'The package to install (name or path)');
     }
 
     public function handle(): int
     {
         ApplicationContext::set('elephant', true);
         $installer = new PackageInstaller([...$this->input->getOptions(), 'dl-prefer-binary' => true]);
-        $installer->addInstallPackage($this->input->getArgument('package'));
+        $installer->addInstallPackage($this->input->getArgument('packages'));
```
henderkes: we still only allow installing one at a time, don't we?

crazywhalecc: Theoretically, yes. Perhaps in the future we will need to implement multiple parameters similar to other package managers, rather than a single comma-separated parameter.

crazywhalecc:
> We should chat about what features we no longer want to support and what we want to rework.

@henderkes  **There's no urgent things.** Just when you return from your vacation, and if you have time, it would be best we could discuss what modifications are still needed for the current framework of v3. Please feel free to contact me if you'd like.

Currently I've finished the major part of framework according to RFC, but I haven't ported most of the implementation yet, mainly to prevent major changes to the framework structure from requiring massive modifications.

And for anyone: If you have any feedback on the framework design, please let me know.

henderkes:
I haven't had much of a look yet and the changelist is quite massive, so it might be a good idea to chat about it one evening after I had a rough glance. I'll give everything (especially vendor mode) a play and rewrite the static-php/packages properly on top of the new branch.

Might have to add versioned packages to it first though, because that's the part still holding up apk and proper apt support.

crazywhalecc:
For a quick reference for command mode, in the current branch, the available commands:

- `bin/spc doctor` on Windows, macOS, Linux (linux is not fully tested)
- `bin/spc build:libs libedit` on *nix
- `bin/spc build:php-cli readline` on *nix
- `bin/spc build bcmath --build-cli --debug` (`--build-XXX` and `--debug` has been marked as deprecated)
- `bin/spc build:libs onig` on Windows
- `bin/spc download {some packages}`
- `bin/spc download --for-extensions={some extensions}`
- `bin/spc extract {some package}`
- `bin/spc install-pkg {some package}`
- `bin/spc spc-config {some extensions}` (still using extension as argument)

The remaining commands and some of the option implementations may not have been tested.

crazywhalecc:
> Might have to add versioned packages to it first though

You mean the dependencies' version from artifacts?

henderkes:
> > Might have to add versioned packages to it first though
> 
> You mean the dependencies' version from artifacts?

No, I mean versioned .deb packages like php-zts*8.3*-... Nothing on the spc side.

> bin/spc build bcmath --build-cli --debug (--build-XXX and --debug has been marked as deprecated)

Why is --build-cli deprecated? What would be the new way to build multiple SAPIs at once?

crazywhalecc:
> Why is --build-cli deprecated? What would be the new way to build multiple SAPIs at once?

I'm thinking of something like `sapis` arguments for `build:php cli,fpm {extensions}` command, not original `build --build-XXX` command. But it needs to separate `build` and `build:php` command, it will break compatibility.

I haven't decided any changes yet, but anyway we should current deprecate `build` command behavior. It should be building pacakge, not building php with extensions.

henderkes:
For gnu docker, do you think we should update to alma 8 or drop it entirely and rely on zig?

crazywhalecc:
> For gnu docker, do you think we should update to alma 8 or drop it entirely and rely on zig?

To be honest, if conditions permit, I will try to avoid using Docker anyway. But whether to keep gnu-docker depends on whether we absolutely have to use it to build anything. I prefer to avoid using Docker, especially when Zig can almost completely cover it.

> The other day, while using Docker on my Mac, engine crashed and a factory reset button clicked, which has no double-checking, caused me to lose all the data in all my Docker volumes, containers, images at all. 
>
> Some of the software I need for my work is developed in Java web, but they are deployed using Docker on ordinary people's computers. Solving Docker problems takes up almost all of their customer service time.

henderkes:
We could drop Docker entirely. Zig also supports building musl-linked binaries, both static and dynamic. It's just that it produces slow binaries on php < 8.5.

crazywhalecc:
# static-php-cli 3.0.0-alpha1 Release

static-php-cli v3 is a ground-up rewrite of the project, bringing a modern PHP 8.4+ architecture, dramatically improved user experience, and a flexible plugin system. This release represents months of work reimagining how static PHP binaries should be built.

## Highlights

### Concurrent Parallel Downloads

The download system now supports **fiber-based parallel downloads** out of the box. Use `--parallel N` to download multiple artifacts concurrently, significantly reducing the time spent fetching sources and pre-built binaries.

### Improved Terminal Output

Terminal output has been completely redesigned. Build progress is now color-coded with clear status indicators, structured indentation for build stages, and smart verbosity levels (`-v`, `-vv`, `-vvv`). Log noise is reduced by default — only actionable information is shown, while full logs are written to file when needed.

### Check for Updates

A new `check-update` command lets you check whether newer versions of any artifact (library, extension source, PHP release) are available, with optional `--json` output for CI integration.

### Developer Tooling

New `dev:*` commands provide deep introspection into the build system:

- `dev:dump-stages` — Visualize all build stages and their hooks
- `dev:dump-capabilities` — Export installable/buildable capabilities as JSON
- `dev:package-info` — Inspect package metadata and dependencies
- `dev:lint-config` — Validate configuration files
- `dev:shell` — Interactive shell inside the build environment
- `dev:is-installed` — Query whether a package has been built

---

## Windows

### FrankenPHP and embed SAPI

v2 had no Windows support for FrankenPHP, and embed SAPI on Windows was marked as incomplete. v3 fully implements both: `php-embed` builds a genuine static library (`phpNNembed.lib`) on Windows, and FrankenPHP can be cross-compiled for Windows using the LLVM/Clang toolchain bundled with Visual Studio. PE metadata (version info and icon) is embedded automatically during the build as part of the standard release workflow.

### Expanded Extension Support

v3 adds explicit Windows build support for 13+ external/PECL extensions, including `amqp`, `mongodb`, `yaml`, `msgpack`, `parallel`, `lz4`, `zstd`, `xlswriter`, and more. Combined with the 56 built-in extensions, Windows builds now have access to the same broad extension ecosystem available on Linux and macOS.

---

## Architecture

### PHP 8.4 Attribute-Driven Build System

v3 replaces the v2 inheritance-based builder hierarchy with a declarative attribute system built on PHP 8.4 attributes:

```php
#[Library('openssl')]
class openssl {
    #[BuildFor('Linux')]
    #[BuildFor('Darwin')]
    public function buildUnix(LibraryPackage $lib): void {
        UnixAutoconfExecutor::create($lib)->configure()->make();
    }

    #[BuildFor('Windows')]
    public function buildWin(LibraryPackage $lib): void { /* ... */ }
}
```

Platform-specific logic is selected at runtime via `#[BuildFor]`. Build lifecycle hooks (`#[BeforeStage]`, `#[AfterStage]`, `#[PatchBeforeBuild]`) allow packages to interact with each other's build stages without tight coupling.

### Dependency Injection

v3 is built on [PHP-DI](https://php-di.org/) with full autowiring support. Build services, configuration, and runtime utilities are injected automatically into commands, build classes, and stage methods via `ApplicationContext`.

### Unified Package Model

v2 maintained three separate configuration tracks: `source.json` for download sources, `lib.json` for libraries, and `ext.json` for extensions, each with their own dependency fields (`lib-depends`, `ext-depends`).

v3 unifies everything into a single per-package YAML configuration with a flat `depends` field. Libraries, PHP extensions, and build targets are all "packages" with a type discriminator:

```yaml
# config/pkg/lib/curl.yml
curl:
  type: library
  artifact:
    source:
      type: ghrel
      repo: curl/curl
      match: curl-.+\.tar\.gz
    binary: hosted
  depends:
    - openssl
    - zlib
  depends@windows:
    - openssl
    - zlib
```

Platform-specific overrides use `@` notation (`depends@windows`, `static-libs@unix`) instead of v2's suffix convention (`lib-depends-windows`, `static-libs-unix`).

### Flattened Dependency Graph

v2 required maintaining two parallel dependency trees (`lib-depends` and `ext-depends`). v3 collapses these into a single `depends` list — extensions reference libraries directly (e.g., `openssl`), and other extensions with the `ext-` prefix (e.g., `ext-openssl`). The dependency resolver performs topological sorting with virtual-target promotion, optional dependency support (`suggests`), and reverse-lookup for debugging why a package was included.

### Registry and Plugin System

v3 introduces a registry-based architecture defined in `spc.registry.yml`. The core registry declares where packages, doctor checks, artifacts, and commands are loaded from. External registries can be added via the `SPC_REGISTRIES` environment variable, enabling third-party packages, custom build logic, and vendor-specific configurations without modifying the core codebase.

### System-Independent Core

v2 used a platform-specific builder inheritance hierarchy (`LinuxBuilder` / `MacOSBuilder` / `WindowsBuilder` extending `UnixBuilderBase` / `BuilderBase`). v3 decouples the build orchestration from the OS layer entirely. Platform-specific utilities are isolated into `LinuxUtil`, `MacOSUtil`, `WindowsUtil` classes, while `PackageInstaller` and `PackageBuilder` operate system-independently.

### Direct CMake Execution

v2 required wrapper BAT scripts (`makeSimpleWrapper()`) to invoke CMake within the Visual Studio SDK environment. v3 introduces `WindowsCMakeExecutor`, which detects the Visual Studio installation directly via `vswhere.exe` and invokes CMake without any wrapper scripts, using the same executor interface as Unix builds:

```php
WindowsCMakeExecutor::new($lib)
    ->configure('-DBUILD_SHARED_LIBS=OFF')
    ->make();
```

---

## Configuration

### Per-Package YAML

Each package now has its own YAML configuration file under `config/pkg/`, organized by type (`lib/`, `ext/`, `target/`). This replaces v2's monolithic JSON files and makes it easier to review, override, and extend package definitions.

### PECL Source Type

v3 adds native PECL download support. External extensions hosted on PECL can be declared simply as:

```yaml
ext-redis:
  type: php-extension
  artifact:
    source:
      type: pecl
      name: redis
```

This eliminates the boilerplate download URL configuration required in v2.

### Environment Variables

Build environment is now managed through `config/env.ini` with support for user overrides via `config/env.custom.ini`. All build paths, toolchain settings, concurrency options, and platform detection variables are centralized and documented in one place.

### License Metadata

Package configurations now include structured license metadata (`license`, `license-files`), and a dedicated `LicenseDumper` utility collects and exports all dependency licenses for compliance.

---

## Package Ecosystem

v3 ships with **75 system libraries** and **112 PHP extension configurations** (56 built-in + 56 external/PECL), covering the same breadth as v2 with improved configuration quality.

---

## Breaking Changes

v3 is a complete rewrite. All internal APIs, configuration formats, and extension points have changed. If you are using static-php-cli as a library (`vendor` mode) or maintaining custom extensions/libraries, you will need to migrate to the new architecture. The CLI interface (`bin/spc`) remains largely compatible for end users running `craft`, `download`, `build:*`, and `doctor` commands.

Key migration points:

- Configuration migrated from JSON (`lib.json`, `ext.json`, `source.json`) to per-package YAML
- `lib-depends` / `ext-depends` merged into unified `depends`
- Platform suffixes changed from `-unix` / `-windows` to `@unix` / `@windows`
- Builder classes replaced by attribute-annotated package classes
- Custom extensions must adopt `#[Extension]` attribute and new lifecycle hooks

crazywhalecc:
@claude[agent]+claude-opus-4.6 Review

henderkes:
Lol I can't select it here, only in issues.
@claude review this

henderkes:
@claude[agent] hello review?
@claude review please?

I'll take a while to meaningfully review this. It will be a lot of work to get through it all and switch our packages to it.

crazywhalecc:
> I'll take a while to meaningfully review this. It will be a lot of work to get through it all and switch our packages to it.

Considering that almost all parts have been refactored, if you are going to refactor packages repo, it might be faster to refer to the documentation after my documentation is complete.

image

henderkes (approved):

crazywhalecc:
:tada:

henderkes:
Something big is incoming as well... been working on it for 5 days.

crazywhalecc:
I have to release v3-alpha next week. I'll take a short five-day vacation.

config/env.ini
```diff
@@ -94,7 +94,7 @@ SPC_MICRO_PATCHES=static_extensions_win32,cli_checks,disable_huge_page,vcruntime
 ; `native-native`               - links against system libc dynamically
 ; `native-native-musl`          - links against musl libc statically
 ; `native-native-musl -dynamic` - links against musl libc dynamically
-SPC_TARGET=native-native-musl
+SPC_TARGET=native-native
```
crazywhalecc: We always assume that we will build musl-static by default, don't we?

config/env.ini
```diff
@@ -114,13 +114,13 @@ SPC_MICRO_PATCHES=cli_checks,disable_huge_page
...
 ; configure command
-SPC_CMD_PREFIX_PHP_CONFIGURE="./configure --prefix= --with-valgrind=no --disable-shared --enable-static --disable-all --disable-phpdbg --with-pic"
+SPC_CMD_PREFIX_PHP_CONFIGURE="./configure --prefix= --with-valgrind=no --disable-shared --enable-static --disable-all --disable-phpdbg --enable-rtld-now --enable-re2c-cgoto --disable-rpath --with-pic"
 
 ; *** default build vars for building php ***
 ; embed type for php, static (libphp.a) or shared (libphp.so)
 SPC_CMD_VAR_PHP_EMBED_TYPE="static"
 ; EXTRA_CFLAGS for `configure` and `make` php
-SPC_CMD_VAR_PHP_MAKE_EXTRA_CFLAGS="-g -fstack-protector-strong -fno-ident -fPIE ${SPC_DEFAULT_C_FLAGS}"
+SPC_CMD_VAR_PHP_MAKE_EXTRA_CFLAGS="-g -fstack-protector-strong -fno-ident -fPIE -fvisibility=hidden -fvisibility-inlines-hidden ${SPC_DEFAULT_C_FLAGS}"
```
crazywhalecc: We could add `fvisibility` and `cgoto` on macOS too.

src/Package/Extension/gearman.php
```diff
@@ -0,0 +1,20 @@
...
+use StaticPHP\Package\PackageInstaller;
+
+#[Extension('gearman')]
+class gearman
+{
+    #[CustomPhpConfigureArg('Linux')]
+    #[CustomPhpConfigureArg('Darwin')]
+    public function getUnixConfigureArg(bool $shared, PackageInstaller $installer): string
+    {
+        return '--with-gearman=' . ($shared ? 'shared,' : '') . $installer->getLibraryPackage('libgearman')->getBuildRootPath();
```
crazywhalecc: Use `--with-gearman@shared_path_suffix@` in config to simplify

crazywhalecc (commented):
@henderkes

config/env.ini
henderkes: Ah yeah, I meant to talk to you on discord. I think that's ready for a change in v3. php is still better on gnu than on musl so I think the better default is to build against the system the user is on.

henderkes: FWIW we also can't use native-native here. It needs to be $arch-$os. native-native enables e.g. -march=x86-64-v3 on a cpu that supports it.

crazywhalecc: I prefer that we provide binaries with the best portability by default, and performance is just one area we're working on. We have the ability to improve performance, but that doesn't mean we should change the default build target though. 

In short: I prefer default portability while providing a configuration that achieves optimal performance.

henderkes: Okay, let me do some changes then.

config/env.ini
```diff
@@ -142,7 +142,7 @@ CXX=${SPC_DEFAULT_CXX}
 AR=${SPC_DEFAULT_AR}
 LD=${SPC_DEFAULT_LD}
 ; default compiler flags, used in CMake toolchain file, openssl and pkg-config build
-SPC_DEFAULT_C_FLAGS="--target=${MAC_ARCH}-apple-darwin -Os"
+SPC_DEFAULT_C_FLAGS="--target=${MAC_ARCH}-apple-darwin -fPIC -O3 -fno-plt -fno-semantic-interposition -fstack-clash-protection -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -ffunction-sections -fdata-sections"
```
henderkes: not sure if -fPIC is required on macOS

config/env.ini
```diff
@@ -94,15 +94,15 @@ SPC_MICRO_PATCHES=static_extensions_win32,cli_checks,disable_huge_page,vcruntime
...
+SPC_TARGET=${GNU_ARCH}-linux-musl
 
 ; compiler environments (default value is defined by selected toolchain)
 CC=${SPC_DEFAULT_CC}
 CXX=${SPC_DEFAULT_CXX}
 AR=${SPC_DEFAULT_AR}
 LD=${SPC_DEFAULT_LD}
 ; default compiler flags, used in CMake toolchain file, openssl and pkg-config build
-SPC_DEFAULT_C_FLAGS="-fPIC -Os"
+SPC_DEFAULT_C_FLAGS="-fPIC -O3 -pipe -fno-plt -fno-semantic-interposition -fstack-clash-protection -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -ffunction-sections -fdata-sections"
```
crazywhalecc: Given that our default is musl-static, I recommend using `-Os` here. We can add a comment explaining that for better performance, please use glibc + O3.

config/env.ini
crazywhalecc: Same here

config/env.ini
henderkes: Musl-static also benefits from -O3 (or at least -O2) a lot compared to -Os since 8.5 with zig, which is now our default.

henderkes: The size difference is not that large either, 2-3 mb more or less don't make a big difference on a 80mb binary.

config/env.ini
henderkes: does it need to be removed? if it's harmless I'd keep it to better convey that we need to build position independent code

henderkes: The reason I think we should use these flags everywhere is because better performance and better debuggability are sensible to have by default. The produced binaries should hit a good balance between performance, security and size. I only chose cheap hardening options here and -O3 doesn't produce much larger binaries than -Os.

crazywhalecc: I think on macOS we only need `-fvisibility=hidden -fvisibility-inlines-hidden` and `--enable-re2c-cgoto`. Other flags is unused and may cause build failure.

crazywhalecc: On my MBP M3 Pro: `bin/spc build bcmath,intl,curl,openssl,tokenizer,ftp,iconv,mysqlnd,xml,soap,imagick,mongodb --dl-parallel=20 --build-cli`

`-O3`: 78535776 - 75M - 415.5 s
`-Os`: 76452016 - 73M - 420.6 s
`-O3 -flto`: 74223280 - 71M - 1446.4 s

crazywhalecc: It does seem that `-O3` would be better, but I haven't done extensive testing, only focusing on this specific case and only for build speed and file size, not runtime performance.

lto seems to have a huge impact, causing slower builds of these sets. If the goal is simply to improve PHP speed, perhaps applying some optimizations only to `php-src` would be better.

For smaller binary files (such as micro.sfx), `-Os` has a greater advantage over O3, while O3 offers almost no significant improvement for CLI applications. But anyway as long as `-O3` doesn't cause many maintenance issues, I think it's acceptable in terms of build speed and size. By default, we only need to find a balance point. We don't need to be particularly aggressive in any particular direction.

henderkes: Yeah I'm using only -O3 for the library builds. For packages I'm not using -flto=thin, which takes ~half the time of full lto, but still ~3x longer than without lto.

With -Os we weren't always getting the hybrid/tailcall VM for some reason, so the default should be -O2 or -O3. That definitely makes a huge difference for cli applications (that actually need php performance). If someone wants a small image, they can still set SPC_DEFAULT_CFLAGS="-Os -fPIC".

henderkes: @crazywhalecc reminds me, we should change `SPC_DEFAULT_C_FLAGS` to `SPC_DEFAULT_CFLAGS` and likewise for CXX/LDFLAGS. We still can now.

crazywhalecc: `-fno-semantic-interposition -fstack-clash-protection` and `-Wl,-z,relro -Wl,--as-needed -Wl,-z,now -Wl,-z,noexecstack -Wl,--gc-sections` is preventing my building though, I just changed them to `SPC_DEFAULT_C_FLAGS=--target=${MAC_ARCH}-apple-darwin -fPIC -O3 -fno-plt -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -ffunction-sections -fdata-sections`, and `SPC_DEFAULT_LD_FLAGS=-Wl,-dead_strip`

crazywhalecc: Done. This is the simplest thing ever

henderkes: yeah im not sure with mac LD, just copied the linux ones for best effort

crazywhalecc:
I'll merge this first, and then we start preparing for the doc version switch and test suite update; this PR is too big.

henderkes:
🎉

PR #982, opened 2025-12-01
update libwebp and libxml2

### LINKED ISSUE #981: new libxslt version fails build
https://github.com/static-php/packages/actions/runs/19807726999

started today, yesterday evening was fine

henderkes:
## What does this PR do?

- Closes #981 
- automatically updates libxml2, libwebp and go-xcaddy for new versions
- fixes build of pdo_sqlsrv from git, rather than a release tarball
- fixes liburing build for non-gcc on old linux header systems (rhel 8 and older)

This was what is required to build php 8.5 extensions (with 4 overridden source urls) by static-php/packages: see https://github.com/static-php/packages/blob/c5f504b532e394bd54391fa47501d7d07c6b03ee/config/templates/craft.yml.twig#L51

### DIFF
--- a/src/SPC/builder/Extension.php
+++ b/src/SPC/builder/Extension.php
@@ -576,7 +576,7 @@ class Extension
 
     protected function getExtraEnv(): array
     {
-        return ['CFLAGS' => ''];
+        return [];
     }
 
     /**


src/SPC/builder/unix/library/libwebp.php
```diff
@@ -5,13 +5,26 @@
...
+                '-DWEBP_BUILD_ANIM_UTILS=OFF',
+                '-DWEBP_BUILD_CWEBP=OFF',
+                '-DWEBP_BUILD_DWEBP=OFF',
+                '-DWEBP_BUILD_GIF2WEBP=OFF',
+                '-DWEBP_BUILD_IMG2WEBP=OFF',
+                '-DWEBP_BUILD_VWEBP=OFF',
+                '-DWEBP_BUILD_WEBPINFO=OFF',
+                '-DWEBP_BUILD_WEBPMUX=OFF',
+                '-DWEBP_BUILD_FUZZTEST=OFF',
+                SPCTarget::getLibcVersion() === '2.31' && GNU_ARCH === 'x86_64' ? '-DWEBP_ENABLE_SIMD=OFF' : '' // fix an edge bug for debian 11 with gcc 10
```
crazywhalecc, line 26: Supporting pre-built content with optional feature is not safe though. Either use fixed features, or remove pre-built content.

src/SPC/builder/unix/library/libjxl.php
```diff
@@ -29,13 +29,17 @@ protected function build(): void
...
         if (ToolchainManager::getToolchainClass() === ZigToolchain::class) {
-            $cmake->addConfigureArgs(
-                '-DCXX_MAVX512F_SUPPORTED:BOOL=FALSE',
-                '-DCXX_MAVX512DQ_SUPPORTED:BOOL=FALSE',
-                '-DCXX_MAVX512CD_SUPPORTED:BOOL=FALSE',
-                '-DCXX_MAVX512BW_SUPPORTED:BOOL=FALSE',
-                '-DCXX_MAVX512VL_SUPPORTED:BOOL=FALSE'
-            );
+            $cflags = getenv('SPC_DEFAULT_C_FLAGS') ?: getenv('CFLAGS') ?: '';
+            $has_avx512 = str_contains($cflags, '-mavx512') || str_contains($cflags, '-march=x86-64-v4');
```
crazywhalecc, line 33: Same as libwebp

src/SPC/builder/extension/pdo_sqlsrv.php
```diff
@@ -0,0 +1,26 @@
...
+class pdo_sqlsrv extends Extension
+{
+    public function patchBeforeBuildconf(): bool
+    {
+        if (!file_exists($this->source_dir . '/config.m4') && is_dir($this->source_dir . '/source/pdo_sqlsrv')) {
+            FileSystem::moveFileOrDir($this->source_dir . '/LICENSE', $this->source_dir . '/source/pdo_sqlsrv/LICENSE');
+            FileSystem::moveFileOrDir($this->source_dir . '/source/shared', $this->source_dir . '/source/pdo_sqlsrv/shared');
+            FileSystem::moveFileOrDir($this->source_dir . '/source/pdo_sqlsrv', SOURCE_PATH . '/pdo_sqlsrv');
+            FileSystem::removeDir($this->source_dir);
+            FileSystem::moveFileOrDir(SOURCE_PATH . '/pdo_sqlsrv', $this->source_dir);
```
crazywhalecc: We could do it with extract hook. It's not patch.

crazywhalecc:
Speaking of which, I haven't set up any new command-line override rules for 3.0 yet. Besides the currently limited custom-url, custom-git, and custom-local options, what good methods do you think could be used to design a command-line overriding way?

The current version 3.0 (dev) allows users to register additional registry entries to override basic configs to let someone wants to override or expand functions on top of them. I don't know if it makes sense to continue expanding `--custom-*`, for common use.

Making another `--custom={source_name}:{json}` is very straightforward, but I don't think I'd even use it myself.

src/SPC/builder/unix/library/libwebp.php
henderkes: This specific feature set is fine with prebuilt archives. This is just a debian 11 gcc 10 bug where it's missing one single avx2 intrinsic when trying to build the library.

src/SPC/builder/unix/library/libjxl.php
henderkes: Also not an issue unless we were to build prebuilt packages with avx512 support, but then most people wouldn't be able to run them.

This exact same check is already used in every single extension, adding it for a library doesn't prevent prebuilt archives.

src/SPC/builder/extension/pdo_sqlsrv.php
henderkes: Yes, agreed, that's better.

henderkes:
I would say we scrap the custom url thing entirely and instead give users an option to add a second source.json? 

We should chat about what features we no longer want to support and what we want to rework.

crazywhalecc:
> I would say we scrap the custom url thing entirely and instead give users an option to add a second source.json?
> 
> We should chat about what features we no longer want to support and what we want to rework.

All the implementation I made has been committed to the v3-dev branch. The refactoring status is in #980 . If you have any other ideas, free to leave a message on Discord or submit a new RFC.

src/SPC/builder/unix/library/libjxl.php
crazywhalecc: Considering the default will not have avx512 support, this would be acceptable.

src/SPC/builder/unix/library/libwebp.php
crazywhalecc: Can't we explicitly specify this value by default? We should try to keep the situation under control.

henderkes: Why would we? Essentially all x86_64 cpus have sse support, it also works on debian 11 or older systems for that matter, it's just a bug with the source code here.

Edit: this is not avx512 like down below!

src/SPC/builder/extension/pdo_sqlsrv.php
henderkes: Done

src/SPC/builder/Extension.php
```diff
@@ -576,7 +576,7 @@ protected function addExtensionDependency(string $name, bool $optional = false):
 
     protected function getExtraEnv(): array
     {
-        return ['CFLAGS' => '-std=c17'];
+        return ['CFLAGS' => ''];
```
henderkes:
```suggestion
        return [];
```

henderkes:
@crazywhalecc this should be merged and tagged with a new release soon. Everyone who is building from main branch or old versions will run into failures if they attempt to use libxslt.

crazywhalecc (approved):

henderkes:
Perfect, please tag a new version soon.

PR #990, opened 2025-12-08
V3 feat/tool check

crazywhalecc:
## What does this PR do?

- [Add dev commands: is-installed, shell (for debugging package status)](https://github.com/crazywhalecc/static-php-cli/commit/d3e29b51dd49ac2831a8b8bca2a33cd40d6491ec)

- [Add Zig package support with downloader and installation checks](https://github.com/crazywhalecc/static-php-cli/commit/d3b98837b6415b0b7eef47224b9d26f171756afb)

- [Fix artifact downloade does not accept boolean options bug](https://github.com/crazywhalecc/static-php-cli/commit/77197f011832c047e816b4599746ea45f0f09062)

- [Fix pkg-config doctor fix using source bug](https://github.com/crazywhalecc/static-php-cli/commit/cda2a8df76293229492b5f0081db8031b78bcbf2)

- [Fix wrong namespace in go-xcaddy package](https://github.com/crazywhalecc/static-php-cli/commit/6a1b832b07b34c71e8a2fae838f6a863d502e4e7)

- [Allow absolute paths for configs](https://github.com/crazywhalecc/static-php-cli/commit/0296026062433ebca648693e7f05042849ffb000)

- [Add PackageInstaller::isPackageInstalled() API](https://github.com/crazywhalecc/static-php-cli/commit/80dec1717d11446e5cf1ab708231c0a4f513da44)

### DIFF
--- a/config/pkg.target.json
+++ b/config/pkg.target.json
@@ -82,10 +82,7 @@
     },
     "zig": {
         "type": "target",
-        "artifact": "zig",
-        "static-bins": [
-            "{pkg_root_path}/zig/zig"
-        ]
+        "artifact": "zig"
     },
     "nasm": {
         "type": "target",


src/globals/scripts/zig-cc.sh
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env bash
+
+if [ "$BUILD_ROOT_PATH" = "" ]; then
+    echo "The script must be run in the SPC build environment."
+    exit 1
```
crazywhalecc, line 5: I assume that everyone should use `zig-cc` inside SPC, so we can use shared env variables. cc @henderkes

henderkes: I'm not sure what you mean. I don't think the zig-cc script needs to check for BUILD_ROOT_PATH, if you do this we can't try to manually compile a source in case build failed in spc.

I use zig-cc from spc for normal compilation unrelated to spc all the time.

crazywhalecc: If you need to use it externally, I will revert it to `if BUILD_ROOT_PATH == '' then BUILDROOT_ABS={original relative path of script}`. How about that?

crazywhalecc: Resolved in #991

PR #1006, opened 2025-12-18
bunch of fixes/changes to make packages build

### LINKED ISSUE #1000: mysqlnd_parsec
https://github.com/mariadb-corporation/mysqlnd_parsec/tree/1.0.2

and there's an ed109 (or whatever it was) too

henderkes:
## What does this PR do?

- fix removal of directories with symlinks to directories
- add mysqlnd_parsec and mysqlnd_ed25519, closes #1000 
- remove grpc library from grpc ext dependencies
- temporary swoole patch (remove when 6.2.0 releases!)
- don't build addons explicitly
- look for pkgconfig files in all PKG_CONFIG_PATHs so we don't build stuff when a system lib exists and we explicitly want to link against it

### DIFF
--- a/src/SPC/ConsoleApplication.php
+++ b/src/SPC/ConsoleApplication.php
@@ -34,7 +34,7 @@ use Symfony\Component\Console\Application;
  */
 final class ConsoleApplication extends Application
 {
-    public const string VERSION = '2.7.11';
+    public const string VERSION = '2.8.0';
 
     public function __construct()
     {


src/SPC/builder/LibraryBase.php
```diff
@@ -195,6 +195,9 @@ public function tryBuild(bool $force_build = false): int
             $this->getBuilder()->emitPatchPoint('before-library[ ' . static::NAME . ']-build');
             $this->build();
             $this->installLicense();
+            if (getenv('CI')) {
+                FileSystem::removeDir($this->source_dir);
```
crazywhalecc: I'd rather using after build hook. Adding new env checker is unclear and undocumented. Adding features to version 2.0 may make it obsolete very quickly.

src/SPC/builder/LibraryBase.php
```diff
@@ -354,11 +357,11 @@ protected function installLicense(): void
         }
         foreach ($license_files as $index => $license) {
             if ($license['type'] === 'text') {
-                FileSystem::writeFile(BUILD_ROOT_PATH . '/source-licenses/' . $this->getName() . "/{$index}.txt", $license['text']);
+                FileSystem::writeFile(BUILD_ROOT_PATH . "/source-licenses/{$source}/{$index}.txt", $license['text']);
```
crazywhalecc, line 357: `FileSystem::createDir()` should also change to this.

henderkes: Good catch!

src/SPC/builder/LibraryBase.php
henderkes: We emit patch points but I'm not sure how to use them. Just realised that they also have one space too much. Can you do this?

crazywhalecc:
Use it externally in v2: `bin/spc build --build-cli bcmath -P a.php`

```php
// a.php
if (preg_match('/after-library\[(.*)\]-build/', patch_point(), $match)) {
    $lib_name = $match[1];
    FileSystem::removeDir(SOURCE_PATH . '/' . $lib_name);
}
```

> v3 will mark this as deprecated, using `#BeforeStage` and `#AfterStage` attribute instead. But removal will be performed in v4 (will be a long time)

henderkes:
That's not working FYI.

henderkes:
I'm silly. Forgot <?php...

henderkes:
~Swoole fails with 8.5.1 despite the patch.~

~Libwebp fails on alma 8 despite the check succeeding...~



And I need another vacation. The composer update before push failed because it failed to remove go package stuff in pkgroot. The bugs actually seem fixed.

henderkes:
Everything builds now, except for the macos intel failure, I guess.

crazywhalecc:
Looks like it's a JIT bug on x86_64 mAC05.

crazywhalecc:
Upstream issue here: https://github.com/php/php-src/issues/20767 . Looks like we will need to exclude clang version manually here or use homebrew llvm instead.

henderkes:
Swoole is broken on 8.5 until 6.2 release. Nothing we can realistically do there except manually using custom source option for download.

Keeping this PR frozen but open until Swoole releases 6.2. We don't need to add a patch for a dev version, I just need it to build swoole extension for packages.

> Upstream issue here: https://github.com/php/php-src/issues/20767 .

That will be fixed soon, we don't need to patch for it.

henderkes:
@crazywhalecc If you have a free minute (or a few) it might be worth still getting this into v2 since it contains actual fixes.

I'm happy to rebase v3 against it afterwards.

crazywhalecc:
I recently had to pause development on v3 because my full-time job was a bit busy, and I was also moving to a new house.

But I will check this again tomorrow / it seemed good last time.

bin/spc-alpine-docker
```diff
@@ -108,8 +108,7 @@ RUN apk update; \
         wget \
         xz \
         gettext-dev \
-        binutils-gold \
-        patchelf
+        binutils-gold
```
crazywhalecc, line 111: I suggest updating SPC_DOCKER_VERSION when changing docker image contents.

bin/spc-gnu-docker
```diff
@@ -92,11 +92,6 @@ RUN echo "source scl_source enable devtoolset-10" >> /etc/bashrc
 RUN source /etc/bashrc
 RUN yum install -y which
 
-RUN curl -fsSL -o patchelf.tgz https://github.com/NixOS/patchelf/releases/download/0.18.0/patchelf-0.18.0-$SPC_USE_ARCH.tar.gz && \
-    mkdir -p /patchelf && \
-    tar -xzf patchelf.tgz -C /patchelf --strip-components=1 && \
-    cp /patchelf/bin/patchelf /usr/bin/
-
```
crazywhalecc, line 99: Same as alpine docker

config/ext.json
```diff
@@ -487,6 +488,36 @@
...
+        "source": "mysqlnd_ed25519",
+        "arg-type": "enable",
+        "target": [
+            "shared"
+        ],
+        "ext-depends": [
+            "mysqlnd"
+        ],
+        "lib-depends": [
+            "libsodium",
```
crazywhalecc: https://github.com/mariadb-corporation/mysqlnd_ed25519/blob/4af1f2fef86ef2eb0a1b0c7fa09d93ea5e8a5e7d/config.m4#L16C1-L17C1

Looks like it also depends on `sodium` extension, and openssl is not a direct dependency.

config/ext.json
```diff
@@ -487,6 +488,36 @@
...
+            "openssl"
```
crazywhalecc, line 530: https://github.com/mariadb-corporation/mysqlnd_parsec/blob/15ca4ff8b1e27ede0f107c96995f44607221a8b0/config.m4#L23-L25

Similar issue here.

config/lib.json
```diff
@@ -361,6 +361,9 @@
         "source": "libargon2",
         "static-libs-unix": [
             "libargon2.a"
+        ],
+        "lib-suggests": [
+            "libsodium"
```
crazywhalecc, line 366: Why it depends on libsodium?

config/source.json
```diff
@@ -670,19 +682,20 @@
         }
     },
     "libpng": {
-        "type": "git",
-        "url": "https://github.com/glennrp/libpng.git",
-        "rev": "libpng16",
+        "type": "ghtagtar",
+        "repo": "pnggroup/libpng",
+        "match": "v1\\.6\\.\\d+",
+        "query": "?per_page=150",
```
crazywhalecc, line 698: Is this necessary, or is it only libpng that needs this?

src/SPC/builder/linux/LinuxBuilder.php
```diff
@@ -324,74 +327,6 @@ private function getMakeExtraVars(): array
...
-                    shell()->exec(sprintf(
-                        'patchelf --set-soname %s %s',
-                        escapeshellarg(basename($target)),
-                        escapeshellarg($target)
-                    ));
-                }
-            }
-        }
-    }
-
```
crazywhalecc, line 394: Did the soname patching put into separate patch point file?

src/SPC/builder/linux/library/openssl.php
```diff
@@ -51,14 +51,16 @@ public function build(): void
             $zlib_extra = '';
         }
 
+        $openssl_conf = getenv('OPENSSL_CONF');
+        $openssl_dir = $openssl_conf ? dirname($openssl_conf) : '/etc/ssl';
```
crazywhalecc: We need to add to docs or env.ini to track.

bin/spc-alpine-docker
henderkes: I don't think it's necessary as we haven't added anything. There's no reason to force users to rebuild.

config/source.json
henderkes: Only libpng needs this. They released a lot of v1.7 betas, so the latest 1.6 version is only on page 3.

src/SPC/builder/linux/LinuxBuilder.php
henderkes: I don't think it triggered any patch points.

src/SPC/builder/linux/library/openssl.php
henderkes: Agreed, I don't like that we even need to do this. Can't understand why there need to be 3 common ssl directories.

src/SPC/builder/linux/LinuxBuilder.php
crazywhalecc: Ah my mistake. I mean you extracted this function into an external script?

henderkes: No, it's just no longer needed.

config/ext.json
henderkes: True, it's required to connect with ssl though, which all servers running on the edge should use. I moved it to lib-suggests

config/lib.json
henderkes: It doesn't *depend* on libsodium, but because they implement a lot of the same symbols, this orders them correctly. I got rid of the patch in password_argon2.php instead as it failed with -lsodium -largon2.

Because this now suggests libsodium, meaning libsodium may contain symbols this library "relies on", SPC makes sure that libsodium comes after libargon2 everywhere. Because libargon2 implements all symbols of libsodium, no further symbols will be needed during linking and libsodium is ignored.

src/SPC/builder/linux/library/openssl.php
henderkes: Changing this to OPENSSLDIR actually, because the option expects a directory, not a path to the config file.

src/SPC/builder/linux/LinuxBuilder.php
```diff
@@ -283,11 +283,14 @@ protected function buildEmbed(): void
...
         if (file_exists($libphpSo)) {
-            // post actions: rename libphp.so to libphp-<release>.so if -release is set in LDFLAGS
-            $this->processLibphpSoFile($libphpSo);
             // deploy libphp.so
-            $this->deployBinary($libphpSo, $libphpSo, false);
+            preg_match('/-release\s+(\S*)/', getenv('SPC_CMD_VAR_PHP_MAKE_EXTRA_LDFLAGS'), $matches);
+            if (!empty($matches[1])) {
+                $libphpSoDest = str_replace('.so', '-' . $matches[1] . '.so', $libphpSo);
+            }
+            $this->deployBinary($libphpSo, $libphpSoDest, false);
```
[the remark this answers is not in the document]
henderkes: If you had a brain you'd understand that setting the soname of gcc-15 binaries with patchelf causes corrupt binaries.

crazywhalecc (approved):

PR #1009, opened 2025-12-26
Add MACOSX_DEPLOYMENT_TARGET to env.ini

### LINKED ISSUE #1008: Compiling on macOS 15 results in undefined symbol on macOS 12
Because GitHub deprecated the macOS 13 Intel runner images, we had to compile PHP on the new macOS 15 runners. 
This seems to result in an error on older macOS versions:

https://github.com/beyondcode/herd-community/issues/1600

Is there any way to prevent this from happening during compilation?

crazywhalecc:
## What does this PR do?

Closes #1008 

## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`

### DIFF
--- a/config/env.ini
+++ b/config/env.ini
@@ -142,6 +142,8 @@ SPC_CMD_PREFIX_PHP_CONFIGURE="./configure --prefix= --with-valgrind=no --enable-
 SPC_CMD_VAR_PHP_EMBED_TYPE="static"
 ; EXTRA_CFLAGS for `configure` and `make` php
 SPC_CMD_VAR_PHP_MAKE_EXTRA_CFLAGS="-g -fstack-protector-strong -fpic -fpie -Werror=unknown-warning-option ${SPC_DEFAULT_C_FLAGS}"
+; minimum compatible macOS version (LLVM vars, availability not guaranteed)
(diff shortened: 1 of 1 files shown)


henderkes:
Shouldn't we set this to current version by default?

crazywhalecc:
> Shouldn't we set this to current version by default?

Using current version will not fix anything and causes portability issue for old macOS. Setting it to the current version is the same as not setting the variable.

henderkes:
I know, but we also don't compile for old glibc versions on linux by default either. Users should overwrite the variable if they need compatibility with old macos.

crazywhalecc:
> I know, but we also don't compile for old glibc versions on linux by default either. Users should overwrite the variable if they need compatibility with old macos.

My understanding is that we've defaulted to maximum portability support on Linux (using musl-static), and that should be true for macOS as well. This is different from the glibc issue.

macOS 12 is the minimum version supported by modern toolchains (maybe). And this variable just fix the edge cases for one of our supported libraries (I didn't find which one). In theory, this variable is not needed for compatibility, but it's **apple** thing.

henderkes:
Yeah that's a good point, static is the default after all.

PR #1014, opened 2026-01-13
Fix nghttp2 and curl build configurations for static linking

### LINKED ISSUE #1013: Curl library: links against nghttp2.dll
CMake Warning:
  Manually-specified variables were not used by the project:

    CURL_STATICLIB
    NGHTTP2_CFLAGS

Therefore it never gets -DNGHTTP2_STATICLIB and links against __imp__... symbols.

crazywhalecc:
## What does this PR do?

Closes #1013

## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`

### DIFF
--- a/src/SPC/builder/windows/library/curl.php
+++ b/src/SPC/builder/windows/library/curl.php
@@ -30,7 +30,6 @@ class curl extends WindowsLibraryBase
                 '-DCMAKE_BUILD_TYPE=Release ' .
                 '-DBUILD_SHARED_LIBS=OFF ' .
                 '-DBUILD_STATIC_LIBS=ON ' .
-                '-DCURL_STATICLIB=ON ' .
                 '-DCMAKE_INSTALL_PREFIX=' . BUILD_ROOT_PATH . ' ' .
                 '-DBUILD_CURL_EXE=OFF ' . // disable curl.exe
                 '-DBUILD_TESTING=OFF ' . // disable tests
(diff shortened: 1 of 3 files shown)


src/SPC/builder/windows/library/curl.php
```diff
@@ -53,5 +52,7 @@ protected function build(): void
             );
         // move libcurl.lib to libcurl_a.lib
         rename(BUILD_LIB_PATH . '\libcurl.lib', BUILD_LIB_PATH . '\libcurl_a.lib');
+
+        FileSystem::replaceFileStr(BUILD_INCLUDE_PATH . '\curl\curl.h', '#ifdef CURL_STATICLIB', '#if 1');
```
henderkes, line 56: We should add  the definition to CMAKE_C_FLAGS instead.

src/SPC/builder/windows/library/nghttp2.php
```diff
@@ -29,11 +29,16 @@ protected function build(): void
...
+                '-DENABLE_STATIC_CRT=ON ' .
+                '-DENABLE_DOC=OFF ' .
+                '-DBUILD_TESTING=OFF '
             )
             ->execWithWrapper(
                 $this->builder->makeSimpleWrapper('cmake'),
                 "--build build --config Release --target install -j{$this->builder->concurrency}"
             );
+
+        FileSystem::replaceFileStr(BUILD_INCLUDE_PATH . '\nghttp2\nghttp2.h', '#ifdef NGHTTP2_STATICLIB', '#if 1');
```
henderkes, line 42: and the same here

src/SPC/builder/windows/library/curl.php
crazywhalecc: Since we have no unified cflags env var for Windows builder, the last executable build will always need it if depending on it. That's not appropriate for current build dependency tree. We have to change many other libs/executables cflags.

In v3 I implemented WindowsCMakeExecutor which may do this properly. it is more important to treat it as a permanent static library to avoid linking issues just for now, I think.

henderkes: Hmm, I haven't looked the windows stuff much, so I'll take your word for it.

Go 1.26 lets us compile cgo with clang on Windows, so this unlocks a msvc backend. In other words frankenphp is soon windows compatible, but the thought of distributing 800mb worth of dll's just to run it seems insane to me.

henderkes (approved):

PR #1026, opened 2026-02-10
Add frankenphp support for v3

crazywhalecc:
## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`

### DIFF
--- a/src/StaticPHP/Exception/SPCException.php
+++ b/src/StaticPHP/Exception/SPCException.php
@@ -175,6 +175,7 @@ abstract class SPCException extends \Exception
                         $package_type = match (true) {
                             $package instanceof LibraryPackage => 'library',
                             $package instanceof PhpExtensionPackage => 'php-extension',
+                            /* @phpstan-ignore-next-line */
                             $package instanceof TargetPackage => 'target',
                             default => 'package',
                         };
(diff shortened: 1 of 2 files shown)


src/StaticPHP/Artifact/ArtifactDownloader.php
```diff
@@ -214,6 +195,9 @@ public function __construct(protected array $options = [])
 
         // read downloads dir
         $this->_before_files = FileSystem::scanDirFiles(DOWNLOAD_PATH, false, true, true) ?: [];
+
+        // load downloaders
+        $this->downloaders = require ROOT_DIR . '/config/downloader.php';
```
henderkes, line 199: This seems like it's not the greatest idea. Should the file instead define the method that we autoload and call here?

crazywhalecc: Make it in core `registry` config?

henderkes: I'm not sure as I haven't had the time to familiarise myself with v3 yet. But I think this should optimally be part of Downloader.php or something as a method.

crazywhalecc: We already have `#[CustomBinary]` and `#[CustomSource]` for using custom download method, but they are registered to a specific package artifact and cannot reuse between different artifacts.

What I'm thinking is that in some edge cases we or others can dynamically add or enhance download types, such as rewriting GitHub URLs or adding commonly used advanced crawlers. For example, replacing the implementation of `ghtar` in a custom method can expand or customize the functionality globally.

PR #1034, opened 2026-02-17
fix redownloading go-xcaddy every time, version 2.8.3

henderkes:
## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`


henderkes:
fixes the other part of https://github.com/php/frankenphp/issues/1164

crazywhalecc (approved):
This bug should not exist in v3 due to the different implementation.

And, need new version release?

henderkes:
yes, for the libavif fix and the frankenphp path quoting.

henderkes:
although... maybe not? the spc binary is used by default I believe.

I won't tag a release and see if it's fixed by the nightly spc build

PR #1040, opened 2026-02-19
Shared

henderkes:
## What does this PR do?

allows building and linking against shared libraries

TODO: building single libraries as shared, rather than all. E.g. to fix mssql issues with static openssl.

## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`


henderkes:
Let's drop this instead of adding this feature in a less-than ideal implementation in v2. V3 should have this build int with the easier config definition, spc-config would have to use --static flags depending on what a library wants to be linked as (static or shared).

PR #1048, opened 2026-03-04
Add php-src mirror and use gmp mirror site

### LINKED ISSUE #1047: implement fallback download for php-src
php.net is constantly down, we should fall back to at least github sources, or mirror php.net

crazywhalecc:
## What does this PR do?

- Add php.net mirror site
- Use ftp.gnu.org instead of gmplib.org (GitHub IP is banned)
- Closes #1047 

## Checklist before merging

  - [X] `composer cs-fix`
  - [X] `composer analyse`
  - [X] `composer test`

### DIFF
--- a/src/SPC/builder/extension/ffi.php
+++ b/src/SPC/builder/extension/ffi.php
@@ -5,11 +5,21 @@ declare(strict_types=1);
 namespace SPC\builder\extension;
 
 use SPC\builder\Extension;
+use SPC\builder\linux\SystemUtil;
+use SPC\store\SourcePatcher;
 use SPC\util\CustomExt;
 
 #[CustomExt('ffi')]
 class ffi extends Extension
 {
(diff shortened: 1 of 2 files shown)


henderkes:
I unfortunately think the mirroring isn't going to be enough, because the site sometimes goes down entirely. I was thinking of a mirror that actually mirrors the entire data, not just reverse proxies to php.net (unless that's what you do?). Could use github releases instead, albeit not optimal, it's reachable at least.

crazywhalecc:
Of course it's full mirror. Using standalone server I mentioned in discord and php official rsync:

```bash
5 * * * * rsync -avzC --timeout=600 --delete --delete-after  rsync.php.net::phpweb /var/www/phpmirror.static-php.dev/
```

This server looks enough to host all of our services.

henderkes (approved):

PR #1055, opened 2026-03-09
Fix grpc build

### LINKED ISSUE #1054: ext-grpc build failed on macOS and Linux
```
bin/spc download --for-extensions=grpc
bin/spc build grpc --build-cli --debug
```

For Linux, using zig toolchain will get different error.

crazywhalecc:
## What does this PR do?

Closes #1054

## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`

### DIFF
--- a/config/lib.json
+++ b/config/lib.json
@@ -862,6 +862,9 @@
     },
     "openssl": {
         "source": "openssl",
+        "pkg-configs": [
+            "openssl"
+        ],
         "static-libs-unix": [
             "libssl.a",
             "libcrypto.a"
@@ -974,6 +977,11 @@
     },
     "unixodbc": {
         "source": "unixodbc",
+        "pkg-configs": [
+            "odbc",
+            "odbccr",
+            "odbcinst"
+        ],
         "static-libs-unix": [
             "libodbc.a",
             "libodbccr.a",
@@ -1015,6 +1023,9 @@
     },
     "zlib": {
         "source": "zlib",
+        "pkg-configs": [
+            "zlib"
+        ],
         "static-libs-unix": [
             "libz.a"
         ],
@@ -1028,6 +1039,9 @@
     },
     "zstd": {
         "source": "zstd",
+        "pkg-configs": [
+            "libzstd"
+        ],
         "static-libs-unix": [
             "libzstd.a"
         ],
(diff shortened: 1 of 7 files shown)


crazywhalecc:
Using ext-grpc to compile all grpc dependencies looks not good for our build procedure.

First of all, it's too slow for building static php binary every time. Building separately could reduce a lot of time when debugging and publishing. And another potential issue is building openssl, libcares and zlib twice. And also, although we can make a patch for boringssl and openssl built by grpc source to solve the CI error above, it often leads to misleading errors when building in-tree ext-grpc with all dependencies.

henderkes:
I know, but the extension is just shit. There's nothing we can do about it, short of completely replacing the extension build logic. We built the library standalone before, but the extension build doesn't use prebuilt libraries, it builds every time no matter what.

src/SPC/builder/extension/grpc.php
```diff
@@ -21,18 +20,50 @@ public function patchBeforeBuildconf(): bool
...
+      PHP_ADD_LIBRARY(rt,,GRPC_SHARED_LIBADD)
+      PHP_ADD_LIBRARY(rt)
+      ;;
+    esac
+
+  PHP_NEW_EXTENSION(grpc, @grpc_c_files@, $ext_shared, , -DGRPC_POSIX_FORK_ALLOW_PTHREAD_ATFORK=1)
+  PHP_SUBST(GRPC_SHARED_LIBADD)
+  PHP_INSTALL_HEADERS([ext/grpc], [php_grpc.h])
+fi
+M4;
```
henderkes, line 58: Very ugly, but the best we can do at this point. We will need to keep it in sync with version changes, though.

src/globals/test-extensions.php
```diff
@@ -16,15 +16,15 @@
     // '8.1',
     // '8.2',
     // '8.3',
```
henderkes, line 18: did you test for 8.3 too? 8.2 can be ignored.

crazywhalecc: I even tested it locally on version 8.0, and this m4 basically has maximum compatibility.

src/SPC/builder/extension/grpc.php
crazywhalecc: Yes, and for now, unless it changes the interface, we can tolerate it adding new files. It's ugly, but much simpler than patching the original files.

PR #1056, opened 2026-03-10
set custom binary name for frankenphp, allow linking against system openssl (fix mssql issues)

henderkes:
## What does this PR do?

closes #1040 
- picked the useful patches from the shared branch
- allows building against system openssl
- sets custom binary name for frankenphp
- reintroduces zig workaround because we use stable zig versions now
- fixes removeConfiguraArgs in UnixAutoConfexecutor

## Checklist before merging

  - [x] `composer cs-fix`
  - [x] `composer analyse`
  - [x] `composer test`

### DIFF
--- a/src/SPC/builder/LibraryBase.php
+++ b/src/SPC/builder/LibraryBase.php
@@ -381,7 +381,10 @@ abstract class LibraryBase
                     return false;
                 }
             }
-            return true; // allow using system dependencies if pkg_config_path is explicitly defined
+            // allow using system dependencies if pkg_config_path is explicitly defined
+            if (count($search_paths) > 1) {
+                return true;
+            }
         }
         foreach (Config::getLib(static::NAME, 'static-libs', []) as $name) {
             if (!file_exists(BUILD_LIB_PATH . "/{$name}")) {


src/SPC/builder/LibraryBase.php
```diff
@@ -365,6 +365,24 @@ protected function installLicense(): void
...
+                    if (file_exists($path . "/{$name}.pc")) {
+                        $found = true;
+                        break;
+                    }
+                }
+                if (!$found) {
+                    return false;
+                }
+            }
+            return true; // allow using system dependencies if pkg_config_path is explicitly defined
```
crazywhalecc: Bypassing check could cause many problems in the future I think. Skipping building and using dynamic libraries should be two separate things, we cannot just do for checking though.

crazywhalecc:
What do you think of it: `bin/spc build openssl,curl --build-cli --with-dynamic-libs=openssl` (For example, adding a .so definition to the config, or something like shared-pkg-configs)

This allows us to explicitly and definitively determine whether the dynamic library exists and is correctly imported.

src/SPC/builder/LibraryBase.php
henderkes: This behaviour was already intended before, you even documented it: https://github.com/crazywhalecc/static-php-cli/blob/main/docs/en/develop/system-build-tools.md#pkg-config-compilation-nix-only

It was simply bugged before.

henderkes:
Yes, it's a good idea, but after some consideration, I'm aiming that at v3. For now this is a good enough workaround that implements the logic we already promised.

crazywhalecc:
> Yes, it's a good idea, but after some consideration, I'm aiming that at v3. For now this is a good enough workaround that implements the logic we already promised.

> It was simply bugged before.

Yes, although its side effect is linking to system libraries, the system part is beyond our control if we really want to achieve this effect, unless we explicitly specify it and make it clear to ourselves what we want to do.

Passing shared libs explicitly, we could also determine and check `lib*.so` in SPCConfigUtil, including using system pkg-config.

henderkes:
That's why this logic only triggers when PKG_CONFIG_PATH is explicitly set. So users will have to e.g. `PKG_CONFIG_PATH=/usr/lib64/pkgconfig bin/spc build ...`. Otherwise, no system libraries are getting linked and we retain the regular static compilation logic.

>Passing shared libs explicitly, we could also determine and check lib*.so in SPCConfigUtil, including using system pkg-config.

Generally, yes, I want this to be configurable. It's quite useful for the system packages to minimise disk space, actually. But I don't think we should showhorn it into v2 anymore and then merge it into v3. V3 already gives us better ways to configure and override library compilation logic, so it should be done there directly.

config/lib.json
```diff
@@ -1028,6 +1039,9 @@
     },
     "zstd": {
         "source": "zstd",
+        "pkg-configs": [
+            "libzstd"
+        ],
```
crazywhalecc, line 1044: I can't remember if we've discussed this before, but if we can remove static-libs and headers for shared libs, then we don't need this bypass logic.

henderkes: Generally, no. But they will be useful to know in case we do package up static libraries.

crazywhalecc: Well, everything else is fine, I just feel that this bypass breaks the logic of the method itself. But I don't want to add a few more features to v2 that will be deprecated soon, either.

crazywhalecc: I think that for v2, it's better to only support pkg-configs definitions in openssl and the libraries you listed, or to do special processing for them, than to bypass PKG_CONFIG_PATH.

henderkes: We already had this in v2 though. Consider it a bug fix, if you will.

If you don't want it in, we can remove it, but should also remove the documentation regarding it.

crazywhalecc: What I mean is that we cannot allow pkg-configs to bypass the static-libs and other check simply because it exists, not that we cannot allow PKG_CONFIG_PATH to be included.

So unless there is a specific need, for dynamic libraries, it is a stable solution to let SPC use pkg-configs only, rather than bypassing the checks.

henderkes: I'd rather bypass the check than remove the static library part of the json just to add it back in again in v3. This ultimately achieves the same thing, the difference is whether we lose information in the process or (I prefer) not.

src/SPC/builder/LibraryBase.php
crazywhalecc:
```suggestion
            // allow using system dependencies if pkg_config_path is explicitly defined
            if (count($search_paths) > 1) {
                return true;
            }
```

If you prefer to keep the bypass, adding a simple check would be better for normal situations.

PR #1078, opened 2026-04-02
[v2] Fix bunch of build bugs for Windows

crazywhalecc:
## What does this PR do?

Fix sqlsrv redundant cflags when building PHP. Resolve https://github.com/static-php/static-php-cli-hosted/actions/runs/23831532972/job/69465987785#step:13:17283 .

## Checklist before merging

  - [X] `composer cs-fix`
  - [X] `composer analyse`
  - [X] `composer test`

### DIFF
--- a/src/globals/ext-tests/openssl.php
+++ b/src/globals/ext-tests/openssl.php
@@ -31,6 +31,6 @@ if (file_exists('/etc/ssl/openssl.cnf')) {
     }
     assert($valid);
 }
-if (PHP_VERSION_ID >= 80500 && !PHP_ZTS && defined('OPENSSL_VERSION_NUMBER') && OPENSSL_VERSION_NUMBER >= 0x30200000) {
+if (PHP_VERSION_ID >= 80500 && (!PHP_ZTS || PHP_OS_FAMILY !== 'Windows') && defined('OPENSSL_VERSION_NUMBER') && OPENSSL_VERSION_NUMBER >= 0x30200000) {
     assert(function_exists('openssl_password_hash'));
 }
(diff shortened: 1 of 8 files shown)


henderkes:
Is this necessary? Redundant means it builds with extra flags that aren't necessary, but we shouldn't patch the build unless it actually leads to failure.

crazywhalecc:
Three things: 

- This build error seems due to sqlsrv treating the warning as error. 
- We haven't supported zstd on windows yet (v2).
- ngtcp2 on Windows needs patched version of gnutls to enable QUIC feature, we haven't supported gnutls and its deps yet.

src/globals/ext-tests/openssl.php
```diff
@@ -31,6 +31,6 @@
     }
     assert($valid);
 }
-if (PHP_VERSION_ID >= 80500 && defined('OPENSSL_VERSION_NUMBER') && OPENSSL_VERSION_NUMBER >= 0x30200000) {
+if (PHP_VERSION_ID >= 80500 && !PHP_ZTS && defined('OPENSSL_VERSION_NUMBER') && OPENSSL_VERSION_NUMBER >= 0x30200000) {
```
henderkes: is ZTS on 8.5 not supported on Windows? on Linux it is

crazywhalecc: https://github.com/php/php-src/blob/50ad45851c597e745d4351d8b7dd930d8853aae5/ext/openssl/config.w32#L18-L26

No, but I haven't dig it deeper. I just read `config.w32` shows that openssl argon2 is not supported on ZTS windows. But we can add condition here indeed.

```suggestion
if (PHP_VERSION_ID >= 80500 && (!PHP_ZTS || PHP_OS_FAMILY !== 'Windows') && defined('OPENSSL_VERSION_NUMBER') && OPENSSL_VERSION_NUMBER >= 0x30200000) {
```

henderkes (approved):

PR #1086, opened 2026-04-08
fix(xlswriter): fix macOS build with modern Clang (C23)

dunglas:
## Summary
- The bundled minizip in xlswriter (pinned at libxlsxwriter `RELEASE_1.0.0`) uses a K&R-style function declaration in `mztools.c`
- Modern Clang on macOS defaults to C23, which **removed** K&R function definitions from the standard — this causes a hard syntax error, not a suppressible warning
- The upstream [libxlsxwriter `main` branch](https://github.com/jmcnamara/libxlsxwriter/blob/main/third_party/minizip/mztools.c) already has the ANSI C fix; the PECL extension just pins an old release
- This patch converts the K&R declaration to ANSI C in `patchBeforeMake()`, matching the upstream fix

## Error
```
ext/xlswriter/library/libxlsxwriter/third_party/minizip/mztools.c:30:30: error: unknown type name 'file'
   30 | extern int ZEXPORT unzRepair(file, fileOut, fileOutTmp, nRecovered, bytesRecovered)
```

## Test plan
- [x] Verify the fix applies cleanly by checking the string replacement targets match the actual source from PECL xlswriter
- [ ] macOS build should succeed with this patch

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/SPC/builder/extension/xlswriter.php
+++ b/src/SPC/builder/extension/xlswriter.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace SPC\builder\extension;
 
 use SPC\builder\Extension;
+use SPC\store\FileSystem;
 use SPC\store\SourcePatcher;
 use SPC\util\CustomExt;
 
@@ -28,6 +29,18 @@ class xlswriter extends Extension
     public function patchBeforeMake(): bool
     {
         $patched = parent::patchBeforeMake();
+
+        // Fix K&R C function declaration in bundled minizip rejected by modern Clang (C23 default)
+        $mztools = $this->source_dir . '/library/libxlsxwriter/third_party/minizip/mztools.c';
+        if (file_exists($mztools)) {
+            FileSystem::replaceFileStr(
+                $mztools,
+                "extern int ZEXPORT unzRepair(file, fileOut, fileOutTmp, nRecovered, bytesRecovered)\nconst char* file;\nconst char* fileOut;\nconst char* fileOutTmp;\nuLong* nRecovered;\nuLong* bytesRecovered;\n{",
+                "extern int ZEXPORT unzRepair(const char* file, const char* fileOut, const char* fileOutTmp, uLong* nRecovered, uLong* bytesRecovered)\n{"
+            );
+            $patched = true;
+        }
+
         if (PHP_OS_FAMILY === 'Windows') {
             // fix windows build with openssl extension duplicate symbol bug
             SourcePatcher::patchFile('spc_fix_xlswriter_win32.patch', $this->source_dir);
--- a/src/SPC/builder/extension/xlswriter.php
+++ b/src/SPC/builder/extension/xlswriter.php
@@ -5,9 +5,9 @@ declare(strict_types=1);
 namespace SPC\builder\extension;
 
 use SPC\builder\Extension;
-use SPC\store\FileSystem;
 use SPC\store\SourcePatcher;
 use SPC\util\CustomExt;
+use SPC\util\GlobalEnvManager;
 
 #[CustomExt('xlswriter')]
 class xlswriter extends Extension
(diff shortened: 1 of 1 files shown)


src/SPC/builder/extension/xlswriter.php
```diff
@@ -28,6 +29,18 @@ public function getWindowsConfigureArg(bool $shared = false): string
...
+
+        // Fix K&R C function declaration in bundled minizip rejected by modern Clang (C23 default)
+        $mztools = $this->source_dir . '/library/libxlsxwriter/third_party/minizip/mztools.c';
+        if (file_exists($mztools)) {
+            FileSystem::replaceFileStr(
+                $mztools,
+                "extern int ZEXPORT unzRepair(file, fileOut, fileOutTmp, nRecovered, bytesRecovered)\nconst char* file;\nconst char* fileOut;\nconst char* fileOutTmp;\nuLong* nRecovered;\nuLong* bytesRecovered;\n{",
+                "extern int ZEXPORT unzRepair(const char* file, const char* fileOut, const char* fileOutTmp, uLong* nRecovered, uLong* bytesRecovered)\n{"
+            );
+            $patched = true;
```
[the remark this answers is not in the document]
dunglas: Doesn't look appropriate to me; let's be consistent with the existing code base

henderkes:
It sounds like it makes sense, but I'm building on Linux in C23 mode with Clang, too. I would like to understand why it only fails on mac.

dunglas:
It's an Apple-specific change: Apple Clang (shipped with Xcode 16 on the macos-15 runners) defaults to C23, while upstream LLVM Clang on Linux still defaults to -std=gnu17 even in recent versions.

crazywhalecc:
Using main branch I have no build error on my machine (macOS 15) but only if I use `-std=c23` explicitly, I got error message. 

And looks like `ext-xlswriter` -> `libxlswriter` -> `minizip` -> `ioapi.c` have K&R C function declaration too, which also need to be patched.

henderkes:
I think the better course of action here is to explicitly pin to std=c17 until it's fixed upstream: https://github.com/viest/php-ext-xlswriter/issues/559

dunglas:
@crazywhalecc @henderkes You're right, `ioapi.c` also has K&R declarations. I've updated the PR to use `-std=gnu17` instead of patching individual files — this covers all K&R occurrences in the bundled minizip. Added via both `SPC_CMD_VAR_PHP_MAKE_EXTRA_CFLAGS` (for static builds) and `getExtraEnv()` (for shared/phpize builds).

I've also opened a PR upstream to update libxlsxwriter: https://github.com/viest/php-ext-xlswriter/pull/560 — once that's merged, this `-std=gnu17` workaround can be removed.

src/SPC/builder/extension/xlswriter.php
```diff
@@ -30,14 +30,9 @@ public function patchBeforeMake(): bool
...
 
-        // Fix K&R C function declaration in bundled minizip rejected by modern Clang (C23 default)
-        $mztools = $this->source_dir . '/library/libxlsxwriter/third_party/minizip/mztools.c';
-        if (file_exists($mztools)) {
-            FileSystem::replaceFileStr(
-                $mztools,
-                "extern int ZEXPORT unzRepair(file, fileOut, fileOutTmp, nRecovered, bytesRecovered)\nconst char* file;\nconst char* fileOut;\nconst char* fileOutTmp;\nuLong* nRecovered;\nuLong* bytesRecovered;\n{",
-                "extern int ZEXPORT unzRepair(const char* file, const char* fileOut, const char* fileOutTmp, uLong* nRecovered, uLong* bytesRecovered)\n{"
-            );
+        // Bundled minizip uses K&R C function declarations rejected by C23 (default on macOS with Xcode 16+)
```
henderkes:
```suggestion
        // Remove when https://github.com/viest/php-ext-xlswriter/pull/560 is merged
```

src/SPC/builder/extension/xlswriter.php
```diff
@@ -53,4 +48,9 @@ public function patchBeforeMake(): bool
         }
         return $patched;
     }
+
+    protected function getExtraEnv(): array
```
henderkes, line 53:
```suggestion
    // Remove when https://github.com/viest/php-ext-xlswriter/pull/560 is merged
    protected function getExtraEnv(): array
```

henderkes (changes requested):

src/SPC/builder/extension/xlswriter.php
```diff
@@ -28,6 +29,13 @@ public function getWindowsConfigureArg(bool $shared = false): string
     public function patchBeforeMake(): bool
     {
         $patched = parent::patchBeforeMake();
+
+        // Remove when https://github.com/viest/php-ext-xlswriter/pull/560 is merged
+        if (PHP_OS_FAMILY !== 'Windows') {
+            GlobalEnvManager::putenv('SPC_CMD_VAR_PHP_MAKE_EXTRA_CFLAGS=' . getenv('SPC_CMD_VAR_PHP_MAKE_EXTRA_CFLAGS') . ' -std=gnu17');
```
henderkes, line 35: Should we use gnu17 here instead of c17?

henderkes (approved):

crazywhalecc (approved):
It works fine on my mac.

PR #1101, opened 2026-04-13
Update README with v3 release announcement

crazywhalecc:
## What does this PR do?

Announce upcoming release of v3 with project rename and repository move.

## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,11 @@
 # static-php-cli
 
+> [!IMPORTANT]
+> We are preparing to release **v3**, which will include a project rename from **static-php-cli** to **StaticPHP**.
+> The repository will also be moved to **static-php/static-php**.
+> Please update your references and stay tuned for the official release.
+
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,5 +1,11 @@
 # static-php-cli
 
+> [!IMPORTANT]
+> We are preparing to release **v3**, which will include a project rename from **static-php-cli** to **StaticPHP**.
+> The repository will also be moved to **static-php/static-php**.
+> Please update your references and stay tuned for the official release.
```
henderkes, line 6: StaticPhp but static-php is still unfortunate :/

crazywhalecc: Yup, but repository name and orgANIZaTioN name being same would be more unified in engineering and sourcing. In any case, this is a small compromise: simplifying static-php-cli and emphasizing branding.

Since we cannot own staticphp org, we can only make this compromise with the hyphen.

crazywhalecc: Well, blame me I'm not a naming genius; I didn't come up with a good name that I could keep to myself from the start.

PR #1109, opened 2026-04-16
Add linux mint to ubuntu/debian variants

Pierstoval:
Since Linux Mint is by default based on Ubuntu, but LMDE is based on Debian, this change should work in both OSes (at least it works on my LMDE 7 machine)


henderkes (approved):

crazywhalecc (approved):
This reminded me that I think it would be better to directly change the version to identify the distro family by parsing ID_LIKE on v3.

PR #1133, opened 2026-05-08
Add dev:gen-ext-test-matrix command

crazywhalecc:
## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`

### DIFF
--- a/config/pkg/ext/ext-swow.yml
+++ b/config/pkg/ext/ext-swow.yml
@@ -14,6 +14,6 @@ ext-swow:
     - curl
     - ext-openssl
     - ext-curl
-    - pgsql
+    - postgresql
   php-extension:
     arg-type: custom
--- a/config/pkg/ext/ext-swow.yml
+++ b/config/pkg/ext/ext-swow.yml
@@ -9,5 +9,10 @@ ext-swow:
     metadata:
       license: Apache-2.0
       license-files: [LICENSE]
+  suggests:
+    - openssl
+    - curl
+    - ext-openssl
+    - ext-curl
   php-extension:
     arg-type: custom
--- a/config/pkg/ext/ext-swow.yml
+++ b/config/pkg/ext/ext-swow.yml
@@ -14,5 +14,6 @@ ext-swow:
     - curl
     - ext-openssl
     - ext-curl
+    - pgsql
   php-extension:
     arg-type: custom
--- a/config/pkg/ext/ext-swow.yml
+++ b/config/pkg/ext/ext-swow.yml
@@ -9,10 +9,5 @@ ext-swow:
     metadata:
       license: Apache-2.0
       license-files: [LICENSE]
-  suggests:
-    - openssl
-    - curl
-    - ext-openssl
-    - ext-curl
   php-extension:
     arg-type: custom
--- a/docs/en/guide/cli-reference.md
+++ b/docs/en/guide/cli-reference.md
@@ -29,7 +29,7 @@ spc download [artifacts] [options]
 | `--for-packages=<list>`         |       | Download artifacts needed by the given packages       |
 | `--without-suggests`            |       | Skip suggested packages when using `--for-extensions` |
 | `--clean`                       |       | Delete existing download cache before fetching        |
-| `--with-php=<ver>`              |       | PHP version in `major.minor` format (default: `8.4`)  |
+| `--with-php=<ver>`              |       | PHP version in `major.minor` format (default: `8.5`)  |
 | `--prefer-binary`               | `-p`  | Prefer pre-built binaries over source archives        |
 | `--prefer-source`               |       | Prefer source archives over pre-built binaries        |
 | `--source-only`                 |       | Only download source artifacts                        |
@@ -47,7 +47,7 @@ spc download [artifacts] [options]
 
 ```bash
 # Download only what the chosen extensions need
-spc download --for-extensions="bcmath,openssl,curl" --with-php=8.4
+spc download --for-extensions="bcmath,openssl,curl" --with-php=8.5
 
 # Download specific artifacts
 spc download "php-src,openssl"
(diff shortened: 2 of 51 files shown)


src/Package/Target/php/unix.php
```diff
@@ -167,14 +168,21 @@ public function beforeMakeUnix(ToolchainInterface $toolchain): void
...
+    public function tryPatchMakefileUnix(ToolchainInterface $toolchain): void
     {
         if (SystemTarget::getTargetOS() !== 'Linux') {
             return;
         }
 
         // replace //lib with /lib in Makefile
         shell()->cd(SOURCE_PATH . '/php-src')->exec('sed -i "s|//lib|/lib|g" Makefile');
+
+        // try to fix minilua with zig-cc, disable vectorize for some edge cases
```
henderkes: what does minilua have to do with php? and why patch the makefile instead of apprnd to cflags?

crazywhalecc: minilua is a tool used to compile PHP and is unrelated to the build artifacts. Segfault seems to only trigger on the GHA runner here, I can't reproduce it locally.

We can also specify `BUILD_CC = cc` to prevent the zig-cc problem, and using the flag to disable the feature will not have any impact too.

crazywhalecc: Well, using a local compiler is the safest way

crazywhalecc:
# Summary of changes

## CI automation

- New CI automation commands and workflows
  - `dev:gen-ext-test-matrix`: Generate a balanced matrix for extension and library tests, avoiding conflicts and matrix waste, and merge extension tests
  - `dev:test-bot`: Analyze PR changed files and labels, call matrix generating function

## Bugfix

- Fix `--retry` command on API calling and `git` calling.
- Remove `binary: hosted` temporarily.
- Fix extension `memcache,imap,xmlreader,opentemetry,swow,intl,ds,glfw,brotli,zstd` build bug
- Remove duplicate phar patch
- Fix UnixCMakeExecutor incorrectly applying macOS frameworks on linux
- Patch `BUILD_CC` using system c compiler instead of `zig-cc`

henderkes:
I'll first see why using zig-cc leads to a crash there. Other than that looks good.

config/pkg/ext/ext-swow.yml
```diff
@@ -9,10 +9,5 @@ ext-swow:
     metadata:
       license: Apache-2.0
       license-files: [LICENSE]
-  suggests:
-    - openssl
-    - curl
-    - ext-openssl
-    - ext-curl
```
henderkes, line 16: https://github.com/swow/swow/blob/develop/ext/config.m4#L181

the libs should remain

docs/en/guide/cli-reference.md
```diff
@@ -36,7 +36,7 @@ spc download [artifacts] [options]
 | `--binary-only`                 |       | Only download binary artifacts                        |
 | `--parallel=<n>`                | `-P`  | Number of parallel downloads (default: `1`)           |
 | `--retry=<n>`                   | `-R`  | Number of retries on failure (default: `0`)           |
-| `--ignore-cache=<list>`         |       | Force re-download the specified artifacts             |
+| `--ignore-cache=<list>`         | `-i`  | Force re-download the specified artifacts             |
```
henderkes, line 39: we don't need a shorthand flag for that I think

docs/zh/guide/cli-reference.md
```diff
@@ -22,23 +22,23 @@ spc download [artifacts] [options]
...
 | `--for-extensions=<list>` | `-e` | 按扩展名下载其所需的制品 |
 | `--for-libs=<list>` | `-l` | 按库名下载其所需的制品 |
-| `--for-packages=<list>` | | 按包名下载其所需的制品 |
-| `--without-suggests` | | 使用 `--for-extensions` 时跳过建议包 |
-| `--clean` | | 下载前删除旧的下载缓存 |
-| `--with-php=<ver>` | | PHP 版本，格式为 `major.minor`（默认 `8.4`）|
+| `--for-packages=<list>` |      | 按包名下载其所需的制品 |
+| `--without-suggests` |      | 使用 `--for-extensions` 时跳过建议包 |
+| `--clean` |      | 下载前删除旧的下载缓存 |
+| `--with-php=<ver>` |      | PHP 版本，格式为 `major.minor`（默认 `8.4`）|
```
henderkes: bump to 8.5 with v3

docs/en/guide/cli-reference.md
crazywhalecc: In my own experience, I use it frequently because it's impossible for me to download all the latest packages every day. Accessing GitHub from China even in Shanghai is terrible.

config/pkg/ext/ext-swow.yml
crazywhalecc: Oh, I almost forgot about this. The generating result from `gen-ext-test-matrix` is a little buggy for swow. Due to feature of swow, it will effect native extension's behavior. Either exclude SWOW from testing or separate it from the curl extension.

henderkes: Ah for static built in compilation yes. Perhaps we should guard the options to off if the regular extensions are enabled and we compile static swow.

henderkes: Wait, that would be simpler for swoole as well.

henderkes: Oh but we'd need to check for every extension that would interfere, so e.g. pgsql,swoole,swow wouldn't work together in any combination.

config/pkg/ext/ext-swow.yml
```diff
@@ -9,5 +9,10 @@ ext-swow:
     metadata:
       license: Apache-2.0
       license-files: [LICENSE]
+  suggests:
+    - openssl
+    - curl
+    - ext-openssl
+    - ext-curl
```
henderkes, line 16: it doesn't really suggest those, only the libraries, no?

and pgsql too

henderkes:
```suggestion
    - ext-openssl
    - ext-curl
    - pgsql
```

crazywhalecc: Hmm... I think either is fine, since its suggestions are hooks, not features, including libs. Swow is not providing features like swoole.

crazywhalecc: This issue has been isolated. The suggested extension and library are working fine, and the latest CI also shows that swow's hook is causing problems with Windows curl. I'll create an issue for swow later, and our tests will temporarily work around this problem

config/pkg/ext/ext-swow.yml
```diff
@@ -14,5 +14,6 @@ ext-swow:
     - curl
     - ext-openssl
     - ext-curl
+    - pgsql
```
crazywhalecc:
```suggestion
    - postgresql
```

henderkes (approved):

PR #1138, opened 2026-05-11
Feat/pgo v3

henderkes:
## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`


henderkes:
(I tested with static-php/packages v3 branch, php 8.5 zig on rpm. Tests with Alpine and gcc are still to finish in a few hours when ci runs.)

henderkes:
Wair, why is curl a target like php and not a package/library? We don't compile the exe right?

crazywhalecc:
> Wair, why is curl a target like php and not a package/library? We don't compile the exe right?

We could compile curl.exe, which shows that StaticPHP is not only suitable for building static PHP, but also for other projects. Just like we already support building pkg-config and re2c. Making them targets simply makes reasonable.

`target` is superset of `library`.

henderkes:
> > Wair, why is curl a target like php and not a package/library? We don't compile the exe right?
> 
> We could compile curl.exe, which shows that StaticPHP is not only suitable for building static PHP, but also for other projects. Just like we already support building pkg-config and re2c. Making them targets simply makes reasonable.
> 
> `target` is superset of `library`.

Okay, added curl.exe on Windows too. Also figured out the root cause of the BUILD_CC patch minilua crash (zigs undefined behaviour sanitizer tripping up, if you run `CC="clang -fsanitize=undefined"` it will also fail) which coincidentally also fixed a possible edge case bug. BUILD_CC is also used for gen_ir so by using a different vm type it could have theoretically clashed.

henderkes:
Oh, there's even an issue running the binaries despite no undefined sanitizer that I can't reproduce locally or in docker. Very weird.

crazywhalecc:
> Oh, there's even an issue running the binaries despite no undefined sanitizer that I can't reproduce locally or in docker. Very weird.

Yup. I have no idea except using native compiler instead.

henderkes:
Hmm, the iconv fail on windows should be unrelated to my changes.

src/StaticPHP/Util/PkgConfigUtil.php
```diff
@@ -67,7 +67,7 @@ public static function getModuleVersion(string $pkg_config_str): string
     public static function getCflags(string $pkg_config_str): string
     {
         // get other things
-        $result = self::execWithResult("pkg-config --static --cflags-only-other {$pkg_config_str}");
+        $result = self::execWithResult("pkg-config --static --cflags {$pkg_config_str}");
```
crazywhalecc, line 70: Do we really need `--cflags` here? Output may be redundant.

henderkes: Imagemagick failed because it was missing -I... paths.

src/Package/Library/watcher.php
```diff
@@ -20,9 +20,8 @@ public function build(): void
         if (stripos($cflags, '-fpic') === false) {
             $cflags .= ' -fPIC';
         }
-        $ldflags = $this->getLibExtraLdFlags() ? ' ' . $this->getLibExtraLdFlags() : '';
         shell()->cd("{$this->getSourceDir()}/watcher-c")
-            ->exec(getenv('CXX') . " -c -o libwatcher-c.o ./src/watcher-c.cpp -I ./include -I ../include -std=c++17 -Wall -Wextra {$cflags}{$ldflags}")
+            ->exec(getenv('CXX') . " -c -o libwatcher-c.o ./src/watcher-c.cpp -I ./include -I ../include -std=c++17 -Wall -Wextra {$cflags}")
```
crazywhalecc, line 24: Why don't we need ldflags anymore?

src/StaticPHP/Util/PkgConfigUtil.php
crazywhalecc: Not sure if this behavior is safe and not tested here, but I'd rather fix ImageMagick than change the entire pkgconfig for a single library.

src/Package/Artifact/zig.php
```diff
@@ -131,5 +132,151 @@ public function postExtractZig(string $target_path): void
         chmod("{$target_path}/zig-ld.lld", 0755);
         chmod("{$target_path}/zig-ranlib", 0755);
         chmod("{$target_path}/zig-objcopy", 0755);
+
+        // Build the clang runtime bits zig 0.15+ doesn't ship: profile runtime
+        // (so -fprofile-generate actually emits .profraw) and crtbegin/crtend
+        // (so shared libs get __dso_handle and the __cxa_finalize atexit hook).
+        // These get auto-linked by the zig-cc wrapper when the right flags fly past.
+        $this->buildClangRuntimeBits($target_path);
```
crazywhalecc: I think separating a target and add another doctor check, calling installer is better here. I'll push later.

src/Package/Library/watcher.php
henderkes: Because this only runs a single compile step without direct linking, it lead to unexpected things with e.g. -Wl,--gc-sections.

src/Package/Artifact/zig.php
henderkes: Using --cflags should be the right approach anyway. I don't understand the point of using only-other.

Yes, it could be turned into an arch-specific target. That would enable full cross-compilation to every arch. I was able to successfully compile for arm-linux-musleabihf -dynamic on my x86 alma 10 system. Zig is so fun.

src/StaticPHP/Doctor/Item/ZigCheck.php
```diff
@@ -39,4 +39,34 @@ public function installZig(): bool
...
+            return CheckResult::ok("{$libDir}/libclang_rt.profile.a");
+        }
+        return CheckResult::fail('clang runtime bits are not built', 'build-clang-runtime-bits');
+    }
+
+    #[FixItem('build-clang-runtime-bits')]
+    public function fixClangRuntimeBits(): bool
+    {
+        $installer = new PackageInstaller(interactive: false);
+        $installer->addInstallPackage('clang-runtime-bits');
```
henderkes: we should probably find a better name for this then 😆

crazywhalecc: Any better suggestions? 🧐

henderkes: compiler-rtCrtObjects?

crazywhalecc: Camel case looks weird here. And `zig-compiler-rt` ?

henderkes: it's not full compiler-rt, only the crt start and end objects.

`compiler-rt-crt-objects`?

henderkes: We might also add llvm-objcopy because gnu objcopy is too dumb to work on other arch binaries.

crazywhalecc: Then I think we download full compiler rt anyway, just `compiler-rt` or `llvm-compiler-rt` is acceptable.

henderkes: From compiler-rt we only need the crt objects and the profiling runtime library. llvm-objcopy is from llvm-binutils. Different download.

Or we download full prebuilt llvm-toolchain. It has everything we need but it's a 2gb download instead of 200kb+2 min build time.

henderkes: Edit: downloading compiler-rt is not a good idea because we'd need to download 2gb per target platform, so at least 2x. For llvm-objcopy we could download, but I feel like 2 minutes of build time are preferable to a 2gb download.

henderkes:
It's working as intended now, but compiling llvm binutils is really taking quite some time. We should host a prebuilt package for that and use that regardless of whether the user prefers prebuilts or not, there's no reason to rebuild this. It's static (always compiled with -musl) per arch-os.

I tried downloading llvm tarball instead, but GitHub gives me such a bad connection here in SEA that the wait time is too ridiculous.

crazywhalecc:
> We should host a prebuilt package for that and use that regardless of whether the user prefers prebuilts or not

Just add a workflow in `static-php/hosted` repo like icu-static-windows one.

crazywhalecc:
> I tried downloading llvm tarball instead, but GitHub gives me such a bad connection here in SEA that the wait time is too ridiculous.

😞 Singapore VPN has the best speed accessing GitHub for me.

src/StaticPHP/Doctor/Item/GoXcaddyCheck.php
```diff
@@ -0,0 +1,44 @@
...
+
+use StaticPHP\Attribute\Doctor\CheckItem;
+use StaticPHP\Attribute\Doctor\FixItem;
+use StaticPHP\Attribute\Doctor\OptionalCheck;
+use StaticPHP\DI\ApplicationContext;
+use StaticPHP\Doctor\CheckResult;
+use StaticPHP\Package\PackageInstaller;
+
+#[OptionalCheck([self::class, 'optionalCheck'])]
+class GoXcaddyCheck
```
crazywhalecc, line 15: Is there any reason that we must check go-xcaddy in doctor check?

src/StaticPHP/Package/PackageBuilder.php
```diff
@@ -199,9 +202,12 @@ public function extractDebugInfo(string $binary_path): string
      */
     public function stripBinary(string $binary_path): void
     {
+        $strip = ApplicationContext::tryGet(ToolchainInterface::class) instanceof ZigToolchain
```
crazywhalecc, line 205: Why tryGet here?

src/StaticPHP/Package/PackageInstaller.php
```diff
@@ -571,6 +573,77 @@ public function getPhpExtensionPackage(string $package_or_ext_name): ?PhpExtensi
         return null;
     }
 
+    private function reconcilePhpSrcVersion(): void
```
crazywhalecc, line 586: This solution seems too wild, and I don't think it is appropriate here.

src/StaticPHP/Package/PackageBuilder.php
henderkes: for some reason the DI library you're using returns "true" for ->has even if the class is in fact not autowirable

src/StaticPHP/Artifact/ArtifactDownloader.php
```diff
@@ -731,6 +734,16 @@ private function generateQueue(Artifact $artifact): array
...
+        if ($source_downloaded && $artifact->getName() === 'php-src' && ($requested = $this->getOption('with-php'))) {
+            $info = ApplicationContext::get(ArtifactCache::class)->getSourceInfo('php-src');
+            $cv = $info['version'] ?? null;
+            $ct = $info['cache_type'] ?? null;
+            $matches = $requested === 'git' ? $ct === 'git' : ($cv !== null && $ct !== 'git' && ($cv === $requested || str_starts_with($cv, $requested . '.')));
+            if (!$matches) {
+                $source_downloaded = false;
+            }
+        }
+
```
crazywhalecc, line 746: you want to try to switch PHP versions simply by changing `--with-php`?

src/StaticPHP/Doctor/Item/GoXcaddyCheck.php
henderkes: not strictly necessary, but it makes things like pre building images by just running doctor a bit easier. you can remove it if you want

src/StaticPHP/Artifact/ArtifactDownloader.php
henderkes: Yes, --dl-with-php switches the php version in the build command, or spc download --with-php=.. switches the php version too. We shouldn't use the old version in source/php-src when the user explicitly switched it.

Even better would be to have different trees for different php versions (php-src-8.2, ... , php-src-8.5), that would simplify the switch more, but it's a larger refactor.

Edit: v2 already did this with the source dir stale logic. In v3 it kept rebuilding the old php versions even after switching.

src/StaticPHP/Doctor/Item/GoXcaddyCheck.php
crazywhalecc: Oh, I did forget to consider the base image in v3. But for FrankenPHP, I think it's easy to just run `bin/spc install-pkg go-xcaddy` after the doctor command.

src/StaticPHP/Artifact/ArtifactDownloader.php
crazywhalecc: Ah, I think I need to look into that. I previously designed a separate command to achieve this `bin/spc switch-php-version`. I have not considered using `--with-php` to switch it directly.

crazywhalecc: Anyway hardcoding looks strange in ArtifactDownloader and PackageInstaller, and I'll take a look at how to fix it later.

henderkes: Ah, could you also add llvm-tools to static-php/hosted? We only need a x86-64 and an aarch64 version for linux (any maybe macOS in the future, pending m5 mini availability next month). They are linked statically (musl target, like all packages) and only update when zig version updates.

crazywhalecc: Of course, but let's take them one by one first 🤯 There's so much to do before releasing the first alpha version.

henderkes:
wtf I have no idea what failed in the tests and why. it passes locally and looking for php-config there makes no sense, neither of the new extensions does it

luthermonson:
had claude pull everything and this is what it thinks the problem is with CI

``` 
Why it broke now (not a PR regression)

  The earlier runs on this same PR (02:27, 06:11 today) passed. Then it broke at 08:43.

  - 06:11 green run used mongodb-2.3.2.tgz
  - 08:43 failed run used mongodb-2.3.3.tgz
  - mongodb 2.3.3 was released today at 07:07:49Z — right between those two runs
  - spc's ext-mongodb.yml uses type: ghrel + match: mongodb.+\.tgz, which always pulls the latest mongodb release. So CI silently picked up the new
  version.
```

I'll get a PR up with the fix and we'll see if it's right

henderkes:
(current CI failures are because I didn't forward the temporary fixes for fastchart and clickhouse extensions, iliaa will probably release versions with my fixes upstreamed in a few days)

crazywhalecc:
I'd like to merge some easily split parts of this branch into v3 first, so that the branch doesn't become too large and make reviews troublesome.

henderkes:
Yes, I'm getting to a point where building static-php/packages based on this branch is possible. The majority of changes here are really just behaviour fixes that can be applied independently.

src/StaticPHP/Util/SPCConfigUtil.php
```diff
@@ -32,123 +35,82 @@ public function __construct(array $options = [])
         $this->absolute_libs = $options['absolute_libs'] ?? false;
     }
 
-    public function config(array $packages = [], bool $include_suggests = false): array
+    public function config(array $packages = []): array
```
crazywhalecc, line 38: Directly and permanently relying on `PackageInstaller`'s resolved packages prevents `SPCConfigUtil` from being used in standalone commands or other vendor-mode scenarios. I suggest splitting this into two methods: keep `config()` as is, and add `function configForResolvedPackages(array $roots, array $resolvedPackages, array $phpExtras = []) {}` to distinguish the use cases.

- `spc-config` retains its original semantics.
- The build process respects the actual resolution results from `PackageInstaller`.
- It avoids linking `suggests` that haven't been built.
- It does not rely on implicit DI state.
- Unit tests only require passing arrays, eliminating the need to construct a full installer instance.
- Identical inputs consistently yield identical outputs.

crazywhalecc (changes requested):

src/StaticPHP/Util/SPCConfigUtil.php
crazywhalecc: I will apply fix for SPCConfigUtil in a separate new branch. This is not strictly part of PGO.

crazywhalecc: This seems a bit tricky; I think it would be better to split the various changes into separate PRs for review and merging. I have already completed some unrelated changes in #1214, and I am about to create a PR for the `SPCConfigUtil` refactoring (including the relevant files). Once all the changes unrelated to PGO have been merged, I will review the `pgo-v3` changes and submit them to a new, clean branch.

config/pkg/ext/ext-opentelemetry.yml
```diff
@@ -2,8 +2,9 @@ ext-opentelemetry:
   type: php-extension
   artifact:
     source:
-      type: pecl
-      name: opentelemetry
+      type: pie
+      repo: open-telemetry/ext-opentelemetry
+      build-root: ext
```
crazywhalecc, line 7: How about `subdir`? `build-root` is definitely ambiguous. Also, move all the previous `metadata.source-root` here.

henderkes: build-dir?

subdir doesn't say what that directory is used for

crazywhalecc: Keep using source-root? build-dir sounds ambiguous too (out-of-source building). The name `source-root` also corresponds to our API `getSourceRoot()`. Although it still doesn't accurately describe our needs, it essentially redirects the source root.

henderkes: Two diffrremt things, we need both of them.

crazywhalecc: Another approach is to completely redirect everything using source-root, eliminating the need to worry about subdirectories during in-tree extension builds:

1. All artifacts are uniformly extracted to source/{name}
2. During static compilation of PHP extensions, a symbolic link is created to the root directory or the specified source-root to php-src/ext/{name}
3. Shared extensions remain unaffected

crazywhalecc: After doing this, our previous source-root and the current source-root for PHP extensions are actually the same thing.

henderkes: Are you sure that doesn't have edge cases with extension configs resolving to absolute paths before? If it's working with all extensions I think that's a good idea.

PR #1143, opened 2026-05-19
add windows static build support for libxslt, gmp, gettext

luthermonson:
## Summary

Flips three Windows-`wip` extensions to supported by adding native static-build Windows library builders. Each library is a separate commit.

### `xsl` — `libxslt` via CMake
Mirrors the existing `windows/library/libxml2.php` pattern. libxslt 1.1.34+ ships native `CMakeLists.txt`; build options match the existing unix trait (`--without-python`, `--without-crypto`, `--without-debugger`, `--without-debug`, libxml2 prefix). CMake install produces suffixed `libxsltS.lib` / `libexsltS.lib` which are renamed to `libxslt.lib` / `libexslt.lib`, same shape as libxml2.php's `libxml2s.lib` → `libxml2_a.lib`.

### `gmp` — MPIR via PHP-SDK deps
GNU GMP has no native MSVC build. This uses the same MPIR (GMP-compatible fork) build that PHP itself bundles for its official Windows builds, pulled from `downloads.php.net/~windows/php-sdk/deps/vs17/x64/mpir-3.0.0-1-vs17-x64.zip`. Pattern follows `postgresql-win` — declarative copy of `lib/mpir_a.lib` + `include/mpir/*.h` into the build root. New \`mpir-win\` lib is wired in via \`lib-depends-windows: [\"mpir-win\"]\` on the `gmp` ext, leaving the unix `gmp` library untouched.

### `gettext` — `libintl` via PHP-SDK deps
GNU gettext has no native MSVC build either. Same approach: pull `libintl-0.18.3-9-vs17-x64.zip` from the canonical PHP-SDK deps server, copy `lib/libintl_a.lib` + `include/libintl.h`. New \`libintl-win\` lib wired in via \`lib-depends-windows\` on `gettext`.

## Metadata cleanup

For all three extensions:
- removed \`\"Windows\": \"wip\"\` from \`support\`
- added \`\"arg-type-windows\": \"with\"\` (Windows \`configure.bat\` doesn't accept \`with-foo=PATH\` the same way autoconf does — matches the existing bz2/iconv/curl/gd pattern)

Also dropped a stale `static-libs-windows: ["libgmp.lib"]` from the `gmp` lib entry — the unix `gmp` lib is now never reached on Windows.

## Constraints

- **vs17 only.** Both PHP-SDK deps zips are version-pinned. Builds on VS 2019 (vs16) will not find these libs; the PHP-SDK deps server has vs16 variants if that becomes a real need (`mpir-3.0.0-vs16-x64.zip`, `libintl-0.18.3-8-vs16-x64.zip`). Most users targeting current PHP on Windows are on VS 2022 already.
- License pointers are short SPDX-style notes pointing at the canonical LGPL text URLs (LGPL v3 for MPIR, LGPL v2.1 for gettext), matching the style of the existing `gmp` source entry rather than the inlined-text style of `postgresql-win`. Easy follow-up to inline full text if `dump-license` consumers want it.

## Test plan

- [x] PHP lint clean on all three new builder classes
- [x] JSON parses (lib.json, ext.json, source.json)
- [x] `spc download mpir-win,libintl-win` resolves both URLs and downloads successfully
- [x] Zip layouts match builder expectations (`lib/mpir_a.lib` + `include/mpir/{gmp,mpir,gmp-mparam,config}.h`, `lib/libintl_a.lib` + `include/libintl.h`)
- [x] Manually executed both builders' copy logic — produces exactly the files declared in `static-libs-windows` / `headers-windows`
- [x] Header API surface verified (`mpir/gmp.h` exposes `__gmpz_*` etc.; `libintl.h` exposes `gettext()` and `__USE_GNU_GETTEXT`)
- [ ] Full end-to-end Windows build with VS 2022 — \`spc build-libs mpir-win,libintl-win,libxslt\` then \`spc build --build-cli "gmp,gettext,xsl"\` — not yet executed; requires a box with VS 2022 installed

### DIFF
--- a/config/ext.json
+++ b/config/ext.json
@@ -1210,11 +1210,11 @@
     },
     "xsl": {
         "support": {
-            "Windows": "wip",
             "BSD": "wip"
         },
         "type": "builtin",
         "arg-type": "with-path",
+        "arg-type-windows": "with",
         "lib-depends": [
             "libxslt"
         ],
--- a/config/ext.json
+++ b/config/ext.json
@@ -211,13 +211,16 @@
     },
     "gettext": {
         "support": {
-            "Windows": "wip",
             "BSD": "wip"
         },
         "type": "builtin",
         "arg-type": "with-path",
+        "arg-type-windows": "with",
         "lib-depends": [
             "gettext"
+        ],
+        "lib-depends-windows": [
+            "libintl-win"
         ]
     },
     "glfw": {
--- a/config/ext.json
+++ b/config/ext.json
@@ -237,13 +237,16 @@
     },
     "gmp": {
         "support": {
-            "Windows": "wip",
             "BSD": "wip"
         },
         "type": "builtin",
         "arg-type": "with-path",
+        "arg-type-windows": "with",
         "lib-depends": [
             "gmp"
+        ],
+        "lib-depends-windows": [
+            "mpir-win"
         ]
     },
     "gmssl": {
--- a/config/lib.json
+++ b/config/lib.json
@@ -730,6 +730,10 @@
             "libxslt.a",
             "libexslt.a"
         ],
+        "static-libs-windows": [
+            "libxslt.lib",
+            "libexslt.lib"
+        ],
         "lib-depends": [
             "libxml2"
         ]
--- a/config/lib.json
+++ b/config/lib.json
@@ -178,9 +178,6 @@
         "static-libs-unix": [
             "libgmp.a"
         ],
-        "static-libs-windows": [
-            "libgmp.lib"
-        ],
         "headers": [
             "gmp.h"
         ]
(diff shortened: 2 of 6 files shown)


henderkes:
these ought to be three separate PR's and I'm not sure downloading from php vs17 deps is appropriate. Also test the extensions that this affects on windows.

crazywhalecc:
v3 version has implemented all `gmp,xsl,gettext` extension support for Windows. You can just use v3 branch to build.

Our main workflow will switch to v3 soon after first alpha version released.

In addition, other updates are recommended to be made based on v3, as we will soon stop maintaining v2.

luthermonson:
thanks for the help again, i was still just using my fork with the windows static stuff working and didnt think you were ready for people to try v3, ill switch and add that to my pipeline now. i am writing the fixes on a windows box and testing them with windows ci pipelines, there is a high level of confidence that they work but changes like vs17 deps from certain places not being appropriate is feedback i would need. thanks for your time

PR #1147, opened 2026-05-20
Add a Log Filter to Strip Github Tokens

### LINKED ISSUE #1144: Debug logs leak `Authorization: Bearer` headers (GitHub token exposure)
### Summary

When `--debug` is enabled and `GITHUB_TOKEN` is set in the environment, spc's downloader logs the full `curl` invocation — including the `-HAuthorization: Bearer <token>` header — to both the terminal and the persisted log file at `log/spc.output.log`. The token ends up in plaintext anywhere that output is captured (CI logs, screen recordings, support pastebins, transcripts, etc.).

### Reproduction

```
GITHUB_TOKEN=ghp_REDACTED ./spc download zlib --debug
```

Output (abbreviated):

```
[D] [PASSTHRU] curl.exe -sfSL   "-HAuthorization: Bearer ghp_REDACTED" "https://api.github.com/repos/madler/zlib/releases"
```

The same line appears in `log/spc.output.log`.

### Root cause

[`src/globals/functions.php:164`](src/globals/functions.php#L164) calls `logger()->debug('[PASSTHRU] ' . \$cmd)` with the raw command string. The bearer header is constructed at [`src/SPC/store/CurlHook.php:32`](src/SPC/store/CurlHook.php#L32) and added unredacted to the curl args before they reach `f_passthru()`.

### Suggested fix

Redact known sensitive args before logging. Either:

1. Sanitize in `f_passthru()` — regex-replace `-HAuthorization: Bearer \S+` and `-HAuthorization: Basic \S+` with `…: <redacted>` before the debug-log call. Cheapest, covers all callers.
2. Sanitize at the source in `CurlHook` — pass the curl command through a wrapper that knows which args are sensitive.

Option 1 is a 3-line change and catches future leaks of the same shape.

### Impact

Any user who has run `--debug` with `GITHUB_TOKEN` set should rotate the token. CI configurations that capture spc output (artifacts, log uploads) are particularly exposed.

luthermonson:
if you run a build locally with --debug=true and an env var for GITHUB_USER and GITHUB_TOKEN you'll get some passthru logs which include the bearer token for curl requests to github calls. this was caught with local builds as i have those envvars set for quick gh cli usage and claude noticed the token was getting leaked into /log/spc.output.log

now this was with v2 and v3 tried to add a [log filter](https://github.com/crazywhalecc/static-php-cli/blob/v3/src/StaticPHP/Artifact/Downloader/Type/GitHubTokenSetupTrait.php#L19) but overlooked adding the encoded username:token (which is the actual contents of the bearer token) to the filters and sending the passthru logs through the filter and this PR remedies that.

i sent claude through your action logs and it found no leaks so it looks like your tokens havent leaked but you probably need to double check just to be safe. note that this would have leaked through the gha scrubbers as it would have only known to scrub GITHUB_USER and GITHUB_TOKEN and not the b64(user:token).

closes #1144

### DIFF
--- a/src/bootstrap.php
+++ b/src/bootstrap.php
@@ -33,6 +33,23 @@ ConsoleLogger::$date_format = 'H:i:s';
 ConsoleLogger::$format = '[%date% %level_long%] %body%';
 $ob_logger = new ConsoleLogger(LogLevel::WARNING);
 
+$ob_logger->addLogCallback(function ($level, &$output, &$message, &$context, bool $shouldLog) {
+    global $spc_log_filters;
+    if (!is_array($spc_log_filters)) {
+        $spc_log_filters = [];
+    }
+    // filter message and context
+    $output = str_replace($spc_log_filters, '***', $output);
+    $message = str_replace($spc_log_filters, '***', $message);
+    $context = array_map(function ($item) use ($spc_log_filters) {
+        if (is_string($item)) {
+            return str_replace($spc_log_filters, '***', $item);
+        }
+        return $item;
+    }, $context);
+    return true;
+});
+
 // setup log file
 if (filter_var(getenv('SPC_ENABLE_LOG_FILE'), FILTER_VALIDATE_BOOLEAN)) {
     // init spc log files
--- a/src/StaticPHP/Artifact/Downloader/Type/GitHubTokenSetupTrait.php
+++ b/src/StaticPHP/Artifact/Downloader/Type/GitHubTokenSetupTrait.php
@@ -16,8 +16,9 @@ trait GitHubTokenSetupTrait
         // GITHUB_TOKEN support
         if (($token = getenv('GITHUB_TOKEN')) !== false && ($user = getenv('GITHUB_USER')) !== false) {
             logger()->debug("Using 'GITHUB_TOKEN' with user {$user} for authentication");
-            spc_add_log_filter([$user, $token]);
-            return ['Authorization: Basic ' . base64_encode("{$user}:{$token}")];
+            $encoded = base64_encode("{$user}:{$token}");
+            spc_add_log_filter([$user, $token, $encoded]);
+            return ["Authorization: Basic {$encoded}"];
         }
         if (($token = getenv('GITHUB_TOKEN')) !== false) {
             logger()->debug("Using 'GITHUB_TOKEN' for authentication");
--- a/tests/GlobalsFunctionsTest.php
+++ b/tests/GlobalsFunctionsTest.php
@@ -1,0 +1,104 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tests;
+
+use PHPUnit\Framework\TestCase;
+use Psr\Log\LogLevel;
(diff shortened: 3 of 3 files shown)


src/globals/functions.php
```diff
@@ -160,10 +166,11 @@ function f_passthru(string $cmd): ?bool
             break;
         }
     }
+    $logged = spc_filter_log_string($cmd);
     if ($danger) {
-        logger()->notice('Running dangerous command: ' . $cmd);
+        logger()->notice('Running dangerous command: ' . $logged);
```
crazywhalecc: We could filter at ConsoleLogger level in `src/bootstrap.php`, using this we don't need to filter every time:

```php
$ob_logger = new ConsoleLogger(LogLevel::WARNING);
$ob_logger->addLogCallback(function ($level, &$output, &$message, &$context, bool $shouldLog) {
    global $spc_log_filters;
    if (!is_array($spc_log_filters)) {
        $spc_log_filters = [];
    }
    // filter message and context
    $output = str_replace($spc_log_filters, '***', $output);
    $message = str_replace($spc_log_filters, '***', $message);
    $context = array_map(function ($item) use ($spc_log_filters) {
        if (is_string($item)) {
            return str_replace($spc_log_filters, '***', $item);
        }
        return $item;
    }, $context);
    return true;
});
```

luthermonson: i could refactor to this if you prefer

henderkes: yes, please

luthermonson: thanks for the suggestion, tests still work and it was an elegant solution. i havent done php in a long time and didnt even know you could do a call back on consolelogger. i've tuned up some of my claude.md files to be more in line with what you two are looking for with issues/pr text and i'm realizing i spent no time training an agent to work in this repo and all my stuff is preconfigured for go/rust development so i've started to fix that. i promise to be less annoying in the future.

crazywhalecc (approved):

src/bootstrap.php
```diff
@@ -33,6 +33,23 @@
 ConsoleLogger::$format = '[%date% %level_long%] %body%';
 $ob_logger = new ConsoleLogger(LogLevel::WARNING);
 
+$ob_logger->addLogCallback(function ($level, &$output, &$message, &$context, bool $shouldLog) {
+    global $spc_log_filters;
```
henderkes, line 37: please find a way of doing this without abusing globals

henderkes (changes requested):

src/bootstrap.php
crazywhalecc: I think here using global is okay, only for adding log filter (same as `spc_add_log_filter()`). Unless we change every global variables to static variables, but this is another topic.

henderkes (approved):

src/bootstrap.php
henderkes: okay, lets refactor it later

luthermonson:
thanks for the help

PR #1151, opened 2026-05-21
[v3] fix: use ftpmirror.gnu.org instead of ftp.gnu.org

luthermonson:
for some reason the ftp was down for me from my home and there's a perfectly fine mirror that worked fine, slight chance you dont want this one but thought i'd offer.

### DIFF
--- a/config/artifact/ncurses.yml
+++ b/config/artifact/ncurses.yml
@@ -4,5 +4,5 @@ ncurses:
       - COPYING
   source:
     type: filelist
-    url: 'https://ftp.gnu.org/pub/gnu/ncurses/'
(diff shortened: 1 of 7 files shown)


henderkes (approved):
We decided against this a year ago or so and well back to our own hosted mirrors, but since it's really failing a lot lately it's probably a good change.

I'll leave it to @crazywhalecc to merge

crazywhalecc (approved):

crazywhalecc:
Well, looks like ftpmirror.gnu.org's redirect is not very stable though. I often redirected to some address that returns  403 and looks like it also broke our regular download in GHA.

PR #1155, opened 2026-05-22
[v3] fix: remove no-op ulimit call from ZigToolchain

luthermonson:
I'm going to make an argument why this ulimit call is unnecessary, if you call this from php you fork into a new shell and call ulimit there and immediately exit losing the change. The real fix is to call ulimit from the shell before you call spc build or do some pcntl_setrlimit from inside the running php process.

In addition ulimit will silently fail if the value it is trying to set is smaller than the current value, you are probably getting along just fine in your environments not noticing that this is actually failing. I'm running this in a container that was defaulted to 1024 and this call exit with non-zero status code and fails the build.

I think this call should be removed mostly because it does nothing but onus should be placed on the spc build callee and you should let spc build fail with too many open files errors on it's own rather than try to elevate and change rconfigs, those are easily diagnosed problems by sysadmins who can decide the level of elevated access. For instance, my build server can pass in higher rates to the pods so I update my config and move on.

### DIFF
--- a/src/StaticPHP/Toolchain/ZigToolchain.php
+++ b/src/StaticPHP/Toolchain/ZigToolchain.php
@@ -43,7 +43,6 @@ class ZigToolchain implements UnixToolchainInterface
     public function afterInit(): void
     {
         GlobalEnvManager::addPathIfNotExists($this->getPath());
-        f_passthru('ulimit -n 2048'); // zig opens extra file descriptors, so when a lot of extensions are built statically, 1024 is not enough
         $cflags = getenv('SPC_DEFAULT_CFLAGS') ?: '';
         $cxxflags = getenv('SPC_DEFAULT_CXXFLAGS') ?: '';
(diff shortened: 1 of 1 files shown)


henderkes:
addressed in v3/pgo PR

luthermonson:
should i be building with v3 + pgo to not waste your time? if that's the next big pr for v3 can I start testing it

henderkes:
probably, yes. still a lot of things to do before v3 but it should make things buildable like on v2 in theory

luthermonson:
if you guys would like to work faster i can join you on discord or slack or something, it might be quicker. my email in my profile if you want to send me an invite to something. i will try with pgo and let you know, looks like you found the gettext + windows fix over there which i was about to prepare so ill give it a full run through and make my PRs against feat/pgo-v3. thanks for the guidance

henderkes:
Check the last commits here for the discord link, but we mainly communicate via GitHub.

PR #1156, opened 2026-05-22
mongodb: export PHP_VERSION_ID for in-tree builds

luthermonson:
mongo-php-driver 2.3.3 added a config.m4 check that falls back to php-config when PHP_VERSION_ID is unset in the shell env. In-tree PHP source builds have no php-config, so configure fails with:

    checking PHP version... configure: error: php-config not found

Set PHP_VERSION_ID from main/php_version.h before configureForUnix so the lookup short-circuits.


henderkes:
thank you

PR #1165, opened 2026-05-24
ext: add ext-fastchart and ext-fastjson registry entries

henderkes:


henderkes:
(tests fail because a new version is pending, I didn't feel like adding the patch here that will be useless next release in a few days

crazywhalecc:
> (tests fail because a new version is pending, I didn't feel like adding the patch here that will be useless next release in a few days

We have too many upstream dependencies to manage. Unless it's a version that can be released immediately, I think it's more appropriate to update at our own pace from a reasonable perspective.

But for this, simply running mongodb as a standalone CI would be good.

crazywhalecc:
And fastchart seems failed to build in workflows.

henderkes:
> And fastchart seems failed to build in workflows.

yes that's pending release: https://github.com/iliaal/fastchart/pull/5

@iliaal will probably tag a new release soon. Same goes for clickhouse.

crazywhalecc (approved):

PR #1169, opened 2026-05-24
[v3] artifact: separate extract and install step

henderkes:
Give zig, rust, go_win and go_xcaddy a small consistent surface for locating the install directory and a binary inside it:

- path(): install/extract root for the artifact
- binary($name = '<default>'): full path to a binary under that root, picking the artifact's natural layout (top-level for zig, bin/ for rust and the go toolchains)
- isInstalled(): is the default binary present on disk

Callers that previously concatenated PKG_ROOT_PATH . '/zig/zig' (and the equivalents for the other artifacts) by hand can call the helpers instead, and any later code that needs to ask "is this toolchain available" can use isInstalled() without rebuilding the path.

## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`


crazywhalecc (commented):
I'll leave this tomorrow. Look for a better way if possible.

crazywhalecc:
```
go-win:
  type: target
  artifact:
    binary: custom
    metadata:
      install-path: go-win
      default-binary: go.exe
      binary-subdir: bin
```

I prefer setting them in config and using `getInstallPath()` and other similar functions to get. Soft-defining functions in artifact itself looks not appropriate.

henderkes:
This is something where I don't really see an advantage to describing it in a config. The function definitions are at least as short as and more ergonomic to use.

I'm not terribly concerned with it though, so if you want to change it, I don't object.

crazywhalecc:
> This is something where I don't really see an advantage to describing it in a config. The function definitions are at least as short as and more ergonomic to use.
> 
> I'm not terribly concerned with it though, so if you want to change it, I don't object.

I just thought that we should keep constant things in config, and dynamic things written in PHP functions.

I don't see your parts containing dynamic codes so that I prefer defining in config. Also we could define dynamic things in another parts as you like.

crazywhalecc:
@henderkes Frankly, the real problem here is that we should treat the tools that the build process depends on as independent package types: `tool`, rather than `target`, and then extend the corresponding methods on `tool`.

While this might make the `DependencyResolver` slightly more complex, it provides a clearer understanding of the tools themselves. We should not to make many architectural changes before a clear type system is established.

crazywhalecc:
This PR is temporarily closed for the following reasons:

- The currently implemented toolkit type does not support the original purpose of this PR.
- Modifying the extractor and installer would not yield significant benefits.

PR #1172, opened 2026-05-24
llvm-tools: build host llvm-objcopy/strip/profdata under ZigToolchain

henderkes:
Add an `llvm-tools` target artifact that downloads llvm-project source matching the version of clang shipped by the active zig install, builds llvm-objcopy, llvm-strip and llvm-profdata into PKG_ROOT_PATH/llvm-tools/bin, and exposes them through the same path/binary/isInstalled static surface as the other artifacts.

A new LlvmToolsCheck doctor item runs when the active toolchain is ZigToolchain and reports whether the three tools are built, with a fix that installs the package and runs the build.

PackageBuilder now picks the right tool when the active toolchain is ZigToolchain:
- extractDebugInfo() honours OBJCOPY from the environment, then falls back to llvm-tools' llvm-objcopy under Zig and plain objcopy otherwise.
- stripBinary() uses llvm-strip under Zig and plain strip otherwise.

System strip/objcopy refuse zig-produced archives and bitcode sections, so without this the strip stage breaks LTO builds. Other toolchains keep using the system binaries.

ApplicationContext::tryGet() wraps the container's get() in a try/catch and returns null on failure, so PackageBuilder can ask "which toolchain is active right now" without PHP-DI throwing on autowirable-but-unconstructable classes.

Depends on v3c/artifact-static-helpers (uses zig::isInstalled() and zig::binary()).

## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`

### DIFF
--- a/config/pkg/target/llvm-compiler-rt.yml
+++ b/config/pkg/target/llvm-compiler-rt.yml
@@ -1,0 +1,10 @@
+llvm-compiler-rt:
+  type: target
+  artifact:
+    source:
+      type: custom
+    metadata:
+      license-files: [LICENSE.TXT]
+      license: Apache-2.0-with-LLVM-exception
+  depends:
+    - zig
--- a/config/pkg/target/llvm-compiler-rt.yml
+++ b/config/pkg/target/llvm-compiler-rt.yml
@@ -6,5 +6,5 @@ llvm-compiler-rt:
     metadata:
       license-files: [LICENSE.TXT]
       license: Apache-2.0-with-LLVM-exception
-  depends:
+  tools:
     - zig
--- a/config/pkg/target/llvm-tools.yml
+++ b/config/pkg/target/llvm-tools.yml
@@ -1,0 +1,6 @@
+llvm-tools:
+  type: target
+  artifact:
+    binary: custom
+  depends:
+    - zig
--- a/config/pkg/tool/llvm-tools.yml
+++ b/config/pkg/tool/llvm-tools.yml
@@ -1,0 +1,16 @@
+llvm-tools:
+  type: tool
+  artifact:
+    source:
+      type: custom
+    binary: custom
+    metadata:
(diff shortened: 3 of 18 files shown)


crazywhalecc:
@henderkes Looks using `tool` instead of `target` is a better approach.

henderkes:
Yes, I'll need to rework this.

config/pkg/target/llvm-compiler-rt.yml
```diff
@@ -0,0 +1,10 @@
+llvm-compiler-rt:
+  type: target
+  artifact:
+    source:
+      type: custom
+    metadata:
+      license-files: [LICENSE.TXT]
+      license: Apache-2.0-with-LLVM-exception
+  depends:
+    - zig
```
crazywhalecc:
```suggestion
  tools:
    - zig
```

Should be `tools: [zig]`

crazywhalecc (changes requested):
Most of it seems fine, but I don't think installing these two packages by default offers any benefit. Most people don't need PGO and shared builds. They could be triggered under specific conditions, similar to `--with-upx-pack` (either in `doctor` or the toolchain).

henderkes:
> Most of it seems fine, but I don't think installing these two packages by default offers any benefit. Most people don't need PGO and shared builds. They could be triggered under specific conditions, similar to `--with-upx-pack` (either in `doctor` or the toolchain).

Do you want to scan cflags for lto/pgo instead and error out?

libcompiler_rt needs to be built for cpu_model anyway, though. Many libraries/extensions use builtin_cpu_supports and building it only takes a few seconds anyways.

henderkes:
Forgot to mention, but gnu objcopy also can't work for cross compilation (e.g. from aarch64 to arm32). I think if we host the tools it's preferable to always have.

PR #1173, opened 2026-05-28
[v3] - fix(windows): define LEXBOR_STATIC for PHP 8.5+ static builds

luthermonson:
The ext/uri extension introduced in PHP 8.5 vendors lexbor headers. On Windows, lexbor's LXB_API macro defaults to __declspec(dllimport), which breaks static linking with unresolved external symbols for all lxb_* functions.

Adding a BeforeStage patch that injects /D LEXBOR_STATIC into CFLAGS_URI makes LXB_API expand to nothing, allowing the uri extension to link into the static PHP build.

Only applies to PHP >= 8.5.0 where ext/uri exists.

### DIFF
--- a/src/Package/Target/php/windows.php
+++ b/src/Package/Target/php/windows.php
@@ -226,6 +226,21 @@ trait windows
         $this->deployWindowsBinary($builder, $package, 'php-cgi');
     }
 
+    #[BeforeStage('php', [self::class, 'makeForWindows'])]
+    #[PatchDescription('Define LEXBOR_STATIC for static builds (PHP 8.5+ uri extension)')]
+    public function patchLexborStaticDefine(TargetPackage $package): void
+    {
+        if ($this->getPHPVersionID() < 80500) {
+            return;
+        }
+        $makefile_path = "{$package->getSourceDir()}\\Makefile";
+        $content = FileSystem::readFile($makefile_path);
+        // ext/uri references lexbor headers which default to __declspec(dllimport) on Windows.
+        // LEXBOR_STATIC makes LXB_API expand to nothing, required for static linking.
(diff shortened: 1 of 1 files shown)


henderkes:
Isn't this better passed as a -D flag directly on configure? And certainly, this should be fixed in php-src, not here.

luthermonson:
i was trying to stay in windows and not step on too many toes, i can refactor

henderkes:
php-src has different configure methods for windows (configure.js) and unix (makefile based).

luthermonson:
im really not familiar with upstream php-src yet just from what i'm exposed through spc. if i make these changes upstream can we get them in here pretty quick as it's a build change or will it take a while to sort through in a real release of 8.x?

luthermonson:
welp claude just walked me through it. thanks for the advice. i swear i'll be good at this one day. i think we can close #1173 and #1176 and i'll get php-src PRs up and ill link them back here before i close

luthermonson:
https://github.com/php/php-src/pull/22165

henderkes:
> im really not familiar with upstream php-src yet just from what i'm exposed through spc. if i make these changes upstream can we get them in here pretty quick as it's a build change or will it take a while to sort through in a real release of 8.x?

It just happens automatically through the next release of 8.5.x (or whatever version you created the PR against, you should always target the oldest supported version the change is relevant to, usually 8.4 now, 8.2 for security fixes).

> welp claude just walked me through it. thanks for the advice. i swear i'll be good at this one day. i think we can close #1173 and #1176 and i'll get php-src PRs up and ill link them back here before i close

If you're genuinely looking to become good at this, I would recommend leaving llms out of the process for now, because while they'll get things done, it's often times not in the fashion it should be done and the learning effect is minimal.

PR #1174, opened 2026-05-28
[v3] - fix(windows): resolve doctor-installed perl and quote perl path

luthermonson:
Two related fixes for OpenSSL builds on Windows:

1. spc doctor installs Strawberry Perl under `pkgroot/strawberry-perl/`, but the validate() check only looks at the arch-specific native path (`strawberry-perl-x86_64-win`). When a user runs `spc doctor` to install perl, the subsequent build fails because the doctor's install layout isn't on the lookup list. Add it as a fallback before falling through to WindowsUtil::findCommand.

2. Wrap the resolved perl path in quote() when building the Configure command. WindowsUtil::findCommand can return a path containing spaces (e.g. `C:\Program Files\...`), which then splits the cmd.exe invocation. Quoting matches how every other interpolated path in this command is handled.


src/Package/Library/openssl.php
```diff
@@ -25,7 +25,12 @@ public function validate(): void
         if (SystemTarget::getTargetOS() === 'Windows') {
             global $argv;
             $perl_path_native = PKG_ROOT_PATH . '\strawberry-perl-' . arch2gnu(php_uname('m')) . '-win\perl\bin\perl.exe';
-            $perl = file_exists($perl_path_native) ? ($perl_path_native) : WindowsUtil::findCommand('perl.exe');
+            $perl_path_doctor = PKG_ROOT_PATH . '\strawberry-perl\perl\bin\perl.exe';
```
henderkes: We should have *one* pkgroot place, not two different ones to juggle.

henderkes (changes requested):

src/Package/Library/openssl.php
luthermonson: i found it with native and doctor, are you asking me to refactor to only use doctor? the findCommand will pull in random things, I think you're asking me to drop all the fluff and just use the strawberry perl pkgroot?

henderkes: No, WindowsUtil::findCommand('perl.exe') is fine. But now you're checking TWO different pkg_root paths that could both only ever be created by static-php itself. Other than the host binary, we should only have one.-

luthermonson: i see your point, dont add the doctor path, just update the native path. sure

luthermonson: think i refactored as you asked, please open new feedback if you dont like it

henderkes (commented):
Looks reasonable to me now, but this would also do well with the package api thing I dropped in another PR (on unix only for now, but I think @crazywhalecc wanted to change it to yml declaration, so this would automatically be possible here too). Or it could be added to PATH instead.

Since I'm not looking to take up windows specific maintenance too, I'll delegate to cc here.

crazywhalecc:
Looks good so far, but some details need improvement. I'll do some other works and  improve some modules for v3 base first though. Let me keep this open.

crazywhalecc (approved):

henderkes (approved):

PR #1175, opened 2026-05-28
[v3] - fix(windows): use 7za-only archive extraction for compressed tarballs

luthermonson:
The previous implementation piped 7za output to Windows' built-in tar.exe to handle gz/xz/bz2 tarballs. This fails on certain archives containing symlinks: Windows tar.exe cannot create symlinks unless the build is running with Developer Mode enabled or with admin privileges, and a failure mid-extract leaves the source tree in a broken state.

Replace the pipeline with a two-pass 7za extraction (compressed wrapper then inner .tar) into a temp directory, with strip-components handled in PHP. 7za silently skips symlinks during tar extraction, which is the desired behavior for spc's source preparation - the symlinks in question are documentation or build-system artifacts that aren't needed for the static build.

### DIFF
--- a/src/StaticPHP/Runtime/Shell/DefaultShell.php
+++ b/src/StaticPHP/Runtime/Shell/DefaultShell.php
@@ -202,14 +202,57 @@ class DefaultShell extends Shell
         };
 
         $extname = FileSystem::extname($archive_path);
-        $tar = SystemTarget::isUnix() ? 'tar' : '"C:\Windows\system32\tar.exe"';
 
         match ($extname) {
             'tar' => $this->executeTarExtract($archive_path, $target_path, 'none'),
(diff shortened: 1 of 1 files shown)


crazywhalecc:
Same as #1174 . BTW I plan to replace php-sdk-binary-tools to msys2 tools because php-sdk-binary-tools' msys2 is missing some components that we may have to use in the future. It will also need to download separated 7z executable first in our toolchain. But I'll handle it later.

And for full-7za instead of `7za.exe|tar.exe`, it's a good way in SPC, but also need some improvement, like `FileSystem::copyDir` is just slow.

crazywhalecc:
This issue is still relevant on Windows, but the current v3 branch handles symlink-only extraction failures through passthruTolerateSymlinks(). The 7za executable is now independently provided by the 7za-win tool package after the MSYS2 migration. Since the original two-pass extraction also adds an expensive directory copy and only approximates --strip-components=1, this PR is no longer needed and can be closed as superseded.

PR #1176, opened 2026-05-28
[v3] - feat(windows): static embed SAPI build support

luthermonson:
From the beginning I couldn't really get windows static builds to work in v3 as you can probably tell by my 4 PRs tonight, if for some reason these changes are not welcome because I'm missing something that already works in v3 please let me know. 

---

Enable building and validating php8embed as a fat static library on Windows (no DLL dependency at runtime). This consists of three Makefile patches and a rewritten compile-time smoke test.

1. CFLAGS_EMBED export defines

   Embed SAPI objects are compiled as DLL consumers by default, so calls into PHP/Zend APIs go through __imp_ thunks. For a fat static lib they need direct linkage. Inject PHP_EXPORTS, LIBZEND_EXPORTS, SAPI_EXPORTS, TSRM_EXPORTS into CFLAGS_EMBED so php_embed.obj references symbols directly.

2. Strip duplicate TSRM cache definition from php_embed.c

   In DLL builds each SAPI has its own _tsrm_ls_cache via TSRMLS_CACHE_DEFINE(). In our fat static lib all objects share one binary, so php_embed.c's definition collides with zend.c's, producing LNK4006 and a corrupt binary. Remove the duplicate in-source so php_embed.obj uses zend.c's copy.

3. Rewrite embed smoke test

   The PHP_EMBED_START_BLOCK/END_BLOCK macros expand to setjmp/longjmp and a zend_first_try wrapper that doesn't work cleanly when the embed lib's TSRM cache was stripped above. Use the explicit php_embed_init / zend_first_try / php_embed_shutdown form instead.

   Also:
   - Add PHP_EXPORTS/LIBZEND_EXPORTS/SAPI_EXPORTS/TSRM_EXPORTS to the test compile flags (mirroring CFLAGS_EMBED).
   - Add pathcch.lib (PHP 8.x main/streams calls PathCchCanonicalizeEx).
   - Remove /FORCE:MULTIPLE from the link line - no longer needed since the TSRM cache duplicate is fixed at source above.
   - Use str_contains() for output validation instead of exact match, so optional stderr noise from the loaded extensions doesn't break the test.

### DIFF
--- a/src/Package/Target/php/windows.php
+++ b/src/Package/Target/php/windows.php
@@ -317,6 +317,25 @@ trait windows
             $content
         );
 
+        // Embed SAPI objects are compiled as DLL consumers (dllimport for PHP/Zend APIs).
+        // For our fat static lib, they need direct linkage. Add export defines to CFLAGS_EMBED
+        // so php_embed.obj references symbols directly instead of through __imp_ thunks.
+        $content = preg_replace(
+            '/^CFLAGS_EMBED=(.+)$/m',
+            'CFLAGS_EMBED=$1 /D PHP_EXPORTS /D LIBZEND_EXPORTS /D SAPI_EXPORTS /D TSRM_EXPORTS',
+            $content,
+            1
+        );
+
+        // In DLL builds each SAPI has its own _tsrm_ls_cache via TSRMLS_CACHE_DEFINE().
+        // In our fat static lib all objects share one binary, so the duplicate TLS definition
+        // from php_embed.c collides with the one in zend.c, producing LNK4006 and a corrupt
+        // binary. Remove it so php_embed.obj uses the core's copy.
+        $embed_src = "{$package->getSourceDir()}\\sapi\\embed\\php_embed.c";
+        if (file_exists($embed_src)) {
+            FileSystem::replaceFileStr($embed_src, 'ZEND_TSRMLS_CACHE_DEFINE()', '/* removed for static embed */');
+        }
+
         // Patch embed lib target to build a REAL static library instead of just an import lib.
         // The default embed target only includes embed SAPI objects and links against php8.lib (import lib).
         // We need to include PHP core objects (PHP_GLOBAL_OBJS) and static extension objects (STATIC_EXT_OBJS)
@@ -646,16 +665,17 @@ HEADER;
 #include <sapi/embed/php_embed.h>
 
 int main(int argc, char **argv) {
-    PHP_EMBED_START_BLOCK(argc, argv)
-
-    zend_file_handle file_handle;
-    zend_stream_init_filename(&file_handle, "embed.php");
-
-    if (!php_execute_script(&file_handle)) {
-        php_printf("Failed to execute PHP script.\n");
+    if (php_embed_init(argc, argv) == FAILURE) {
+        return 1;
     }
 
-    PHP_EMBED_END_BLOCK()
+    zend_first_try {
+        zend_file_handle file_handle;
+        zend_stream_init_filename(&file_handle, "embed.php");
+        php_execute_script(&file_handle);
+    } zend_end_try();
+
+    php_embed_shutdown();
     return 0;
 }
 C_CODE;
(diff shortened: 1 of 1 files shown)


src/Package/Target/php/windows.php
```diff
@@ -317,6 +317,25 @@ public function patchEmbedTarget(TargetPackage $package): void
...
+        );
+
+        // In DLL builds each SAPI has its own _tsrm_ls_cache via TSRMLS_CACHE_DEFINE().
+        // In our fat static lib all objects share one binary, so the duplicate TLS definition
+        // from php_embed.c collides with the one in zend.c, producing LNK4006 and a corrupt
+        // binary. Remove it so php_embed.obj uses the core's copy.
+        $embed_src = "{$package->getSourceDir()}\\sapi\\embed\\php_embed.c";
+        if (file_exists($embed_src)) {
+            FileSystem::replaceFileStr($embed_src, 'ZEND_TSRMLS_CACHE_DEFINE()', '/* removed for static embed */');
+        }
```
henderkes, line 337: like in the other PR, patching here is definitely not the correct approach, instead passing the defined to configure

but even here it begs the question whether this shouldn't be fixed upstream instead

src/Package/Target/php/windows.php
```diff
@@ -718,9 +738,9 @@ public function smokeTestEmbedForWindows(PackageInstaller $installer, TargetPack
         InteractiveTerm::setMessage('Running php-embed run smoke test');
         [$ret, $output] = cmd()->cd($test_dir)->execWithResult('embed.exe');
         $raw_output = implode('', $output);
-        if ($ret !== 0 || trim($raw_output) !== 'hello') {
+        if ($ret !== 0 || !str_contains($raw_output, 'hello')) {
```
henderkes, line 741: why this change?

src/Package/Target/php/windows.php
```diff
@@ -646,16 +665,17 @@ public function smokeTestEmbedForWindows(PackageInstaller $installer, TargetPack
...
 
-    PHP_EMBED_END_BLOCK()
+    zend_first_try {
+        zend_file_handle file_handle;
+        zend_stream_init_filename(&file_handle, "embed.php");
+        php_execute_script(&file_handle);
+    } zend_end_try();
+
+    php_embed_shutdown();
     return 0;
```
henderkes, line 679: what's the point?

src/Package/Target/php/windows.php
luthermonson: i was getting extensions printing warnings on startup, this got rid of the noise. you have to static build with extension to see this happen and it was quicker to confirm we actually got the hello than debug the extensions to get the test working again. i wouldn't back this change out until maybe i spend some time figuring out what the extensions are doing. i can add a comment as to why this was done if  that helps

src/Package/Target/php/windows.php
luthermonson: upstream php? oof... i wouldnt even know how to get that through

src/Package/Target/php/windows.php
luthermonson: felt the test mirrored a realistic call shape for an embed user it's more in line how i'm using it and frankenphp uses it. if you hate it i can figure something else out

src/Package/Target/php/windows.php
henderkes: It's intended to fail if we're getting warnings printed.

src/Package/Target/php/windows.php
henderkes: Perhaps, but it has no bearing on the actual test, so the change shouldn't happen in a different PR for purely aesthetic reasons

src/Package/Target/php/windows.php
henderkes: by adding the same flags you're adding here conditionally in the configure.js build script :)

src/Package/Target/php/windows.php
luthermonson: understood, ill switch it back. didn't understand the intention

src/Package/Target/php/windows.php
luthermonson: sage advice, ill stop trying to keep everything in spc + windows and expand my horizons a bit

luthermonson:
We are going to do this upstream 😨 ill post linked PRs when I get them up and close

luthermonson:
https://github.com/php/php-src/pull/22166 thanks again for the help @henderkes i think this is right based on the hacks that worked here to get builds working again. let's see if they get through.

PR #1179, opened 2026-06-01
Add macports support

beryllium:
## What does this PR do?

This PR attempts to add support for operating in a MacPorts environment, if detected & if a Homebrew environment is not found.

(I do not use Homebrew at all, so I was unable to use spc until I addressed this in my local checkout.)

I have a concern that this change may not play nicely on a system that has Homebrew but does not have MacPorts, as I was only able to test the inverse of that. Perhaps I should add a fixPort method that detects if brew is installed (preferably in a non-circular way), and no-ops the macports check failure? Or will the lack of a fixPort method accomplish this implicitly?

Also, this is branched off of `v3`, not `feat/pgo-v3`. Unsure if that complicates things, but hopefully it isn't an issue.

## Checklist before merging

  - [x] `composer cs-fix`
  - [x] `composer analyse`
  - [x] `composer test`
  - [x] `bin/spc dev:lint-config`

### DIFF
--- a/src/StaticPHP/Doctor/Item/MacOSToolCheck.php
+++ b/src/StaticPHP/Doctor/Item/MacOSToolCheck.php
@@ -45,6 +45,18 @@ class MacOSToolCheck
         return CheckResult::ok();
     }
 
+    #[CheckItem('if macports has installed', limit_os: 'Darwin', level: 998)]
+    public function checkPorts(): ?CheckResult
+    {
+        if (($path = MacOSUtil::findCommand('port')) === null) {
+            return CheckResult::fail('MacPorts is not installed', 'port');
+        }
+        if ($path !== '/opt/local/bin/port' && getenv('GNU_ARCH') === 'aarch64') {
+            return CheckResult::fail('Current macports (/opt/local/bin/port) is not installed for M1 Mac, please re-install macports!');
+        }
+        return CheckResult::ok();
+    }
+
     #[CheckItem('if necessary tools are installed', limit_os: 'Darwin')]
     public function checkCliTools(): ?CheckResult
     {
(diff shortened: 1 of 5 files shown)


henderkes:
Generally looks good, but I don't have a mac to check if macports has recent enough versions of tools. Are they quick to update like brew?

src/StaticPHP/Doctor/Item/MacOSToolCheck.php
```diff
@@ -45,6 +45,15 @@ public function checkBrew(): ?CheckResult
         return CheckResult::ok();
     }
 
+    #[CheckItem('if macports has installed', limit_os: 'Darwin', level: 998)]
+    public function checkPorts(): ?CheckResult
+    {
+        if (MacOSUtil::findCommand('port') === null) {
+            return CheckResult::fail('MacPorts is not installed', 'port');
+        }
```
crazywhalecc: There's no reason that we need to always install homebrew and macports both at the same time

crazywhalecc (changes requested):

src/StaticPHP/Doctor/Item/MacOSToolCheck.php
crazywhalecc: I think it's better to only revert to MacPorts when Homebrew is unavailable. If neither is installed, I believe it's better to prioritize Homebrew (because it has a larger user base).

beryllium: I very much agree, but I wasn't sure how to implement that within the checkitem/fixitem scaffolding.

beryllium:
> _Generally looks good, but I don't have a mac to check if macports has recent enough versions of tools. Are they quick to update like brew?_

There was a period of quite some time where their PHP coverage lagged behind, but now they have a dedicated maintainer for PHP stuff so it is much more responsive. Currently at 8.5.4 for their PHP 8.5 support (though PHP itself is at 8.5.6).

LLVM, as another example, is at 22.1.6, which matches the LLVM project itself.

henderkes:
If they're at llvm 22 already I see no reason not to support macports.

src/StaticPHP/Doctor/Item/MacOSToolCheck.php
crazywhalecc: We could update homebrew check item to `check if homebrew or macports installed`, and fallback to macports check when homebrew is not installed. If neither is installed, automatic repair should install Homebrew. 

In tools installation fix item, use homebrew first and fallback to macports in a similar way.

beryllium: I've applied the suggested change. It seems to work on my system.

crazywhalecc (approved):

PR #1195, opened 2026-06-25
pin zstd, proper fix: https://github.com/kjdev/php-ext-zstd/pull/100

henderkes:
## What does this PR do?

fix https://github.com/php/frankenphp/pull/2364#issuecomment-4796853191

## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`

### DIFF
--- a/config/source.json
+++ b/config/source.json
@@ -333,10 +333,9 @@
         }
     },
     "ext-zstd": {
-        "type": "git",
+        "type": "ghtagtar",
+        "repo": "kjdev/php-ext-zstd",
         "path": "php-src/ext/zstd",
-        "rev": "master",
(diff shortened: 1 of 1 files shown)


crazywhalecc (approved):

henderkes:
@crazywhalecc do we need to update the version to 2.8.6 for the `spc` binary to update? Or does it get rebuilt daily.

crazywhalecc:
Of course we can, but we havent bump version number though. SPC binary from dl.static-php.dev nightly dir is rebuilding daily.

henderkes:
I think we use nightly in frankenphp so should be ok. I'll retrigger CI tomorrow.

PR #1199, opened 2026-07-06
CRLF to LF

crazywhalecc:
## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`


henderkes (approved):
should add a .gitattributes file that sets *.* line endings to LF

.gitattributes
crazywhalecc, line 1: Added here.

henderkes: I should've looked x)

PR #1207, opened 2026-07-08
Fable v3 windows

henderkes:
## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`


crazywhalecc:
postgresql build depends on meson that we haven't installed in our procedure. Either install in msys2, or add a separate `tool` package to install, if we really need to build pgsql from source.

m-this:
Fixes for the three failing Windows tests are in #1208 (four separate bugs, details over there). Hope it helps getting this branch merged 🙂

crazywhalecc:
Even in `C:\im`, we still have long command line issue: https://github.com/crazywhalecc/static-php-cli/actions/runs/28992053641/job/86034014713

henderkes:
Freaking Windows 😆 
https://stackoverflow.com/questions/40727748/windows-clang-command-line-too-long

crazywhalecc:
DigitalOcean spaces exceeds 1024GB/month bandwidth and RackNerd VPS looks not stable though. Perhaps it's worth considering switching to Cloudflare R2 and enabling CDN.

henderkes:
Digital Ocean is limited to just 1TB? I exceed that on just my package delivery.

If Cloudflare R2 isn't too expensive, perhaps. Though we don't really need CDN, just stable connection for the GH runners.

m-this:
Maybe using a response file like in php-src’s Makefile would work for the command length limit on Windows ?clang and lld both accept @libs.rsp, php-src's Windows Makefile already does that for its lists (the *_RESP variables).

m-this:
Opened #1209 for the cmd command line limit that killed the imagick [frankenphp] job.

crazywhalecc (approved):

PR #1208, opened 2026-07-08
Fix the Windows extension test failures on fable-v3-windows

m-this:
Dug into the three Windows failures, turned out to be four separate bugs. All fixed here and verified green on `windows-latest` with the same build commands the test bot runs: https://github.com/m-this/static-php-cli/actions/runs/28927716538

- **postgresql** (breaks pdo_pgsql,pgsql and swow): the meson build expects meson/ninja/win_flex/win_bison on PATH, CI doesn't have them. They're now tool packages via `tools@windows`, like jom on openssl. Meson gets installed from its wheel into a venv, since postgres needs a real meson.exe (it re-invokes meson through `find_program`).
- **imagick**: php.exe was missing `gdi32.lib`, and php-cgi/micro then hit LNK2005 on duplicate symbols; they now link with `/FORCE:MULTIPLE` like php.exe already did.
- **libiconv**: the smoke test failed with DLL_NOT_FOUND because the import lib was on the link line next to the static one, so php.exe imported libiconv.dll. Only the static lib is declared now.
- **pgsql**: libpq needs secur32 and libcrypto needs crypt32, and nothing links them when the openssl extension isn't in the build. Added them in `ext/pgsql/config.w32`, same trick as your imagick patch.

### DIFF
--- a/src/Package/Artifact/meson_win.php
+++ b/src/Package/Artifact/meson_win.php
@@ -22,8 +22,11 @@ class meson_win
         $dir = dirname($target_path);
         $venv = "{$dir}\\venv";
 
+        // Prefer the python-win tool package (installed alongside via tools@windows),
+        // fall back to whatever Python the machine already has.
+        $candidates = ['"' . PKG_ROOT_PATH . '\python-win\tools\python.exe"', 'python', 'py -3'];
         $python = null;
-        foreach (['python', 'py -3'] as $candidate) {
+        foreach ($candidates as $candidate) {
             [$code] = cmd()->execWithResult("{$candidate} --version", false);
             if ($code === 0) {
                 $python = $candidate;
@@ -31,7 +34,7 @@ class meson_win
             }
         }
         if ($python === null) {
-            throw new EnvironmentException('meson needs Python 3 on PATH. Install it from https://www.python.org or with `winget install Python.Python.3.13`.');
+            throw new EnvironmentException('meson needs Python 3; the python-win tool package did not install and no system Python was found.');
         }
 
         if (is_dir($venv)) {
--- a/src/Package/Artifact/meson_win.php
+++ b/src/Package/Artifact/meson_win.php
@@ -1,0 +1,45 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Package\Artifact;
+
+use StaticPHP\Attribute\Artifact\AfterBinaryExtract;
+use StaticPHP\Exception\EnvironmentException;
+use StaticPHP\Util\FileSystem;
+
+class meson_win
+{
+    /**
+     * Meson 1.11 dropped the Windows MSI, so the artifact is the wheel, installed offline
+     * into a private venv. The venv is not just packaging hygiene: pip generates a real
+     * meson.exe launcher there, and PostgreSQL needs one because its build re-invokes meson
+     * through find_program(), which rejects non-executables like the meson.pyz zipapp.
+     */
+    #[AfterBinaryExtract('meson', ['windows-x86_64'])]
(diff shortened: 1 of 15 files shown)


henderkes (approved):

henderkes:
Another possibility would be structuring to use winget/vcpkg to install these tools, as they don't affect the library or php binary links.

`winget install -e --id mesonbuild.meson` in doctor.

src/Package/Artifact/meson_win.php
crazywhalecc, line 1: Not test yet. Have you tried installing Meson and Ninja directly using MSYS2?

crazywhalecc: EDIT: well, meson on MSYS2 suffers from significant issues regarding paths and system library resolution. It may not be as straightforward as the current standalone version.

src/Package/Artifact/meson_win.php
```diff
@@ -0,0 +1,45 @@
...
+        $python = null;
+        foreach (['python', 'py -3'] as $candidate) {
+            [$code] = cmd()->execWithResult("{$candidate} --version", false);
+            if ($code === 0) {
+                $python = $candidate;
+                break;
+            }
+        }
+        if ($python === null) {
+            throw new EnvironmentException('meson needs Python 3 on PATH. Install it from https://www.python.org or with `winget install Python.Python.3.13`.');
```
crazywhalecc: I'd prefer bundling the `python-win` tools with the installation, or embedding the Python installation into the Meson setup process. Otherwise, there would be the issue of requiring manual environment configuration.

Ideally, if the build process is overly complex, I would prefer handling the build in the cloud (similar to `icu-static-win`) or simply downloading a pre-built package like before. However, if installing a lightweight Python runtime solves the problem and not struggle the procedure, I think that would be acceptable as well.

henderkes (changes requested):
I think if we have that kind of dependency not handled by doctor, it should be. Winget is available on every Windows 10+ machine now so we really should be using it.

crazywhalecc:
> I think if we have that kind of dependency not handled by doctor, it should be. Winget is available on every Windows 10+ machine now so we really should be using it.

I have been using the LTSC version of Windows for testing, and it does not include winget.

I am not entirely sure about the coverage here, but if we intend to use winget, we should first check for its presence and ensure native support within our package system (we could also include msys2 package together).

Native support means we could declare these packages as libs' tool dependencies like we just added `tool` type.

src/Package/Artifact/meson_win.php
m-this: I think that's the cleanest option actually, so I went with it: there's now a `python-win` tool package that just unzips the official python.org nuget package into the pkgroot. It's a plain zip, nothing to install, and it ships venv so meson can build its env from it (with a fallback to system python just in case). Should also work on your LTSC since there's no winget or msys2 involved.

Checked on windows-latest that the venv really comes from the pkgroot python and not the runner's one: https://github.com/m-this/static-php-cli/actions/runs/28941945448

henderkes:
The LTSC versions are still missing it? It was a bit of a pain to install manually I remember.

It would definitely simplify the dependency stuff though, both in terms of doctor fixing VS build tools as well as perl, python, meson and cmake. If you think it's acceptable to have users install winget if they don't have it, I think we should do it.

crazywhalecc:
Since Python is already installed, I think sticking with the current approach is sufficient for now. Even if we were to switch to winget, that should be handled as an improvement in a future branch.

henderkes (approved):

crazywhalecc (approved):

PR #1210, opened 2026-07-13
feat(manifest): implement BuildManifestDumper for generating build manifest

crazywhalecc:
## What does this PR do?

Implement BuildManifestDumper for generating build manifest.

This would fix missing build manifest bug when migrating our pre-built php workflows to v3. Also, this will be a good start for adding full SBOM attestation.

## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`

### DIFF
--- a/.github/workflows/build-unix.yml
+++ b/.github/workflows/build-unix.yml
@@ -331,5 +331,4 @@ jobs:
         with:
           name: build-meta-${{ inputs.php-version }}-${{ inputs.os }}
           path: |
-            buildroot/build-extensions.json
-            buildroot/build-libraries.json
+            buildroot/build-manifest.json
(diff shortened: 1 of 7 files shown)


henderkes (commented):
I think the manifest should contain:

all libraries, extensions, tools, in the versions used

all environment variables to reproduce an exact build

crazywhalecc:
Well, I plan to add them gradually. This PR only add a simple manifest interface and basic information. 

We are still missing many prerequisites, such as locking the version and obtaining the version of each library, which will be added as we gradually improve the system.

This is not currently equivalent to SBOM/SPDX-license-dumper, etc., so I won't add them all at once in this PR. It's simply a very basic list of build results that will be output on every build. If we need to output SBOM, we should output it separately.

henderkes (approved):

PR #1214, opened 2026-07-15
Feat/pgo v3 split

crazywhalecc:
## Checklist before merging

- If you modified `*.php` or `*.yml`, run them locally to ensure your changes are valid:
  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`
  - [ ] `bin/spc dev:lint-config`

PR #1216, opened 2026-07-16
feat(spc-config): Separate resolved and non-resolved functions in spc-config

crazywhalecc:
## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`


henderkes (approved):
on the resolved path this acts the same way pgo-v3 does currently, I assume?

crazywhalecc:
> on the resolved path this acts the same way pgo-v3 does currently, I assume?

This change primarily aims to avoid directly modifying `config()` and breaking the `include_suggests` option. Internally, we need the resolved result, so a separate method is better. This aligns with the approach in pgo-v3, but the details may differ.

I think we can perform full testing after merging #1214 and before merging this PR, and if there are no issues, continue migrating the changes from pgo-v3 based on this. feat/pgo-v3-split basically includes most of the modifications that are unrelated to spc-config, pgo, and the toolchain.

PR #1218, opened 2026-07-21
Migrate extension sources from pecl to pie

crazywhalecc:
## What does this PR do?

Port from feat/pgo-v3 (henderkes):
- switch 30 ext-*.yml sources from pecl to pie (packagist)
- PIE downloader: expand minified packagist p2 metadata and pick the newest stable release instead of the newest entry (may be RC/beta)
- PIE downloader: default extract path to php-src/ext/<name>, so per-extension extract lines are no longer needed

## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`


henderkes (approved):

crazywhalecc:
@henderkes Some extensions build test failed.

- xlswriter pie version does not include submodule and build failed.
- opentelemetry pie version extension dir is in subdir `php-src/ext/opentelemetry/ext/`.
- igbinary pecl has [3.2.17RC1](https://pecl.php.net/package/igbinary/3.2.17RC1) version that has fixed this bug, but pie latest version is 3.2.16.

Either we revert these two ext, or just fix it .

henderkes:
Long-term we need to migrate. pecl is going down quite often lately and will not be up forever, it's in discussion often.

henderkes:
For xlswriter and opentelemetry the current workaround solution wouldn't be optimal. I wonder if pie even works in installing those?

henderkes:
IMplemented a workaround for igbinary and opentelemetry. Imagick and xlswriter will have to stay pecl for now.

crazywhalecc:
@henderkes CI looks green with new entire `source-root` refactor. I also implemented `getBuildDir()` API for PhpExtensionPackage. It's different from `getSourceDir()` and `getSourceRoot()`.

PR #1226, opened 2026-08-07
php 8.6 compatibility with all bundled extensions

henderkes:
## What does this PR do?

prepares php 8.6 compatibility
there's still a small issue with RHEL 9 and older, but that's for another pr

used craft.yml to test this:
```yml
php-version: 8.6
extensions: "ctype,curl,date,dom,fileinfo,filter,ftp,hash,intl,json,lexbor,libxml,mbstring,mbregex,opcache,openssl,password-argon2,pcntl,pcre,phar,posix,random,readline,reflection,session,shmop,simplexml,snmp,soap,sockets,sodium,spl,sqlite3,standard,tokenizer,uri,xml,xmlreader,xmlwriter,zip,zlib"
shared-extensions: "bcmath,bz2,calendar,dba,exif,ffi,gd,gettext,gmp,iconv,ldap,mysqli,mysqlnd,odbc,pdo,pdo_mysql,pdo_pgsql,pdo_sqlite,pdo_odbc,pgsql,tidy,xsl,sysvmsg,sysvsem,sysvshm"
sapi:
  - cli
  - frankenphp
  - embed
download-options:
  ignore-cache: php-src
build-options:
  enable-zts: true
```


src/Package/Extension/pgsql.php
```diff
@@ -24,16 +25,20 @@ class pgsql extends PhpExtensionPackage
...
     }
 
+    /** These override pkg-config, so they must carry libpq itself too */
+    public static function libpqConfigureVars(PackageBuilder $builder, PackageInstaller $installer): string
+    {
+        $closure = DependencyResolver::getResolvedPackageClosure(['postgresql'], array_keys($installer->getResolvedPackages()));
+        $libfiles = new SPCConfigUtil(['no_php' => true, 'libs_only_deps' => true])->configWithResolvedPackages($closure)['libs'];
+        return ' PGSQL_CFLAGS=-I' . $builder->getIncludeDir() .
+            ' PGSQL_LIBS="-L' . $builder->getLibDir() . ' ' . $libfiles . '"';
+    }
```
crazywhalecc: This is definitely not the optimal solution. We could use `configForResolvedBuild()` instead of a static helper to achieve the same result.

crazywhalecc (changes requested):

henderkes:
Swoole couldn't possibly work, I'm not sure why it's even tested.

henderkes:
all the failing tests are now not-build-in extensions @crazywhalecc

src/globals/patch/php-src-patches/cli_checks_86.patch
crazywhalecc, line 1: Haven't verified yet, and phpmicro also need this patch.

src/globals/test-extensions.php
crazywhalecc, line 1: We don't need this file any more.

src/Package/Artifact/libaom.php
```diff
@@ -17,6 +18,18 @@ class libaom
...
+        if (str_contains($content, '_POSIX_C_SOURCE')) {
+            return;
+        }
+        foreach (['if(ENABLE_APPS)', 'if(ENABLE_EXAMPLES)'] as $guard) {
+            if (str_contains($content, $guard)) {
+                FileSystem::replaceFileStr($cmakelists, $guard, $guard . "\n  add_definitions(-D_POSIX_C_SOURCE=200112L)");
+                return;
+            }
+        }
+        throw new ValidationException('libaom CMakeLists.txt has neither an ENABLE_APPS nor an ENABLE_EXAMPLES guard to patch');
```
crazywhalecc: Why here?

src/globals/patch/php-src-patches/cli_checks_86.patch
henderkes: this is the micro patch, phpmicro just doesn't have it upstream yet

src/Package/Artifact/libaom.php
henderkes: Fixed, upstream v3 already had the patch file.

src/globals/test-extensions.php
henderkes: Should we remove it?

src/StaticPHP/Artifact/Downloader/Type/PhpRelease.php
```diff
@@ -108,9 +168,13 @@ protected function fetchPhpReleaseInfo(string $name, array $config, ArtifactDown
...
+        if (!is_array($info)) {
             throw new DownloaderException("Invalid PHP release info received for version {$phpver}");
         }
+        if (!isset($info['version'])) {
+            logger()->debug("php.net has no release for PHP version {$phpver}: " . ($info['error'] ?? 'no version in response'));
+            return null;
+        }
         return $info;
     }
 }
```
crazywhalecc, line 130:
```suggestion
<?php

declare(strict_types=1);

namespace StaticPHP\Artifact\Downloader\Type;

use StaticPHP\Artifact\ArtifactDownloader;
use StaticPHP\Artifact\Downloader\DownloadResult;
use StaticPHP\Exception\DownloaderException;

class PhpRelease implements DownloadTypeInterface, ValidatorInterface, CheckUpdateInterface
{
    public const string DEFAULT_PHP_DOMAIN = 'https://www.php.net';

    public const string API_URL = '/releases/index.php?json&version={version}';

    public const string DOWNLOAD_URL = '/distributions/php-{version}.tar.xz';

    public const string GIT_URL = 'https://github.com/php/php-src.git';

    public const string GIT_REV = 'master';

    private ?string $sha256 = '';

    public function download(string $name, array $config, ArtifactDownloader $downloader): DownloadResult
    {
        $phpver = $downloader->getOption('with-php', '8.5');
        // Handle 'git' version to clone from php-src repository
        if ($phpver === 'git') {
            $this->sha256 = null;
            return (new Git())->download($name, ['url' => self::GIT_URL, 'rev' => self::GIT_REV], $downloader);
        }
        // php.net does not publish pre-releases: fall back to the php-src git tag archive
        if (($info = $this->fetchPhpReleaseInfo($phpver, $config, $downloader)) === null) {
            $this->sha256 = null;
            return (new PhpTag())->download($name, ['version' => $phpver, 'extract' => $config['extract'] ?? null], $downloader);
        }
        ['version' => $version, 'url' => $url, 'filename' => $filename, 'sha256' => $this->sha256] = $this->resolveRelease($info, $config);
        logger()->debug("Downloading PHP release {$version} from {$url}");
        $path = DOWNLOAD_PATH . "/{$filename}";
        default_shell()->executeCurlDownload($url, $path, retries: $downloader->getRetry());
        return DownloadResult::archive($filename, config: $config, extract: $config['extract'] ?? null, version: $version, downloader: static::class);
    }

    public function validate(string $name, array $config, ArtifactDownloader $downloader, DownloadResult $result): bool
    {
        if ($this->sha256 === null) {
            logger()->debug('Php-src is downloaded from non-release source, skipping validation.');
            return true;
        }

        if ($this->sha256 === '') {
            logger()->error("No SHA256 checksum available for validation of {$name}.");
            return false;
        }

        $path = DOWNLOAD_PATH . "/{$result->filename}";
        $hash = hash_file('sha256', $path);
        if ($hash !== $this->sha256) {
            logger()->error("SHA256 checksum mismatch for {$name}: expected {$this->sha256}, got {$hash}");
            return false;
        }
        logger()->debug("SHA256 checksum validated successfully for {$name}.");
        return true;
    }

    public function checkUpdate(string $name, array $config, ?string $old_version, ArtifactDownloader $downloader): CheckUpdateResult
    {
        $phpver = $downloader->getOption('with-php', '8.5');
        if ($phpver === 'git') {
            // git version: delegate to Git checkUpdate with master branch
            return (new Git())->checkUpdate($name, ['url' => 'https://github.com/php/php-src.git', 'rev' => 'master'], $old_version, $downloader);
        }
        if (($info = $this->fetchPhpReleaseInfo($phpver, $config, $downloader)) === null) {
            // pre-release: delegate to the php-src git tag check
            return (new PhpTag())->checkUpdate($name, ['version' => $phpver], $old_version, $downloader);
        }
        $new_version = $info['version'];
        return new CheckUpdateResult(
            old: $old_version,
            new: $new_version,
            needUpdate: $old_version === null || $new_version !== $old_version,
        );
    }

    /** @return array{version: string, url: string, filename: string, sha256: null|string} */
    protected function resolveRelease(array $info, array $config): array
    {
        $version = $info['version'];
        $filename = null;
        $sha256 = '';
        foreach ($info['source'] ?? [] as $source) {
            if (str_ends_with($source['filename'] ?? '', '.tar.xz')) {
                $sha256 = $source['sha256'] ?? '';
                $filename = $source['filename'];
                break;
            }
        }
        if ($filename === null) {
            throw new DownloaderException("No suitable source tarball found for PHP version {$version}");
        }
        $url = $config['domain'] ?? self::DEFAULT_PHP_DOMAIN;
        $url .= str_replace('{version}', $version, self::DOWNLOAD_URL);
        return ['version' => $version, 'url' => $url, 'filename' => $filename, 'sha256' => $sha256];
    }

    /** @return null|array null when php.net does not publish this version (yet) */
    protected function fetchPhpReleaseInfo(string $phpver, array $config, ArtifactDownloader $downloader): ?array
    {
        $url = $config['domain'] ?? self::DEFAULT_PHP_DOMAIN;
        $url .= self::API_URL;
        $url = str_replace('{version}', $phpver, $url);
        logger()->debug("Fetching PHP release info for version {$phpver} from {$url}");

        // Fetch PHP release info first
        $info = default_shell()->executeCurl($url, retries: $downloader->getRetry());
        if ($info === false) {
            throw new DownloaderException("Failed to fetch PHP release info for version {$phpver}");
        }
        $info = json_decode($info, true);
        if (!is_array($info)) {
            throw new DownloaderException("Invalid PHP release info received for version {$phpver}");
        }
        if (!isset($info['version'])) {
            logger()->debug("php.net has no release for PHP version {$phpver}: " . ($info['error'] ?? 'no version in response'));
            return null;
        }
        return $info;
    }
}
```

I think it's clearer to separate `php-tag` part (just like we use `git` delegation):

```php
class PhpTag implements DownloadTypeInterface, CheckUpdateInterface
{
    use GitHubTokenSetupTrait;

    public const string TAGS_API = 'https://api.github.com/repos/php/php-src/git/matching-refs/tags/{prefix}';

    public const string ARCHIVE_URL = 'https://github.com/php/php-src/archive/refs/tags/php-{version}.tar.gz';

    public function download(string $name, array $config, ArtifactDownloader $downloader): DownloadResult
    {
        $version = $this->resolveLatestTag($config['version'], $downloader->getRetry());
        logger()->notice("PHP {$version} is a pre-release, downloading its source archive from php-src git tag.");
        $url = str_replace('{version}', $version, self::ARCHIVE_URL);
        logger()->debug("Downloading PHP release {$version} from {$url}");
        $path = DOWNLOAD_PATH . "/php-{$version}.tar.gz";
        default_shell()->executeCurlDownload($url, $path, retries: $downloader->getRetry());
        return DownloadResult::archive(basename($path), $config, $config['extract'] ?? null, version: $version, downloader: static::class);
    }

    public function checkUpdate(string $name, array $config, ?string $old_version, ArtifactDownloader $downloader): CheckUpdateResult
    {
        $new_version = $this->resolveLatestTag($config['version'], $downloader->getRetry());
        return new CheckUpdateResult(
            old: $old_version,
            new: $new_version,
            needUpdate: $old_version === null || $new_version !== $old_version,
        );
    }

    /**
     * Resolve the highest php-src git tag matching the requested version.
     * Accepts a branch ('8.6', picks the latest 8.6.x tag) or an exact version ('8.6.0RC1').
     */
    protected function resolveLatestTag(string $phpver, int $retries = 0): string
    {
        $is_branch = preg_match('/^\d+\.\d+$/', $phpver) === 1;
        $prefix = $is_branch ? "php-{$phpver}." : "php-{$phpver}";
        $url = str_replace('{prefix}', $prefix, self::TAGS_API);
        logger()->debug("PHP version {$phpver} is not published on php.net, looking it up in php-src tags from {$url}");

        $data = default_shell()->executeCurl($url, headers: $this->getGitHubTokenHeaders(), retries: $retries);
        if ($data === false) {
            throw new DownloaderException("Failed to fetch php-src git tags for PHP version {$phpver}");
        }
        $data = json_decode($data, true);
        if (!is_array($data)) {
            throw new DownloaderException("Invalid php-src git tag list received for PHP version {$phpver}");
        }

        $pattern = '/^php-(' . preg_quote($phpver, '/') . ($is_branch ? '\.\d+' : '') . '(?:(?:alpha|beta|RC)\d+)?)$/';
        $versions = [];
        foreach ($data as $ref) {
            $tag = substr((string) ($ref['ref'] ?? ''), strlen('refs/tags/'));
            if (preg_match($pattern, $tag, $match) === 1) {
                $versions[] = $match[1];
            }
        }
        if ($versions === []) {
            throw new DownloaderException("PHP version {$phpver} is not available on php.net nor tagged in php-src.");
        }
        usort($versions, version_compare(...));
        return end($versions);
    }
}
```

crazywhalecc: This solution was generated by kimi on a temporary basis and is for reference only. But anyway I think it's better to separate them like this than to mix them up.

src/globals/test-extensions.php
crazywhalecc: Yeah we are relying on test-bot now.

docs/deps-craft-yml.md
```diff
@@ -26,6 +26,9 @@ build-options:
   enable-zts: false
   # Disable smoke test, or for specific SAPIs comma-separated (default: false)
   no-smoke-test: false
+  # Do not abort when a shared extension fails to build or to load: skip it, drop its .so, and
+  # record it in buildroot/skipped-shared-extensions.json (default: false)
+  allow-shared-ext-failure: false
```
crazywhalecc: Our current test pipeline (the StaticPHP project itself) doesn't test pipeline with the shared build, so I'm unsure if it should be merged here.

In any case, our default test suite should be guaranteed to be green, but that doesn't mean test/php-86, test/tier2, etc., should be guaranteed to be green.

If this failure-tolerance option is absolutely necessary, perhaps we should consider separating it into a separate PR and finding a framework change that makes it easy to hook into this process (detect failure - skip).

henderkes: it's necessary for packaging to work until all extensions are php 8.6 compatible.

crazywhalecc: Therefore, if this is a temporary measure, we need to narrow the scope of the modification so that it can be easily deleted later, rather than using it in multiple places (or making an exception only for this instance).

Otherwise, I think we should handle this failure skipping logic through separate public relations requests.

henderkes: should I also remove the XtOffsetOf patch then? It's currently fixing 13 extensions for a relatively small change. But of course they'll be working without it at some point.

crazywhalecc: XtOffsetOf patches should be kept. It's a clean patch and suitable for this PR.

src/Package/Target/php/unix.php
```diff
@@ -41,14 +41,16 @@ public function patchBeforeBuildconf(TargetPackage $package): void
         // php-src patches from micro (reads SPC_MICRO_PATCHES env var)
         SourcePatcher::patchPhpSrc();
 
-        // patch configure.ac for musl and musl-toolchain
+        // patch libc detection for musl and musl-toolchain
         $musl = SystemTarget::getTargetOS() === 'Linux' && SystemTarget::getLibc() === 'musl';
         FileSystem::backupFile(SOURCE_PATH . '/php-src/configure.ac');
```
crazywhalecc:
```suggestion
```

crazywhalecc (approved):

PR #1229, opened 2026-08-13
Fix Swoole thread context handling for NTS builds

crazywhalecc:
## What does this PR do?

swoole thread mode (multi-threading) requires ZTS. Our chosen swoole-thread and thread-context are mutually exclusive, because thread-context uses process-global state (a single lock and reactor pointer), which conflicts with multi-reactor thread mode. For NTS builds, just use the upstream default (boost asm context): it is faster and better tested. 

Enabling thread-context on NTS is known to cause segfaults, e.g. Process::signal() in onManagerStart. I will submit the real patch upstream later.

## Checklist before merging

  - [ ] `composer cs-fix`
  - [ ] `composer analyse`
  - [ ] `composer test`

### DIFF
--- a/src/Package/Extension/swoole.php
+++ b/src/Package/Extension/swoole.php
@@ -73,7 +73,7 @@ class swoole extends PhpExtensionPackage
         // commonly used feature: coroutine-time
         $arg .= ' --enable-swoole-coro-time';
 
-        $arg .= $builder->getOption('enable-zts') ? ' --enable-swoole-thread --disable-thread-context' : ' --disable-swoole-thread --enable-thread-context';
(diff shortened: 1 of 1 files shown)


henderkes:
The swoole team is usually very fast fixing these kinds of things, worth looking if they can't fix it quick before we lose the capability.

crazywhalecc:
Looks like `--enable-thread-context` is slower than `--disable-thread-context`. Its more suitable for debugging compatibility, so even if the upstream is fixed, continuing to use `--disable-thread-context` on NTS is still the correct choice (config.m4 also disables it by default).

henderkes (approved):

PR #1236, opened 2026-08-21
fix pgsql and other polyfill libs leaking into configure

henderkes:
currently when building pgsql on old glibc systems the polyfill libs from pgsql and libedit advertise `strlcpy` and `strlcat` as available, even though they only are when linking against those libs.

### DIFF
--- a/src/Package/Target/php/unix.php
+++ b/src/Package/Target/php/unix.php
@@ -143,15 +143,12 @@ trait unix
             $configure_str = str_replace('--with-pic', '--enable-pic', $configure_str);
         }
 
-        // reuse the same make vars so configure conftest links use the same LIBS (incl. -framework flags)
-        $vars = $this->makeVars($installer);
-
         // run ./configure with args
         $this->seekPhpSrcLogFileOnException(fn () => shell()->cd($package->getSourceDir())->setEnv([
             'CFLAGS' => getenv('SPC_CMD_VAR_PHP_MAKE_EXTRA_CFLAGS'),
             'CPPFLAGS' => "-I{$package->getIncludeDir()}",
             'LDFLAGS' => "-L{$package->getLibDir()} " . getenv('SPC_CMD_VAR_PHP_MAKE_EXTRA_LDFLAGS'),
-            'LIBS' => $vars['EXTRA_LIBS'] ?? '',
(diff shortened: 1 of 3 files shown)


crazywhalecc (approved):

src/Package/Target/php/unix.php
```diff
@@ -143,15 +143,12 @@ public function configureForUnix(TargetPackage $package, PackageInstaller $insta
...
-        // reuse the same make vars so configure conftest links use the same LIBS (incl. -framework flags)
-        $vars = $this->makeVars($installer);
-
         // run ./configure with args
         $this->seekPhpSrcLogFileOnException(fn () => shell()->cd($package->getSourceDir())->setEnv([
             'CFLAGS' => getenv('SPC_CMD_VAR_PHP_MAKE_EXTRA_CFLAGS'),
             'CPPFLAGS' => "-I{$package->getIncludeDir()}",
             'LDFLAGS' => "-L{$package->getLibDir()} " . getenv('SPC_CMD_VAR_PHP_MAKE_EXTRA_LDFLAGS'),
-            'LIBS' => $vars['EXTRA_LIBS'] ?? '',
+            'LIBS' => SystemTarget::getRuntimeLibs(),
```
crazywhalecc, line 151: @henderkes 

The LIBS configuration in PHP's `./configure` was downgraded from `makeVars()['EXTRA_LIBS']` (the complete library, including `-lstdc++/-lc++`, `-lssl -lcrypto`, and `frameworks`) to `SystemTarget::getRuntimeLibs()` (only the basic runtime library).

Then, the `conftest` linker test in `configure` failed due to missing C++ runtime and OpenSSL symbols, causing some jobs crash with the error `configure: error`.

This leads to a regression problem.

henderkes: SystemTarget::getRuntimeLibs() should be returning -lstdc++ if we have any C++ extensions.

The problem with passing $vars['EXTRA_LIBS'] is passing all the polyfill libraries which will make php get wrong HAVE_XXX defines which will lead to make fail.

Are frameworks missing from getRuntimeLibs()? Then they need to be added. The implementation we have in main branch is correct.

crazywhalecc: Fixed in my PR

PR #1240, opened 2026-09-08
frankenphp: build the Mercure hub with its 0.x compatibility tags

dunglas:
Mercure 1.0 (https://github.com/dunglas/mercure/releases/tag/v1.0.0-alpha.3) gates its 0.x behaviors behind two build tags, `deprecated_topic` and `deprecated_claim`, and only honors them when the Caddyfile sets `protocol_version_compatibility 8`. Without the tags, a hub in compatibility mode rejects 0.x tokens and `topic=` selectors outright.

FrankenPHP is adopting the 1.0 hub in https://github.com/php/frankenphp/pull/2611 and ships both tags in its Docker images and CI builds, like the official Mercure binaries, so users can migrate progressively. Static binaries take their `-tags` list from here, and `GOFLAGS` cannot add to a `-tags` given on the command line, hence this change.

Go ignores unknown build tags, so FrankenPHP versions still on Mercure 0.24 build unchanged.

### DIFF
--- a/src/Package/Target/php/frankenphp.php
+++ b/src/Package/Target/php/frankenphp.php
@@ -144,7 +144,7 @@ trait frankenphp
                 '-X \'github.com/caddyserver/caddy/v2.CustomBinaryName=frankenphp\' ' .
                 '-X \'github.com/caddyserver/caddy/v2.CustomVersion=FrankenPHP ' .
                 "v{$frankenphp_version} PHP {$libphp_version} Caddy'\\\" " .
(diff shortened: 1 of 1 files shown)


henderkes (approved):

henderkes:
For this to take any effect, we need to [switch to spc v3](https://github.com/php/frankenphp/pull/2618) which is still open because I don't have a mac to fix x86.

crazywhalecc (approved):

